PR #2, opened 2017-04-03
Truffle testsuite

holiman:
This PR moves the project layout to Truffle standard, and thus renames some of the files. 
The solidity-version has been downgraded to `0.4.8`, due to to a bug in solc-js, which caused the `truffle compile` to place the artefacts in the wrong place (see https://gitter.im/ConsenSys/truffle?at=58e21d1bd7ab15e81724367e ) . 

Note, however, that the `pragma solidity` defines compatibility, and does not preclude compiling with `0.4.10` as is. 

This PR also adds a [Jupyter Notebook](http://jupyter.org/) , which can be used to generete testdata for the formulas, and to graph precision-loss.


contracts/BancorEtherToken.sol
```diff
@@ -1,4 +1,4 @@
-pragma solidity ^0.4.10;
```
yudilevi: Any reason why you changed all file names from lowercase + underscore to camel case? Is there a best practice in solidity for that?

contracts/Migrations.sol
```diff
@@ -0,0 +1,23 @@
+pragma solidity ^0.4.4;
```
yudilevi: What is this used for in the context of our contracts?

contracts/BancorEtherToken.sol
holiman: It's because the truffle framework expects the class `FooBar` to reside in a file called `FooBar.sol`. Can't say if it's best practice or not, but it seems pretty common.

contracts/Migrations.sol
holiman: It's documented here: http://truffleframework.com/docs/getting_started/migrations . I haven't dived into it - for now it's just needed to deploy the solidity to `testrpc` and run the tests. It's not something that's needed for production deployment (although I'm guessing it can be used for production deployment aswell, particularly if there are non-trivial deployments that are needed, where you need to follow a certain order and initialize things with other thing's addresses and such)

PR #10, opened 2017-06-10
Moving the special case for selling the entire supply after the case of CRR=100

sidko:
The case of selling the entire supply is a very special case and will likely not occur very often (because it essentially means killing the whole token). However, the CRR=100 is much more common. It seems unnecessary for all contracts to pay gas to check for this condition, as the same result can be achieved by moving this down. It can be verified that the case of selling the entire supply works fine with the existing case of CRR=100.


yudilevi:
Indeed better to switch the other of these two cases.
I couldn't currently merge your pull request because of conflicts, so I made the change directly in the main branch.

Thanks!

PR #42, opened 2017-07-13
Improve accuracy and performance

barakman:
If base <= 20, then the upper bound of ln(base) is 3.
This will improve the accuracy (and performance) for the cases of 9 <= base <= 20.


barakman:
Got a slightly better improvement, will submit that in a minute

PR #175, opened 2017-08-03
Add BancorWalletFactory to generate user-friendly BancorWallet.

dabdevelop:


dabdevelop:
@yudilevi Actually, I don't have enough resources to fully test the Wallet Factory. But It passes the compile and almost all the code is originated from your version of BancorBuyer contract. If you think this pull request is useful, then consider admitted it, or any suggestion for revisions.

dabdevelop:
@yudilevi I created a DAO on https://github.com/DAB-Foundation/dab-dao. Which I think is helpful for the Bancor Protocol. It may move Bancor to a DAO.

PR #182, opened 2017-10-23
qq.wwExtend python utilities

Ewgen9:


barakman:
@Ewgen9:
Can you please clarify your intentions in this activity?
I have not been able to figure out what change (if any) was suggested here.
Thank you.

PR #190, opened 2018-02-07
Upgrade solidity compiler version

ordd:


barakman:
Better change those `public` functions to `external`.

For example:

- function calculatePurchaseReturn
- function calculateSaleReturn

PR #201, opened 2018-04-05
validateGasPrice does not allow lower gas prices

Detoo:
Hi all,

Not sure if it was intentional, but I've noticed that the new [BancorGasLimit contract](https://etherscan.io/address/0x7fd539ef2be3192b6d6bdf095968bac2d74daa6e) deployed yesterday somehow only validates gas prices exactly the set price and does not allow lower prices.

This has led to tx failures such as [this one](https://etherscan.io/tx/0x86c599aa7c811aa9c290020d12acd6b743db2de84cc3b2aebf251802fef5b434). On the contrary, the only [transactions](https://etherscan.io/tx/0xf9046e7615e02cd153703a1c721626dcbdc44fac615a48f00a24bd9fb9405012) that worked are those with gas price set exactly the limit price (5 Gwei).

In addition, Etherscan.io is no longer able to display the contract's gasPrice because validateGasPrice() would fail on contract read:
screen shot 2018-04-05 at 12 56 47 am

### DIFF
--- a/solidity/contracts/BancorGasPriceLimit.sol
+++ b/solidity/contracts/BancorGasPriceLimit.sol
@@ -58,6 +58,6 @@ contract BancorGasPriceLimit is IBancorGasPriceLimit, Owned, Utils {
         view
         greaterThanZero(_gasPrice)
     {
-        require(_gasPrice == gasPrice);
+        require(_gasPrice <= gasPrice);
     }
 }


Detoo:
Yes, thanks for the effort :)

Detoo:
Seen what was happening there. Fascinating! And nice counter measurement

yudilevi:
Hehe, thanks. It's a work in progress but hopefully we can eventually get rid that sorted out. :)

PR #206, opened 2018-04-30
Upgrader v9

ordd:


solidity/contracts/converter/BancorConverterFactory.sol
```diff
@@ -7,6 +7,7 @@ import '../utility/interfaces/IContractFeatures.sol';
     Bancor Converter Factory
 */
 contract BancorConverterFactory is IBancorConverterFactory {
+    string public version = '0.9';
```
yudilevi: why is it needed?

solidity/contracts/converter/BancorConverterFactory.sol
```diff
@@ -38,7 +39,7 @@ contract BancorConverterFactory is IBancorConverterFactory {
         uint32 _maxConversionFee,
         IERC20Token _connectorToken,
         uint32 _connectorWeight
-    ) 
+    )
         public
```
yudilevi: why the line breaks?

solidity/contracts/converter/BancorConverterUpgrader.sol
```diff
@@ -1,12 +1,14 @@
...
 import './interfaces/IBancorConverterFactory.sol';
 import '../utility/Owned.sol';
+import '../utility/interfaces/IContractFeatures.sol';
 import '../utility/interfaces/IWhitelist.sol';
 
 /*
     Bancor converter dedicated interface
 */
 contract IBancorConverter is IOwned {
+    uint256 public constant FEATURE_CONVERSION_WHITELIST = 1 << 0;
```
yudilevi: I think we should just rename this contract, import IBancorConverter and inherit from it instead of duplicating all these functions

solidity/contracts/converter/BancorConverterUpgrader.sol
```diff
@@ -84,6 +90,18 @@ contract BancorConverterUpgrader is Owned {
         bancorConverterFactory = _bancorConverterFactory;
     }
 
+    /*
+        @dev allows the owner to update the contract features contract address
+
+        @param _features   address of a contract features contract
+    */
+    function setContractFeatures(IContractFeatures _features)
+        public
```
yudilevi: why the line breaks?

PR #207, opened 2018-04-30
Management

ordd:


barakman:
Since the `ownerOrManagerOnly` modifier is implemented inside the `Managed` contract, it looks like this contract doesn't need to be `Owned` (i.e., inherit from the `Owned` contract).

PR #210, opened 2018-05-10
Add multiple convert for.

ordd:


solidity/contracts/BancorNetwork.sol
```diff
@@ -173,6 +173,107 @@ contract BancorNetwork is IBancorNetwork, TokenHolder, ContractIds {
...
+        @dev converts token to any other token in the bancor network
+        by following a predefined conversion paths and transfers the result
+        tokens to a target account.
+
+        @param _path        conversion path, see conversion path format above
+        @param _amount      amount to convert from (in the initial source token)
+        @param _minReturn   if the conversion results in an amount smaller than the minimum return - it is cancelled, must be nonzero
+        @param _for         account that will receive the conversion result
+        @param _block       if the current block exceeded the given parameter - it is cancelled
+        @param _nonce       the nonce of the sender address
```
yudilevi: since the nonce is not really used/verified, let's get rid of it

solidity/contracts/BancorNetwork.sol
```diff
@@ -173,6 +173,107 @@ contract BancorNetwork is IBancorNetwork, TokenHolder, ContractIds {
...
+        @param _v           parameter that can be parsed from the transaction signature
```
yudilevi: doesn't look like a good description

PR #219, opened 2018-06-25
Return by path

ordd:


solidity/contracts/BancorNetwork.sol
```diff
@@ -11,6 +12,13 @@ import './utility/interfaces/IWhitelist.sol';
 import './token/interfaces/IEtherToken.sol';
 import './token/interfaces/ISmartToken.sol';
 
+/*
+    Bancor converter dedicated interface
+*/
+contract IBancorConverterExtended is IBancorConverter, IOwned {
```
yudilevi: this should be added to the real interface, there might be other use cases for it.

solidity/contracts/BancorNetwork.sol
```diff
@@ -329,6 +337,76 @@ contract BancorNetwork is IBancorNetwork, TokenHolder, ContractIds, FeatureIds {
         return (toToken, _amount);
     }
 
+    /**
+        @dev returns the expected return for converting a specific amount by following
```
yudilevi: "returns the expected return amount", "following a given conversion path"

solidity/contracts/BancorNetwork.sol
```diff
@@ -329,6 +337,76 @@ contract BancorNetwork is IBancorNetwork, TokenHolder, ContractIds, FeatureIds {
...
+
+        // iterate over the conversion path
+        for (uint256 i = 1; i < pathLength; i += 2) {
+            smartToken = ISmartToken(_path[i]);
+            toToken = _path[i + 1];
+            converter = IBancorConverterExtended(smartToken.owner());
+
+            if (toToken == smartToken) { // buy the smart token
+                // check if the current smart token supply had changed in the previous iteration
+                supply = smartToken == prevSmartToken ? supply : smartToken.totalSupply();
```
yudilevi: why do we need it? don't we know exactly when it changes?

solidity/contracts/converter/interfaces/IBancorConverter.sol
```diff
@@ -11,4 +11,6 @@ contract IBancorConverter {
     function conversionWhitelist() public view returns (IWhitelist) {}
     // deprecated, backward compatibility
     function change(IERC20Token _fromToken, IERC20Token _toToken, uint256 _amount, uint256 _minReturn) public returns (uint256);
+    function getConnectorBalance(IERC20Token _connectorToken) public view returns (uint256);
```
yudilevi: 1. these two aren't deprecated so they need to move up
2. getConnectorBalance will only return the balance but we might as well just add all the connector values by returning all connector values (isSet shouldn't be returned but should be checked when returning the values)
3. getFinalAmount is definitely not a proper function name for an external interface and should be internal if not private in the converter. I think it's better to return the fee and deal with the calculation externally

solidity/test/BancorNetwork.js
```diff
@@ -599,6 +599,104 @@ contract('BancorNetwork', accounts => {
         }
     });
 
+    it('verifies that getReturnByPath return the correct amount for buying smart token', async () => {
```
yudilevi: `verifies that getReturnByPath returns` (missed the 's')
`for buying the smart token` (missed the 'the')

solidity/test/BancorNetwork.js
```diff
@@ -599,6 +599,104 @@ contract('BancorNetwork', accounts => {
...
+
+    it('verifies that getReturnByPath return the correct amount for buying smart token through multiple converters', async () => {
+        let returnByPath = await bancorNetwork.getReturnByPath.call(smartToken2QuickBuyPath, 10000);
+        let balanceBeforeTransfer = await smartToken2.balanceOf.call(accounts[1]);
+        await bancorNetwork.convertFor(smartToken2QuickBuyPath, 10000, 1, accounts[1], { value: 10000 });
+        let balanceAfterTransfer = await smartToken2.balanceOf.call(accounts[1]);
+        assert.equal(returnByPath, balanceAfterTransfer - balanceBeforeTransfer);
+    });
+
+    it('verifies that getReturnByPath return the correct amount for connector conversion', async () => {
```
yudilevi: `for cross connector conversion`

solidity/test/BancorNetwork.js
```diff
@@ -599,6 +599,104 @@ contract('BancorNetwork', accounts => {
...
+        let returnByPath = await bancorNetwork.getReturnByPath.call(smartToken2QuickSellPath, 100);
+        let balanceBeforeTransfer = web3.eth.getBalance(accounts[1]);
+        let res = await converter2.quickConvert(smartToken2QuickSellPath, 100, 1, { from: accounts[1] });
+        let transaction = web3.eth.getTransaction(res.tx);
+        let transactionCost = transaction.gasPrice.times(res.receipt.cumulativeGasUsed);
+        let balanceAfterTransfer = web3.eth.getBalance(accounts[1]);
+        assert.equal(returnByPath.toNumber(), balanceAfterTransfer.minus(balanceBeforeTransfer).plus(transactionCost).toNumber());
+    });
+
+    it('verifies that getReturnByPath return the correct amount for selling smart token with long selling path', async () => {
```
yudilevi: `with a long conversion path`

solidity/test/BancorNetwork.js
```diff
@@ -599,6 +599,104 @@ contract('BancorNetwork', accounts => {
...
+        let returnByPath = await bancorNetwork.getReturnByPath.call(path, 100);
+        let balanceBeforeTransfer = web3.eth.getBalance(accounts[1]);
+        let res = await converter3.quickConvert(path, 100, 1, { from: accounts[1] });
+        let transaction = web3.eth.getTransaction(res.tx);
+        let transactionCost = transaction.gasPrice.times(res.receipt.cumulativeGasUsed);
+        let balanceAfterTransfer = web3.eth.getBalance(accounts[1]);
+        assert.equal(returnByPath.toNumber(), balanceAfterTransfer.minus(balanceBeforeTransfer).plus(transactionCost).toNumber());
+    });
+
+    it('verifies that getReturnByPath returns the same amount as getReturn when converting from a connector to the token', async () => {
```
yudilevi: `when converting a connector to the smart token`

PR #225, opened 2018-08-15
Test mnemonic fix

davidbancor:
trusted/untrusted private keys are no longer hardcoded into the tests (So ganache will work with any mnemonic), and have been replaces with accounts[3] and accounts[1] (addresses) respectively.

### DIFF
--- a/solidity/test/BancorNetwork.js
+++ b/solidity/test/BancorNetwork.js
@@ -16,10 +16,6 @@ const utils = require('./helpers/Utils');
 const ethUtil = require('ethereumjs-util');
 const web3Utils = require('web3-utils');
 
-// mnemonic: rose limit dog cannon adult lizard siege tumble job puzzle only trim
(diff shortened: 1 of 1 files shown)


solidity/test/BancorNetwork.js
```diff
@@ -43,34 +39,11 @@ let smartToken1QuickSellPath;
...
     try {
-        const sig = ethUtil.ecsign(
-            new Buffer(msgToSign.slice(2), 'hex'),
-            new Buffer(privateKey, 'hex'));
-        const r = `0x${sig.r.toString('hex')}`;
-        const s = `0x${sig.s.toString('hex')}`;
-        const v = sig.v;
-
-        return { v: v, r: r, s: s };
+        const sig = web3.eth.sign(signerAddress, ethUtil.bufferToHex(msgToSign));
```
yudilevi: nice

PR #237, opened 2018-09-04
 fix bug that can not access constant variable in FeatureIds

sekisamu:


yudilevi:
Hey :)

This also includes a formula document change - was that intentional?

barakman:
@yudilevi:
Intentional or not, it is wrong (should be reverted).

yudilevi:
@hammeWang hey man, can you please revert the change to the documentation?

Thanks!

PR #258, opened 2018-11-01
Safemath library ban 1612

yuval-weiss:


barakman:
@yuval-weiss:
In the `SafeMath` library, you should probably replace all `assert`s with `require`s, because:
1. It is the correct way. You use `assert` for something that can never happen (so theoretically, an `assert` statement can be omitted once the code has been 100%-verified). You use `require` for asserting user input, which may well be wrong (since it's up to the user, not up to you).
2. Because it saves gas to the caller upon failure (and if the SafeMath function is invoked from an "expensive path", then it may well be a lot of gas).

Keep in mind that failure tests should be fixed as well (by replacing `catchInvalidOpcode` with `catchRevert`).

PR #262, opened 2018-11-25
Minor: use memory list rather than storing values

wanseob:
It seems that the variable `convertPath` is only used for passing to other functions as a parameter.
How about using memory array instead of using storage?


James-Lim:
This PR passes the `npm test`, please check.

barakman:
@wanseob: Thank you very much for this.

Unaware of this PR, I have recently applied it (almost identically), and it is currently in branch 0.4.11.

At the time, I believe that the reason for this (rather-undesirable) usage of storage was a workaround for the stack restriction (maximum 16 slots) of the Solidity compiler.

On top of the subsequent extra gas cost (~80K on the first operation, ~20K on successive operations), it is of course logically incorrect to save the path in storage to begin with.

Luckily, we've been able to figure out a way to avoid overloading the stack, and subsequently fix this issue as well.

I am closing this PR, but would like to thank you very much for bringing this up and for suggesting a workable solution!

BTW, I am going to adopt your `(path[0], path[1], path[2]) = (_fromToken, token, _toToken);` short syntax in a few minutes...

Thanks :)

wanseob:
That's good news! I understand the situation. Thanks for the kind reply :+1:

PR #269, opened 2019-01-28
Erc20 implementation workaround

davidbancor:
These updates give support for token contracts which don't have a return value for `transfer`, `transferFrom`, and `approve`.


solidity/contracts/BancorNetwork.sol
```diff
@@ -10,8 +10,10 @@ import './utility/SafeMath.sol';
 import './utility/interfaces/IContractRegistry.sol';
 import './utility/interfaces/IContractFeatures.sol';
 import './utility/interfaces/IWhitelist.sol';
+import './utility/interfaces/ITokenWhitelist.sol';
```
yudilevi: what's the logic behind that name? why whitelist?

solidity/contracts/BancorNetwork.sol
```diff
@@ -627,7 +629,56 @@ contract BancorNetwork is IBancorNetwork, TokenHolder, ContractIds, FeatureIds {
...
+
+        if (!tokenWhitelist.whitelistedTokens(_token)) {
+            // if the token isn't whitelisted, we assert on transfer
+            assert(_token.transfer(_to, _amount));
+        } else {
+            uint256 prevBalance = _token.balanceOf(_to);
+            // we have to cast the token contract in an interface which has no return value
+            INonStandardERC20(_token).transfer(_to, _amount);
+            uint256 postBalance = _token.balanceOf(_to);
+            assert(postBalance.sub(prevBalance) == _amount);
```
yudilevi: that might not work for tokens that have fees, maybe just check that the new amount is greater than the prev amount?

solidity/contracts/BancorNetwork.sol
```diff
@@ -640,10 +691,10 @@ contract BancorNetwork is IBancorNetwork, TokenHolder, ContractIds, FeatureIds {
 
         // if the allowance is nonzero, must reset it to 0 first
         if (_token.allowance(this, _spender) != 0)
-            assert(_token.approve(_spender, 0));
+            INonStandardERC20(_token).approve(_spender, 0);
```
yudilevi: so worst case the tx will fail?

solidity/contracts/converter/BancorConverter.sol
```diff
@@ -824,6 +826,54 @@ contract BancorConverter is IBancorConverter, SmartTokenController, Managed, Con
...
 
+    /**
+        @dev ensures transfer of tokens, taking into account that some ERC-20 implementations don't return
+        true on success but revert instead
+
+        @param _token     the token to transfer
+        @param _to        the address to transfer the tokens to
+        @param _amount    the amount to transfer
+    */
+    function ensureTransfer(IERC20Token _token, address _to, uint256 _amount) private {
```
yudilevi: how about moving these functions to a shared utility?

solidity/contracts/helpers/BadERC20Token.sol
```diff
@@ -0,0 +1,101 @@
+pragma solidity ^0.4.24;
+import './interfaces/IBadERC20Token.sol';
```
yudilevi: Bad or NonStandard?

solidity/contracts/utility/TokenWhitelist.sol
```diff
@@ -0,0 +1,24 @@
+pragma solidity ^0.4.24;
+import './Owned.sol';
+import './interfaces/ITokenWhitelist.sol';
+
+/*
+    Token Whitelist
+
+    manages tokens who don't return true/false on transfer/transferFrom but revert on failure instead 
+*/
+contract TokenWhitelist is ITokenWhitelist, Owned {
```
yudilevi: maybe just use the IWhitelist contract?

solidity/contracts/helpers/BadERC20Token.sol
davidbancor: "bad" for tests, "nonstandard interface" for production, silly naming

solidity/contracts/converter/BancorConverter.sol
davidbancor: definitely seems like the right idea, probably will move them to Utils

solidity/contracts/BancorNetwork.sol
davidbancor: when calling transferFrom afterwards, if the approve didn't succeed the tx should revert

solidity/contracts/BancorNetwork.sol
davidbancor: hmmm, didn't consider that case... we should discuss this further

solidity/contracts/BancorNetwork.sol
davidbancor: Not the most intuitive name, just didn't want to spend too long thinking about the name... we should think of a more explanatory name

PR #281, opened 2019-05-19
Fix getReturnByPath to respect old converter implementations returning 32 bytes instead of 64 bytes

k06a:
You can try this implementation here:
https://etherscan.io/address/0x24618ec3a496e576a65503a16afd3ccfcf613f96#readContract

Try to call `getReturnByPath` with arguments to test ETH => DAI:
```
["0xc0829421c1d260bd3cb3e0f06cfe2d52db2ce315", "0x1f573d6fb3f13d689ff844b4ce37794d79a7ff1c", "0x1f573d6fb3f13d689ff844b4ce37794d79a7ff1c", "0xee01b3ab5f6728adc137be101d99c678938e6e72", "0x89d24a6b4ccb1b6faa2625fe562bdd9a23260359"]
```
```
1000000000000000000
```

Issue details:
- https://github.com/ethereum/solidity/issues/4116
- https://medium.com/@chris_77367/explaining-unexpected-reverts-starting-with-solidity-0-4-22-3ada6e82308c

### DIFF
--- a/solidity/contracts/BancorNetwork.sol
+++ b/solidity/contracts/BancorNetwork.sol
@@ -504,14 +504,13 @@ contract BancorNetwork is IBancorNetwork, TokenHolder, ContractIds, FeatureIds {
         return (amount, fee);
     }
 
-    function converterGetReturn(address destination, bytes data) private returns(uint256 amount, uint256 fee) {
+    function converterGetReturn(address destination, bytes data) private view returns(uint256 amount, uint256 fee) {
         bytes memory ret = new bytes(64);
         bool success;
         assembly {
-            success := call(
+            success := staticcall(
                 div(mul(gas, 63), 64),
                 destination,
-                0,
                 add(data, 32),
                 mload(data),
                 add(ret, 32),
--- a/solidity/contracts/BancorNetwork.sol
+++ b/solidity/contracts/BancorNetwork.sol
@@ -486,7 +486,15 @@ contract BancorNetwork is IBancorNetwork, TokenHolder, ContractIds, FeatureIds {
                 supply = smartToken.totalSupply() - amount;
             }
             else { // cross connector conversion
-                (amount, fee) = converter.getReturn(fromToken, toToken, amount);
+                (amount, fee) = converterGetReturn(
+                    converter,
+                    abi.encodeWithSelector(
+                        converter.getReturn.selector,
+                        fromToken,
+                        toToken,
+                        amount
+                    )
+                );
             }
 
             prevSmartToken = smartToken;
@@ -494,6 +502,29 @@ contract BancorNetwork is IBancorNetwork, TokenHolder, ContractIds, FeatureIds {
         }
 
         return (amount, fee);
+    }
+
+    function converterGetReturn(address destination, bytes data) private returns(uint256 amount, uint256 fee) {
+        bytes memory ret = new bytes(64);
+        bool success;
+        assembly {
+            success := call(
+                sub(gas, 34710),
+                destination,
+                0,
+                add(data, 32),
+                mload(data),
+                add(ret, 32),
+                64
+            )
+        }
+
+        if (success) {
+            assembly {
+                amount := mload(add(ret, 32))
+                fee := mload(add(ret, 64))
+            }
+        }
     }
 
     /**
--- a/solidity/contracts/BancorNetwork.sol
+++ b/solidity/contracts/BancorNetwork.sol
@@ -16,6 +16,12 @@ import './token/interfaces/ISmartToken.sol';
 import './token/interfaces/INonStandardERC20.sol';
 import './bancorx/interfaces/IBancorX.sol';
 
+interface IBancorConverterOld {
+    function version() public view returns (bytes);
+    //function version() public view returns (uint16);
+    function getReserveBalance(IERC20Token _connectorToken) public view returns (uint256);
+}
+
 /*
     The BancorNetwork contract is the main entry point for bancor token conversions.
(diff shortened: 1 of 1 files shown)


k06a:
@barakman sorry to disturb you, but how it is possible that you push commits to our fork? O_o

k06a:
Ok, I got it:

image

k06a:
@barakman we also added one more fix here: https://etherscan.io/address/0x7a9b986420d734bb3fe98439a2d945ab97757595#contracts

k06a:
Add one more fix, relvated to converter API differences ^^

barakman:
@k06a: Hi.

Thank you for your work on this.

Can you please specify the version (and possibly the date as well), when `BancorConverter` had a function named `getReserveBalance`?

k06a:
@barakman this is an example of old BancorConverter which has `getReserveBalance`:
https://etherscan.io/address/0xeade8b32735f9c7871da862c18bf12caeaa4e6e9#contracts

barakman:
@k06a:

So I believe that your patch above ([Fix `getConnectorBalance`/`getReserveBalance` API difference](https://github.com/bancorprotocol/contracts/pull/281/commits/d0215cce96e64787cc4da8bad7275328d3956a4b)) is rather risky.

Here are the different version types in our contracts over time:
- `string  public version = '0.10'; // before 14/09/2018`
- `bytes32 public version = '0.11'; // from 14/09/2018`
- `uint16  public version = 12;     // from 18/12/2018`

Your patch identifies the difference between `string` and the other types, as illustrated in the following example:

**Contract:**
```
pragma solidity 0.4.24;

contract Test {
    string  public version10 = '0.10';
    bytes32 public version11 = '0.11';
    uint16  public version12 = 12;
    
    function test10() external view returns(bool success, uint256 version) {(success, version) = getVersion(this.version10.selector);}
    function test11() external view returns(bool success, uint256 version) {(success, version) = getVersion(this.version11.selector);}
    function test12() external view returns(bool success, uint256 version) {(success, version) = getVersion(this.version12.selector);}

    function getVersion(bytes4 selector) internal view returns(bool success, uint256 version) {
        address destination = this;
        bytes memory ret = new bytes(64);
        bytes memory data = abi.encodeWithSelector(selector);
        assembly {
            success := staticcall(
                gas,
                destination,
                add(data, 32),
                mload(data),
                add(ret, 32),
                64
            )
        }
        if (success) {
            assembly {
                version := mload(add(ret, 64))
            }
        }
    }
}
```
**Truffle Test:**
```
contract("Test", function(accounts) {
    it("test", async function() {
        const c = await artifacts.require("Test").new();
        await test(c.test10);
        await test(c.test11);
        await test(c.test12);
    });
    async function test(func) {
        try {
            const [success, version] = await func();
            console.log(`success = ${success}, version = ${version.toFixed()}`);
        }
        catch (error) {
            console.log(error.message);
        }
    }
});
```
**Printout:**
```
success = true, version = 4
success = true, version = 0
success = true, version = 0
```

However, function `getReserveBalance` does not exist when `string version = '0.10'`, and possibly also prior to that (in the example that you've linked, the version is `'0.4'`).

This means that in your code, there will be an attempt to execute a non-existing function, which I'm guessing will lead to an Invalid Jump exception (or some other type of failure).

k06a:
@barakman I agree that the final solution should respect exact versions instead of just version type. Had not enough time to solve it, also because my solution worked fine for our purposes.

barakman:
@k06a:
So we will not be able to adopt it officially on our repo.
Possibly the solution is to fix it in a manner similar to what you'e used for calling function `BancorConverter.getReturn`.
We are planning to integrate this fix once we finish considering some alternatives, and after adding some coverage (testing) for it.
All in all it seems to be a good solution.
Thanks :)

k06a:
@barakman yep, the provided solution is not production ready and you have to improve and stabilize it. It was originally developed by me on the hackathon, so I have no enough time for this.

barakman:
@k06a:

Thank you.

May I also ask how you came up with the magic constant in `sub(gas, 34710)`?

This solution works fine, however, it fails when we execute `solidity-coverage` (e.g., via `npm test 2`).

There's a good reason why it fails, because `solidity-coverage` modifies the contracts before the tests are executed (in order to detect every execution path).

So naturally, both `call` and `staticcall` can run out of gas here.

I am trying to find a workaround for this.

Thanks

k06a:
@barakman got it originally from here: https://github.com/gnosis/MultiSigWallet/blob/master/contracts/MultiSigWallet.sol

I think this constant is not actual anymore.

barakman:
@k06a:

Thank you.

And why have you chosen to use `add(ret, 32)` instead of just `ret`?

It seems like an outright error to copy 64 bytes into the address starting at `ret + 32`, when `ret` is an array of 64 bytes.

k06a:
@barakman I add 32, because first 32 bytes of the array contains length.

barakman:
@k06a:

Yes, that's true for `bytes memory ret = new bytes(64)`.

But why didn't you use `byte[64] memory ret`, or even `uint256[2] memory ret`?

Thanks

k06a:
@barakman yep this is also possible, arrays of static size do not have length field.

barakman:
@k06a:

[This comment of yours](https://github.com/bancorprotocol/contracts/pull/281#issuecomment-505330815) also answers my question on why you've used `outsize = mload(data)` (because the length appears at the first 32 bytes of `data`) and `out = add(data, 32)` (because the actual data starts after the length.

Thanks :)

PR #290, opened 2019-08-01
General fixes

barakman:
1. Add package-lock.
2. Update `solidity-docgen` version.
3. Minor fix in the `generate-docs.js`.
4. Fix the test of `BancorX` in compliance with a bug-fix between ganache 6.4.4 and ganache 6.5.1.
5. Remove '.sol' from all `artifacts.require` statements (i.e., use contract names instead of file names).


barakman:
Decided to revert the changes related to the upgrade of ganache-cli from v6.4.4 to v6.5.1, because they require fixing one of the tests, which in turns breaks solidity-coverage (since the latter relies on the same core used by ganache-cli v6.4.4).

PR #322, opened 2019-12-01
New contract registry

barakman:


solidity/contracts/utility/ContractRegistryClient.sol
```diff
@@ -0,0 +1,110 @@
...
+import './Utils.sol';
+import './interfaces/IContractRegistry.sol';
+
+/**
+  * @dev Base contract for ContractRegistry clients
+*/
+contract ContractRegistryClient is Owned, Utils {
+    IContractRegistry public registry;      // address of the current contract-registry
+    IContractRegistry public prevRegistry;  // address of the previous contract-registry
+    bool public allowRegistryUpdate = true; // allow/prevent changing the contract-registry
```
yudilevi: the comment isn't accurate - it restricts the change to the owner

solidity/contracts/utility/ContractRegistryClient.sol
```diff
@@ -0,0 +1,110 @@
...
+
+    /**
+      * @dev returns whether or not the caller is authorized to change the contract-registry
+     */
+    function callerIsAuthorized() internal view returns (bool) {
+        return msg.sender == owner;
+    }
+
+    /**
+      * @dev sets the contract-registry to the address of the contract-registry in the contract-registry
```
yudilevi: this comment needs work... maybe:
`updates the registry address to the new address defined by the current contract registry`
?

solidity/contracts/utility/ContractRegistryClient.sol
```diff
@@ -0,0 +1,110 @@
...
+    */
+    constructor(IContractRegistry _registry) internal validAddress(_registry) {
+        registry = IContractRegistry(_registry);
+        prevRegistry = IContractRegistry(_registry);
+    }
+
+    /**
+      * @dev returns whether or not the caller is authorized to change the contract-registry
+     */
+    function callerIsAuthorized() internal view returns (bool) {
```
yudilevi: maybe call it `isCallerAdmin`? Authorized implies authorized to execute a specific action I think

solidity/contracts/utility/ContractRegistryClient.sol
```diff
@@ -0,0 +1,110 @@
...
+    }
+
+    /**
+      * @dev sets the contract-registry to the address of the contract-registry in the contract-registry
+     */
+    function updateRegistry() public {
+        // verify that either updating is allowed or the caller is authorized
+        require(allowRegistryUpdate || callerIsAuthorized());
+
+        // get the address of the contract-registry in the contract-registry
```
yudilevi: get the address of the new contract registry from the existing contract registry?

solidity/contracts/utility/ContractRegistryClient.sol
```diff
@@ -0,0 +1,110 @@
...
+        address newRegistry = addressOf(CONTRACT_REGISTRY);
+
+        // verify that the new contract-registry is different and not zero
+        require(newRegistry != address(registry) && newRegistry != address(0));
+
+        // set the previous contract-registry as current contract-registry
+        prevRegistry = registry;
+
+        // set the current contract-registry as the new contract-registry
```
yudilevi: set the new contract registry address as the current registry

solidity/contracts/utility/ContractRegistryClient.sol
```diff
@@ -0,0 +1,110 @@
...
+
+        // set the current contract-registry as the previous contract-registry
+        registry = prevRegistry;
+
+        // ensure that only an authorized caller can perform future changes
+        allowRegistryUpdate = false;
+    }
+
+    /**
+      * @dev allows/prevents changing the contract-registry
```
yudilevi: it restricts it, not disables it

solidity/contracts/utility/ContractRegistryClient.sol
```diff
@@ -0,0 +1,110 @@
...
+      * 
+      * @param _disable    true to disable changes, false to enable changes
+    */
+    function disableRegistryUpdate(bool _disable) public ownerOnly {
```
yudilevi: disable -> restrict?

solidity/contracts/utility/ContractRegistryClient.sol
```diff
@@ -0,0 +1,110 @@
...
+        // verify that the caller is authorized
+        require(callerIsAuthorized());
```
yudilevi: it's already ownerOnly - why do we need this additional check?

solidity/contracts/utility/ContractRegistryClient.sol
```diff
@@ -0,0 +1,110 @@
...
+    function addressOf(bytes32 _contractName) internal view returns (address) {
+        return registry.addressOf(_contractName);
+    }
+
+    /**
+      * @dev verifies that the caller is mapped to the given contract name
+      * 
+      * @param _contractName    contract name
+    */
+    modifier only(bytes32 _contractName) {
```
yudilevi: this and the rest of the code should be at the top (constants -> modifiers -> public functions etc.)

the modifier itself is a nice addition

solidity/contracts/utility/ContractRegistryClient.sol
```diff
@@ -0,0 +1,110 @@
...
+        prevRegistry = registry;
+
+        // set the current contract-registry as the new contract-registry
+        registry = IContractRegistry(newRegistry);
+    }
+
+    /**
+      * @dev reverts back to the previous contract-registry
+    */
+    function restoreRegistry() public {
```
yudilevi: I wanna discuss this function, it can potentially be abused

solidity/contracts/converter/BancorConverterUpgrader.sol
```diff
@@ -69,17 +65,23 @@ contract BancorConverterUpgrader is IBancorConverterUpgrader, Owned, ContractIds
...
+    constructor(IContractRegistry _registry) ContractRegistryClient(_registry) public {
     }
 
     /**
       * @dev allows the owner to update the contract registry contract address
       * 
       * @param _registry   address of a contract registry contract
     */
-    function setRegistry(IContractRegistry _registry) public ownerOnly {
+    function setRegistry(IContractRegistry _registry)
```
yudilevi: I think we can probably live without this function and rely on updateRegistry if we need to.
We can also always just deploy a new upgrader and update the registry if we really have to

solidity/test/BancorNetwork.js
```diff
@@ -18,14 +17,20 @@ const ContractFeatures = artifacts.require('ContractFeatures');
 const EtherToken = artifacts.require('EtherToken');
 const TestNonStandardERC20Token = artifacts.require('TestNonStandardERC20Token');
 
+const contractFeaturesId = web3.fromAscii('ContractFeatures');
```
yudilevi: we can add these to a new helper util or just the utils we have

README.md
```diff
@@ -190,158 +190,9 @@ Triggered when a conversion between one of the convertible tokens takes place.
...
-        "reserve1": "250000000000000000",
-        "ratio2": 500000,
-        "reserve2": "1300000000000000000"
-    },
-    "priceLimitParams": {
-        "value": "6000000000"
-    }
-}
-```
+### [Operational](./solidity/operational/README.md)
```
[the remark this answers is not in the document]
barakman: NP, will fix.

solidity/contracts/BancorNetwork.sol
```diff
@@ -75,8 +72,8 @@ contract BancorNetwork is IBancorNetwork, TokenHolder, ContractIds, FeatureIds {
         public
         ownerOnly
         validAddress(_registry)
-        notThis(_registry)
     {
+        prevRegistry = registry;
```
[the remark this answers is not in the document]
barakman: That's in function `setRegistry`, which is not implemented in the `ContractRegistryClient` contract.

solidity/contracts/BancorNetwork.sol
```diff
@@ -291,7 +288,7 @@ contract BancorNetwork is IBancorNetwork, TokenHolder, ContractIds, FeatureIds {
         verifyConversionParams(_path, msg.sender, this, _signature);
 
         // verify that the destination token is BNT
-        require(_path[_path.length - 1] == registry.addressOf(ContractIds.BNT_TOKEN));
+        require(_path[_path.length - 1] == addressOf(BNT_TOKEN));
```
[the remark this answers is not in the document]
barakman: Yes, same with `modifier only`, no?

solidity/contracts/utility/ContractRegistryClient.sol
barakman: I wouldn't use `defined` here. Let's think of something else... maybe `registered`?

solidity/contracts/utility/ContractRegistryClient.sol
barakman: It allows/prevents the change to/from non-authorized callers.
In the case of `BancorX`, it's anyone who is not an owner.
In the case of `BancorConverter`, it's anyone who is neither an owner nor a manager.
This is determined by the implementation of the (internal) function `callerIsAuthorized`.

solidity/contracts/utility/ContractRegistryClient.sol
barakman: Hahaha.
No, we need to remove the `ownerOnly`, which I must have forgotten in this specific function (did it in the other two).

solidity/contracts/converter/BancorConverterUpgrader.sol
barakman: OK, so remove it from both `BancorNetwork` and `BancorConverterUpgrader`?

solidity/contracts/utility/ContractRegistryClient.sol
barakman: Both `function addressOf(bytes32 _contractName)` and `modifier only(bytes32 _contractName)` are intended for internal use by the different registry clients.
So ideally, I'd like to:
1. Keep them next to each other in the code.
2. Place them after all public functions.
However, while the function is declared `internal`, but the modifier cannot be, purely for technical reasons. Therefore, if we put them both at the top, then we will have an internal function before all public functions.
What do you think?

solidity/contracts/BancorNetwork.sol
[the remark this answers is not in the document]
yudilevi: so yes, let's get rid of setRegistry altogether

solidity/contracts/utility/ContractRegistryClient.sol
yudilevi: Yup, though from a users (docs) perspective, "disabling" means that nobody can do that, not just a list of authorized users

solidity/contracts/utility/ContractRegistryClient.sol
yudilevi: sure, sounds good

solidity/contracts/utility/ContractRegistryClient.sol
yudilevi: Gotcha

solidity/contracts/utility/ContractRegistryClient.sol
yudilevi: I think that in terms of readability, it's not ideal that you have a list of constants at the end of the file and less important to have them together.
Someone who's inheriting from the contract would need to scroll down to see the constants and it's really not standard (we don't have a header file here...)

solidity/contracts/converter/BancorConverterUpgrader.sol
yudilevi: yup

barakman: Done

solidity/contracts/utility/ContractRegistryClient.sol
barakman: Yes, the constants can go up.
My argument is with regards to `function addressOf(bytes32 _contractName)` and `modifier only(bytes32 _contractName)`.

solidity/contracts/BancorNetwork.sol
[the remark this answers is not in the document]
barakman: Done.

solidity/contracts/utility/ContractRegistryClient.sol
barakman: I believe that either `isAdminCaller` or `callerIsAdmin` would be more suitable.

solidity/contracts/utility/ContractRegistryClient.sol
yudilevi: I would put the modifier last in the list of modifiers and the function last in the list of functions.
Thoughts?

solidity/contracts/utility/ContractRegistryClient.sol
barakman: Yeah, true. Well, it seems that we'll need to revise our entire terminology in this contract.
It's also gonna be an API-breaking change, because I guess that it won't be much of an issue since all of the functionality is generally restricted to privileged users, right?

solidity/test/BancorNetwork.js
barakman: Done

solidity/contracts/utility/ContractRegistryClient.sol
barakman: That would be the same as what you've originally asked for. Subsequently, the modifier will appear before all public functions and the (internal) function will appear after all public functions.
Proceed?

solidity/contracts/utility/ContractRegistryClient.sol
barakman: Can you please check the new terminology that I've applied in this file?

solidity/contracts/utility/ContractRegistryClient.sol
barakman: Done

solidity/contracts/utility/ContractRegistryClient.sol
```diff
@@ -0,0 +1,107 @@
...
+        // verify that this function is permitted
+        require(!onlyAdmin || isAdmin());
+
+        // get the new contract-registry
+        address newRegistry = addressOf(CONTRACT_REGISTRY);
+
+        // verify that the new contract-registry is different and not zero
+        require(newRegistry != address(registry) && newRegistry != address(0));
+
+        // set the previous contract-registry as current contract-registry
```
yudilevi: generally the comments here are not 100% clear, seem backwards to me - set the previous as the current sounds like restore, set the current as the new sounds like it does nothing.

solidity/contracts/utility/ContractRegistryClient.sol
```diff
@@ -0,0 +1,107 @@
...
+    bytes32 internal constant BANCOR_GAS_PRICE_LIMIT = "BancorGasPriceLimit";
+    bytes32 internal constant BANCOR_CONVERTER_UPGRADER = "BancorConverterUpgrader";
+    bytes32 internal constant BANCOR_CONVERTER_FACTORY = "BancorConverterFactory";
+    bytes32 internal constant BNT_TOKEN = "BNTToken";
+    bytes32 internal constant BANCOR_X = "BancorX";
+    bytes32 internal constant BANCOR_X_UPGRADER = "BancorXUpgrader";
+
+    IContractRegistry public registry;      // address of the current contract-registry
+    IContractRegistry public prevRegistry;  // address of the previous contract-registry
+    bool public onlyAdmin;                  // only an administrator can change the contract-registry
```
yudilevi: onlyAdmin -> adminOnly

solidity/contracts/utility/ContractRegistryClient.sol
barakman: Can you please review again?

solidity/contracts/utility/ContractRegistryClient.sol
barakman: Done

solidity/contracts/utility/ContractRegistryClient.sol
```diff
@@ -0,0 +1,107 @@
...
+        // verify that this function is permitted
+        require(!adminOnly || isAdmin());
+
+        // get the new contract-registry
+        address newRegistry = addressOf(CONTRACT_REGISTRY);
+
+        // verify that the new contract-registry is different and not zero
+        require(newRegistry != address(registry) && newRegistry != address(0));
+
+        // update the previous contract-registry
```
yudilevi: keep a backup of the current registry before updating it?

solidity/contracts/utility/ContractRegistryClient.sol
```diff
@@ -0,0 +1,107 @@
...
+        prevRegistry = registry;
+
+        // update the current contract-registry
```
yudilevi: replace the current registry with the new one?

solidity/contracts/utility/ContractRegistryClient.sol
yudilevi: check my comments

PR #352, opened 2020-03-21
Remove ether token

barakman:


solidity/contracts/converter/BancorConverter.sol
```diff
@@ -733,11 +801,17 @@ contract BancorConverter is IBancorConverter, SmartTokenController, ContractRegi
...
+            bool isETH = reserveToken == IERC20Token(0);
+            uint256 reserveBalance = getReserveBalance(reserveToken) - (isETH ? msg.value : 0);
             uint256 reserveAmount = formula.calculateFundCost(supply, reserveBalance, totalReserveRatio, _amount);
 
             // transfer funds from the caller in the reserve token
-            ensureTransferFrom(reserveToken, msg.sender, this, reserveAmount);
+            if (isETH && msg.value > reserveAmount)
+                msg.sender.transfer(msg.value - reserveAmount);
+            else if (isETH && msg.value < reserveAmount)
+                ensureTransferFrom(IERC20Token(0xc0829421C1d260BD3cB3E0F06cfE2D52db2cE315), msg.sender, this, reserveAmount - msg.value);
```
yudilevi: if the converter has ETH reserve, it doesn't know how to handle EtherToken so it won't help much to transfer the EtherToken to the converter (the converter will just have EtherToken balance - not supposed to)
For ETH converters, there are 2 options -
1. Enough ETH was sent with the call - process it and return the remainder (done)
2. NO ETH was sent with the call (require that value == 0) - in this case do the ensureTransferFrom just like any token and then specifically for ETH converter, also call EtherToken.withdraw() to get the ETH from the EtherToken

So if some ETH was sent with the call but not enough, revert.

solidity/contracts/converter/BancorConverterUpgrader.sol
```diff
@@ -66,7 +68,8 @@ contract BancorConverterUpgrader is IBancorConverterUpgrader, ContractRegistryCl
       * 
       * @param _registry    address of a contract registry contract
     */
-    constructor(IContractRegistry _registry) ContractRegistryClient(_registry) public {
+    constructor(IContractRegistry _registry, IEtherToken _etherToken) ContractRegistryClient(_registry) public {
+        etherToken = _etherToken;
```
yudilevi: no need to change the interface, can just hardcode the ether token address here

solidity/contracts/converter/BancorConverterUpgrader.sol
```diff
@@ -230,9 +237,15 @@ contract BancorConverterUpgrader is IBancorConverterUpgrader, ContractRegistryCl
 
         for (uint16 i = 0; i < connectorTokenCount; i++) {
             address connectorAddress = _oldConverter.connectorTokens(i);
-            IERC20Token connector = IERC20Token(connectorAddress);
-            connectorBalance = connector.balanceOf(_oldConverter);
-            _oldConverter.withdrawTokens(connector, address(_newConverter), connectorBalance);
+            if (connectorAddress == address(0)) {
+                connectorBalance = etherToken.balanceOf(_oldConverter);
```
yudilevi: If the connector address is 0, it means there's no EtherToken involved anymore

solidity/contracts/converter/BancorConverterUpgrader.sol
```diff
@@ -230,9 +237,15 @@ contract BancorConverterUpgrader is IBancorConverterUpgrader, ContractRegistryCl
...
             address connectorAddress = _oldConverter.connectorTokens(i);
-            IERC20Token connector = IERC20Token(connectorAddress);
-            connectorBalance = connector.balanceOf(_oldConverter);
-            _oldConverter.withdrawTokens(connector, address(_newConverter), connectorBalance);
+            if (connectorAddress == address(0)) {
+                connectorBalance = etherToken.balanceOf(_oldConverter);
+                etherToken.withdrawTo(address(_newConverter), connectorBalance);
+            }
+            else {
+                IERC20Token connector = IERC20Token(connectorAddress);
```
yudilevi: if the connector is EtherToken, we need to withdraw the ETH from it and send it to the new converter

solidity/contracts/converter/BancorConverter.sol
```diff
@@ -368,23 +410,24 @@ contract BancorConverter is IBancorConverter, SmartTokenController, ContractRegi
...
       * 
       * @param _reserveToken    contract address of the reserve token
       * @param _depositAmount   amount of reserve-tokens received from the user
+      * @param _delta           ???
       * 
       * @return amount of supply-tokens that the user will receive
       * @return amount of supply-tokens that the user will pay as fee
     */
-    function getPurchaseReturn(IERC20Token _reserveToken, uint256 _depositAmount)
+    function getPurchaseReturn(IERC20Token _reserveToken, uint256 _depositAmount, uint256 _delta)
```
yudilevi: make it private, users can just call getReturn, simpler.
same for getSaleReturn and getCrossReserveReturn

solidity/test/BancorConverter.js
```diff
@@ -593,13 +593,13 @@ contract('BancorConverter', accounts => {
     it('should throw when attempting to get the purchase return while the converter is not active', async () => {
         let converter = await initConverter(accounts, false);
 
-        await utils.catchRevert(converter.getPurchaseReturn.call(reserveToken.address, 500));
+        await utils.catchRevert(converter.getPurchaseReturn.call(reserveToken.address, 500, 0));
```
yudilevi: see the converter - these should just be private

solidity/contracts/BancorNetwork.sol
```diff
@@ -260,6 +273,24 @@ contract BancorNetwork is IBancorNetwork, TokenHolder, ContractRegistryClient, F
...
     }
 
+    function change(
+        IBancorConverter _converter,
+        IERC20Token _fromToken,
+        IERC20Token _toToken,
+        uint256 _amount,
+        uint256 _minReturn
+    ) private returns (uint256) {
+        if (etherTokens[_fromToken] || etherTokens[_toToken]) {
```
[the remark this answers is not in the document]
barakman: @yudilevi : Yes, for saving gas. Otherwise, `isETHConverter(_converter)`, which yields an external function-call, is always invoked (i.e., not only when it needs to be).

solidity/contracts/converter/BancorConverter.sol
```diff
@@ -44,7 +44,7 @@ contract BancorConverter is IBancorConverter, SmartTokenController, ContractRegi
     /**
       * @dev version number
     */
-    uint16 public version = 27;
+    uint16 public version = 28;
```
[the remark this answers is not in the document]
barakman: @yudilevi : Actually, yes - first in 0.5.19, and then in this branch.
I believe that we would like this separation, because changes in the converter, even if technically "cosmetic", **have** taken place in 0.5.19.
So in case there are any issues, I think that we would like to be able to version-distinguish between those "cosmetic" changes and the "real" ones.

solidity/contracts/converter/BancorConverter.sol
```diff
@@ -368,23 +410,24 @@ contract BancorConverter is IBancorConverter, SmartTokenController, ContractRegi
...
-            return getCrossReserveReturn(_fromToken, _toToken, _amount);
+            return getCrossReserveReturn(_fromToken, _toToken, _amount, 0);
     }
 
     /**
       * @dev calculates the expected return of buying with a given amount of tokens
       * 
       * @param _reserveToken    contract address of the reserve token
       * @param _depositAmount   amount of reserve-tokens received from the user
+      * @param _delta           ???
```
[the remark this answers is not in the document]
barakman: @yudilevi : Yes, I've used that `???` as a place-holder.

solidity/contracts/converter/BancorConverter.sol
```diff
@@ -440,11 +483,12 @@ contract BancorConverter is IBancorConverter, SmartTokenController, ContractRegi
       * @param _fromReserveToken    contract address of the reserve token to convert from
       * @param _toReserveToken      contract address of the reserve token to convert to
       * @param _amount              amount of tokens received from the user
+      * @param _delta               ???
```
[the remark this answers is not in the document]
barakman: @yudilevi : same as above...

solidity/contracts/converter/BancorConverterUpgrader.sol
barakman: @yudilevi : It would be very hard to properly test this contract without it (I can add an only-owner setter, but I prefer it this way). And this is actually a crucial test in our system (been working on it for about two days now).
My thought was that changing the constructor isn't an API-breaking change. Am I wrong?

solidity/contracts/converter/BancorConverterUpgrader.sol
barakman: @yudilevi : Note the flow:
1. In `copyConnectors`, we need to check `if (connectorAddress == address(etherToken))`. I've extended it to `if (connectorAddress == address(0) || connectorAddress == address(etherToken))`, so that it will be usable also for upgrading our new ETH converter in the future. Otherwise, we'd have to redeploy the upgrader in order to do that.
2. In `transferConnectorsBalances`, we've already added a zero reserve representing ETH, so we just need to withdraw the funds from Ether token into the converter.

solidity/contracts/converter/BancorConverterUpgrader.sol
barakman: @yudilevi : The connector is not EtherToken at this point, it is the zero address.

solidity/contracts/converter/BancorConverter.sol
barakman: @yudilevi : I did. But then I found out that it breaks some of our tests. And since I consider these tests crucial for the current task at hand, I decided to leave those functions `public` for now.
Do you want me to make them `private`, and then add `public` functions which call the `private` ones with `delta = 0`?
I suggest we just leave it as is and take care of it later, after we know for sure that everything is working correctly.

solidity/test/BancorConverter.js
barakman: @yudilevi : See the comment above.

solidity/contracts/BancorNetwork.sol
```diff
@@ -64,6 +64,13 @@ contract BancorNetwork is IBancorNetwork, TokenHolder, ContractRegistryClient, F
       * @param _registry    address of a contract registry contract
     */
     constructor(IContractRegistry _registry) ContractRegistryClient(_registry) public {
+        etherTokens[address(0)] = true;
```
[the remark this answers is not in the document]
barakman: @yudilevi : I initially indeed removed that function, though no need to add the EtherToken address hard-coded as far as I understand. In any case, I then decided to leave it for now. Can't remember why, it might have to do with breaking some of our existing tests. I'll look into it...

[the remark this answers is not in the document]
yudilevi: if we remove that function, we'll need to hardcode the ether token address, I don't see a workaround, but that's fine. We can just define both ether token and address(0) in the ctor.

solidity/contracts/converter/BancorConverter.sol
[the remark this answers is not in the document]
yudilevi: We're going to merge them together with 0.5.19 into master so it's redundant - if we increase it again, there won't ever be 27 in master.

solidity/contracts/converter/BancorConverter.sol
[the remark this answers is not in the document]
yudilevi: so change to my desc?

solidity/contracts/converter/BancorConverterUpgrader.sol
yudilevi: yeah if it makes testing easier, sure - but how are we gonna test the converter that has the ether token address hardcoded in it?

solidity/contracts/converter/BancorConverter.sol
yudilevi: no, I want to make the private and remove the tests, or change the tests so that the will call these functions indirectly

solidity/contracts/converter/BancorConverterUpgrader.sol
```diff
@@ -195,11 +199,21 @@ contract BancorConverterUpgrader is IBancorConverterUpgrader, ContractRegistryCl
...
-            IERC20Token connectorToken = IERC20Token(connectorAddress);
-            _newConverter.addConnector(connectorToken, weight, isVirtualBalanceEnabled);
-
-            if (isVirtualBalanceEnabled)
-                _newConverter.updateConnector(connectorToken, weight, isVirtualBalanceEnabled, virtualBalance);
+            // Ether reserve
+            if (connectorAddress == address(0)) {
+                _newConverter.addETHReserve(weight);
+            }
+            // Ether reserve token
```
yudilevi: `Ether reserve token` ->`EtherToken reserve`

solidity/contracts/converter/BancorConverterUpgrader.sol
```diff
@@ -195,11 +199,21 @@ contract BancorConverterUpgrader is IBancorConverterUpgrader, ContractRegistryCl
...
+            else if (connectorAddress == address(etherToken)) {
```
yudilevi: can just handle ETH reserve and EtherToken reserve in the same block here - exact same implementation

solidity/contracts/converter/BancorConverterUpgrader.sol
```diff
@@ -195,11 +199,21 @@ contract BancorConverterUpgrader is IBancorConverterUpgrader, ContractRegistryCl
...
+                _newConverter.addETHReserve(weight);
+            }
+            // ERC20 reserve token
+            else {
+                IERC20Token connectorToken = IERC20Token(connectorAddress);
```
yudilevi: virtual balance isn't handled for EtherToken / ETH - it should be a separate section - first add the reserve (and distinguish between ERC20, ETH and EtherToken, then add virtual balance if exists regardless of the reserve type

solidity/contracts/converter/BancorConverterUpgrader.sol
```diff
@@ -230,9 +244,22 @@ contract BancorConverterUpgrader is IBancorConverterUpgrader, ContractRegistryCl
...
         for (uint16 i = 0; i < connectorTokenCount; i++) {
             address connectorAddress = _oldConverter.connectorTokens(i);
-            IERC20Token connector = IERC20Token(connectorAddress);
-            connectorBalance = connector.balanceOf(_oldConverter);
-            _oldConverter.withdrawTokens(connector, address(_newConverter), connectorBalance);
+            // Ether reserve
+            if (connectorAddress == address(0)) {
+                _oldConverter.withdrawETH(address(_newConverter));
+            }
+            // Ether reserve token
```
yudilevi: `Ether reserve token` -> `EtherToken reserve`

solidity/contracts/converter/BancorConverterUpgrader.sol
barakman: @yudilevi : Correct, but due to your previous comment on this section, as well as due to the `if`/`else` structure of the other function (`transferConnectorsBalances`), I've decided to write it this way for better clarity. This should yield very little impact on deployment gas-cost, and no impact at all on runtime gas-cost.

solidity/contracts/BancorNetwork.sol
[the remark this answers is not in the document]
barakman: @yudilevi : So this one is done?

solidity/contracts/converter/BancorConverter.sol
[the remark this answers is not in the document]
barakman: @yudilevi : done.

solidity/contracts/converter/BancorConverter.sol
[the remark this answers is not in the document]
barakman: @yudilevi : Changed to `amount deposited into the reserve prior to calling this function`.

solidity/contracts/BancorNetwork.sol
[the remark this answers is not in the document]
yudilevi: yeah I think we're good here for now

solidity/contracts/BancorNetwork.sol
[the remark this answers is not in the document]
yudilevi: ok

solidity/contracts/converter/BancorConverterUpgrader.sol
yudilevi: maybe reconsider after the `else` block is fixed to also cover ETH reserve

solidity/contracts/converter/BancorConverter.sol
barakman: @yudilevi : Since we're planning to remove the `delta` solution, perhaps we should keep these tests?

solidity/contracts/converter/BancorConverterUpgrader.sol
barakman: @yudilevi : done.

solidity/contracts/converter/BancorConverterUpgrader.sol
barakman: @yudilevi : We have 66 occurrences of `reserve token` and 0 occurrences of `token reserve` in our contract documentation. Are you sure?

solidity/contracts/converter/BancorConverterUpgrader.sol
barakman: @yudilevi : Sorry, I don't get it.

solidity/contracts/converter/BancorConverter.sol
```diff
@@ -479,7 +523,7 @@ contract BancorConverter is IBancorConverter, SmartTokenController, ContractRegi
       * @return amount of tokens received (in units of the target token)
     */
     function convertInternal(IERC20Token _fromToken, IERC20Token _toToken, uint256 _amount, uint256 _minReturn)
-        internal
+        public
```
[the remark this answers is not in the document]
barakman: @yudilevi : Good catch! It probably happened when I changed the `getXXXReturn` functions from `public` to `internal`, and then back to `public` in order to avoid breaking the tests (so when I changed them back to `public`, I must have changed that one to `public` too).
Fixed.

solidity/contracts/converter/BancorConverterUpgrader.sol
yudilevi: Nm, I think it's fine - the 2nd case is temporary anyway

solidity/contracts/converter/BancorConverter.sol
barakman: @yudilevi : Applied in 0.5.19 (since this is not directly related to removing ether-token), and then pulled to this branch.

solidity/contracts/converter/BancorConverterUpgrader.sol
barakman: @yudilevi : We have 66 occurrences of `reserve token` and 0 occurrences of `token reserve` in our contract documentation. Are you sure?

solidity/contracts/converter/BancorConverter.sol
yudilevi: yeah I'd rather limit the API to getReturn - gives more flexibility and less dependency on external integrations

solidity/contracts/converter/BancorConverterUpgrader.sol
yudilevi: Not critical, going out soon anyway

solidity/contracts/converter/BancorConverterUpgrader.sol
yudilevi: As above - not critical

solidity/contracts/converter/BancorConverter.sol
barakman: @yudilevi : By `EtherToken.withdraw()`, do you mean `EtherToken.withdrawTo(this, reserveAmount)` (or simply `EtherToken.withdraw(reserveAmount)`)?

barakman: @yudilevi : By `etherToken.withdraw()`, do you mean `etherToken.withdrawTo(this, reserveAmount)` (or simply `etherToken.withdraw(reserveAmount)`)?

barakman: @yudilevi : Done.

PR #354, opened 2020-04-07
Gas reduction

yudilevi:


solidity/test/BancorNetwork.js
```diff
@@ -283,7 +287,7 @@ contract('BancorNetwork', accounts => {
 
     it('verifies that convert transfers the converted amount correctly', async () => {
         let balanceBeforeTransfer = await smartToken1.balanceOf.call(accounts[1]);
-        await bancorNetwork.convert(smartToken1BuyPath, 10000, 1, { from: accounts[1], value: 10000 });
+        let res = await bancorNetwork.convert(smartToken1BuyPath, 10000, 1, { from: accounts[1], value: 10000 });
```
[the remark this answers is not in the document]
yudilevi: I guess... removed

solidity/test/XConversions.js
```diff
@@ -7,7 +7,6 @@ const ContractRegistryClient = require('./helpers/ContractRegistryClient');
 const BancorConverter = artifacts.require('BancorConverter');
 const BancorX = artifacts.require('BancorX');
 const SmartToken = artifacts.require('SmartToken');
-const EtherToken = artifacts.require('EtherToken');
```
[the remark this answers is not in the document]
yudilevi: yeah I think so

solidity/contracts/BancorNetwork.sol
```diff
@@ -122,35 +131,24 @@ contract BancorNetwork is IBancorNetwork, TokenHolder, ContractRegistryClient, F
...
-            else
-                IEtherToken(toToken).withdrawTo(_for, amount);
-        }
-        else {
-            uint256 prevBalance = toToken.balanceOf(_for);
-            INonStandardERC20(toToken).transfer(_for, amount);
-            uint256 postBalance = toToken.balanceOf(_for);
-            require(postBalance > prevBalance);
-        }
+        ConversionStep[] memory data = createConversionData(_path, _minReturn, _for, affiliateFeeEnabled);
```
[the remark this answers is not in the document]
yudilevi: it wouldn't be correct, since conversion data can actually be composed of more stuff other than the steps, right now it's just the steps but might change in the future.
in general the argument is the concept of all the data needed for conversion, whatever it may be.

solidity/contracts/BancorNetwork.sol
```diff
@@ -450,6 +405,191 @@ contract BancorNetwork is IBancorNetwork, TokenHolder, ContractRegistryClient, F
...
     }
 
+    /**
+      * @dev validates msg.value and prepares the conversion source tokens for the conversion
+      * 
+      * @param _sourceToken source token of the first conversion step
+      * @param _smartToken  smart token of the first conversion step
+      * @param _amount      amount to convert from, in the source token
+    */
+    function handleSourceTokens(IERC20Token _sourceToken, ISmartToken _smartToken, uint256 _amount) private {
```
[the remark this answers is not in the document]
yudilevi: Hmm.. I meant to reference the tokens themselves, not the actual ERC20 contract but both could work I guess... thoughts?

solidity/contracts/converter/BancorConverter.sol
```diff
@@ -340,6 +356,7 @@ contract BancorConverter is IBancorConverter, SmartTokenController, ContractRegi
         reserves[address(0)].ratio = _ratio;
         reserves[address(0)].virtualBalance = 0;
         reserves[address(0)].isSet = true;
+        reserveBalances[address(0)] = 0;
```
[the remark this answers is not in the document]
yudilevi: Yup, that was the reason.
I'm thinking to add external function for all the data there and then, once the internal structure is not used externally, move the stored balance in there.

solidity/contracts/converter/BancorConverter.sol
```diff
@@ -812,7 +829,7 @@ contract BancorConverter is IBancorConverter, SmartTokenController, ContractRegi
         for (uint16 i = 0; i < reserveTokens.length; i++) {
             IERC20Token reserveToken = reserveTokens[i];
             bool isETH = reserveToken == IERC20Token(0);
```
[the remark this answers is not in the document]
yudilevi: done

solidity/contracts/converter/BancorConverter.sol
```diff
@@ -864,17 +884,31 @@ contract BancorConverter is IBancorConverter, SmartTokenController, ContractRegi
...
             emit PriceDataUpdate(reserveToken, supply - _amount, reserveBalance - reserveAmount, reserves[reserveToken].ratio);
         }
     }
 
+    /**	
+      * @dev syncs the stored reserve balance for a given reserve with the real reserve balance
+      *
+      * @param _reserveToken    address of the reserve token, or address(0) for ETH reserve
+    */
+    function syncReserveBalance(IERC20Token _reserveToken) internal validReserve(_reserveToken) {
```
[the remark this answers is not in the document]
yudilevi: only way I see is to add another function but the cost for checking if it's ETH reserve is really low, there's no storage access, so I imagine a simple check in memory is insignificant

solidity/contracts/converter/BancorConverter.sol
```diff
@@ -47,15 +41,16 @@ contract BancorConverter is IBancorConverter, SmartTokenController, ContractRegi
...
-                                                    // the total reserve ratio above 100%
-    uint32 public maxConversionFee = 0;             // maximum conversion fee for the lifetime of the contract,
-                                                    // represented in ppm, 0...1000000 (0 = no fee, 100 = 0.01%, 1000000 = 100%)
-    uint32 public conversionFee = 0;                // current conversion fee, represented in ppm, 0...maxConversionFee
-    bool public conversionsEnabled = true;          // deprecated, backward compatibility
+    IWhitelist public conversionWhitelist;                  // whitelist contract with list of addresses that are allowed to use the converter
+    IERC20Token[] public reserveTokens;                     // ERC20 standard token addresses (prior version 17, use 'connectorTokens' instead)
+    mapping (address => Reserve) public reserves;           // reserve token addresses -> reserve data (prior version 17, use 'connectors' instead)
+    mapping (address => uint256) private reserveBalances;   // reserve token addresses -> reserve balance
```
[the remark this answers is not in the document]
yudilevi: Since it's not supposed to be public information. The getter function is mostly there in order to be able to override it as well as support a cleaner implementation once we have virtual balance back in the contract.
But users aren't supposed to query the storage balance by accessing that var.

solidity/contracts/converter/BancorConverter.sol
```diff
@@ -402,9 +419,7 @@ contract BancorConverter is IBancorConverter, SmartTokenController, ContractRegi
         validReserve(_reserveToken)
         returns (uint256)
     {
-        if (_reserveToken == IERC20Token(0))
-            return address(this).balance;
-        return _reserveToken.balanceOf(this);
+        return reserveBalances[_reserveToken];
```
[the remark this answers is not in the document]
yudilevi: See my answer about re. getReserveBalance.

Other than that, I imagine we can remove the check from some of these functions but how are we gonna handle getPurchaseReturn? That function has to be able to indicate if the input is invalid, it can't just return 0 since that's a valid return amount (theoretically speaking).

solidity/test/XConversions.js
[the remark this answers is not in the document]
barakman: So we'll need to revert this file then, right?

solidity/contracts/BancorNetwork.sol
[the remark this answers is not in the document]
barakman: In that case, the current struct name is not correct.
Either both reflect conversion step, or both reflect something more general.
It doesn't make sense to use different names IMO.
The variable name should somehow imply about its type here IMO.

solidity/contracts/converter/BancorConverter.sol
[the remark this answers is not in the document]
barakman: Whoever calls `getPurchaseReturn`, should verify beforehand that the reserve is valid. I imagine that's already the case anyway (either that, or the reserve is known to be valid, as is the case when this function is called from one of our conversion functions (`convertInternal` or whatever)).

solidity/contracts/converter/BancorConverter.sol
```diff
@@ -237,6 +244,10 @@ contract BancorConverter is IBancorConverter, SmartTokenController, ContractRegi
         totalSupplyGreaterThanZeroOnly
     {
         super.acceptTokenOwnership();
+
+        // sync all reserve balances	
+        for (uint16 i = 0; i < reserveTokens.length; i++)
```
barakman, line 249: @yudilevi:
No point using `uint16` here (`uint256` is good enough, and probably even slightly better in terms of performance).

solidity/contracts/converter/BancorConverter.sol
```diff
@@ -590,11 +606,13 @@ contract BancorConverter is IBancorConverter, SmartTokenController, ContractRegi
...
       * @param _minReturn       if the conversion results in an amount smaller the minimum return - it is cancelled, must be nonzero
+      * @param _beneficiary     wallet to receive the conversion result
       * 
       * @return amount of tokens received (in units of the reserve token)
     */
-    function sell(IERC20Token _reserveToken, uint256 _sellAmount, uint256 _minReturn) internal returns (uint256) {
-        require(_sellAmount <= token.balanceOf(msg.sender)); // validate input
+    function sell(IERC20Token _reserveToken, uint256 _sellAmount, uint256 _minReturn, address _beneficiary) internal returns (uint256) {
+        // ensure that the input amount was already deposited
+        require(_sellAmount <= token.balanceOf(this));
```
barakman, line 615: @yudilevi:
Why have you changed `msg.sender` to `this` here?

solidity/contracts/converter/BancorConverter.sol
```diff
@@ -606,16 +624,17 @@ contract BancorConverter is IBancorConverter, SmartTokenController, ContractRegi
         uint256 reserveBalance = getReserveBalance(_reserveToken);
         assert(amount < reserveBalance || (amount == reserveBalance && _sellAmount == tokenSupply));
 
-        // destroy _sellAmount from the caller's balance in the smart token
-        token.destroy(msg.sender, _sellAmount);
+        // destroy _sellAmount from the converter balance in the smart token
+        token.destroy(this, _sellAmount);
```
barakman, line 628: @yudilevi:
Why have you changed `msg.sender` to `this` here?

solidity/test/XConversions.js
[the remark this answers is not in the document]
yudilevi: oh I think we can live without it in the XConversions tests, no need to still have it all over the system but just where necessary

solidity/contracts/BancorNetwork.sol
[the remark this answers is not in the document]
yudilevi: I don't see why - it's abstraction vs. implementation - the abstract name is 'data' and the implementation is whatever we see as the current requirements are.
I don't have a problem with changing the type if we decide to add more stuff to the data and create a new structure, I just want to make sure it involves minimal code without adding unnecessary overhead now and the current implementation accomplishes both.

solidity/contracts/converter/BancorConverter.sol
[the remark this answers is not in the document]
yudilevi: `getPurchaseReturn` is called from two places today -
- `getReturn`
- `buy`

Neither of these verify the reserve address - they actually rely on `getPurchaseReturn` for that - do you suggest we remove it from the internal function and put it externally in both functions?

solidity/contracts/BancorNetwork.sol
```diff
@@ -34,10 +39,20 @@ contract BancorNetwork is IBancorNetwork, TokenHolder, ContractRegistryClient, F
     uint256 private constant CONVERSION_FEE_RESOLUTION = 1000000;
     uint256 private constant AFFILIATE_FEE_RESOLUTION = 1000000;
 
+    struct ConversionStep {
+        IBancorConverter converter;
+        ISmartToken smartToken;
+        IERC20Token sourceToken;
+        IERC20Token targetToken;
+        uint256 minReturn;
```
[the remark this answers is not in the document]
yudilevi: About that - we didn't discuss it yet cause I tried not to change it until we finish the initial review but I'm thinking to remove the minimum return completely from the converters and just check it in the network.
If you don't see an issue with it, it will make this point irrelevant since the value will be removed from the ConversionStep struct.

solidity/contracts/BancorNetwork.sol
```diff
@@ -450,6 +405,191 @@ contract BancorNetwork is IBancorNetwork, TokenHolder, ContractRegistryClient, F
...
+    }
+
+    /**
+      * @dev handles the conversion target tokens if the network still holds them at the end of the conversion
+      * 
+      * @param _data        conversion data, see ConversionStep struct above
+      * @param _amount      conversion return amount, in the target token
+      * @param _beneficiary wallet to receive the conversion result
+    */
+    function handleTargetTokens(ConversionStep[] _data, uint256 _amount, address _beneficiary) private {
```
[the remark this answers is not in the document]
yudilevi: Will deal with this once we conclude the `handleSourceToken` function

solidity/contracts/BancorNetwork.sol
```diff
@@ -450,6 +405,191 @@ contract BancorNetwork is IBancorNetwork, TokenHolder, ContractRegistryClient, F
...
+        ConversionStep memory stepData = _data[_data.length - 1];
+
+        // network contract doesn't hold the tokens, do nothing
+        if (stepData.beneficiary != address(this))
```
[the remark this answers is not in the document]
yudilevi: Each step has a beneficiary for that specific step - who the tokens go to next.
The main point here after the conversion is complete is to check whether the network still has the tokens - and thus, needs to transfer it to the conversion beneficiary - or if the tokens were already sent there by the converter.
We do this by checking if the network is the previous beneficiary. If it is - it needs to sends tokens out.

solidity/contracts/BancorNetwork.sol
```diff
@@ -10,9 +10,14 @@ import './utility/interfaces/IContractFeatures.sol';
 import './utility/interfaces/IWhitelist.sol';
```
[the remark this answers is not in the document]
yudilevi: And replace it with what? We still need to verify path validity.

solidity/contracts/converter/BancorConverter.sol
yudilevi: Does the EVM need to work harder to "limit" it to 16 bits?

solidity/contracts/converter/BancorConverter.sol
yudilevi: I thought it would be more consistent that the converter handle the smart token the same way it handles other tokens.
Plus it's more "correct" in a way that the converter destroys its own holdings instead of destroying the tokens held by another account.
We could in the future remove the option for smart tokens to get destroyed by the converter entirely and only allow the holder of smart tokens to destroy their own tokens.

solidity/contracts/converter/BancorConverter.sol
yudilevi: Same as above

solidity/contracts/BancorNetwork.sol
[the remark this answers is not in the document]
barakman: @yudilevi:
Why limit the path length?
We can let the block gas-limit take care of that, when the transaction becomes too large.

solidity/contracts/converter/BancorConverter.sol
barakman: @yudilevi:
1. The type of array.length is `uint256`
2. AFAIK, it is generally recommended to use `uint256`. Not sure how critical this is for local variables, but I think that we should limit the usage of `uint` types smaller than `uint256` only for when it's absolutely necessary (i.e., packed structures), so that we never have any doubts about possible overflows, etc (even if it is obvious that the total number of reserves will never grow beyond 65535 in this case).

solidity/contracts/converter/BancorConverter.sol
barakman: @yudilevi:
But here `token.balanceOf(msg.sender)` and `token.balanceOf(this)` are two different things.

solidity/contracts/BancorNetwork.sol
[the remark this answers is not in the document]
barakman: @yudilevi: `handleSourceTokenAmount`?

solidity/contracts/BancorNetwork.sol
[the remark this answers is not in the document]
barakman: @yudilevi: `handleTargetTokenAmount`?

solidity/contracts/converter/BancorConverter.sol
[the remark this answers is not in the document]
barakman: @yudilevi:
Function `getReturn` is read-only. I don't see a problem with it returning 0, because that's the expected return when a reserve is not valid.
Function `buy` can (and maybe should) validate the reserve beforehand, and IMO it should rely on `getPurchaseReturn` to do that. BTW, it doesn't really matter if this function doesn't validate the reserve, because it does `require(amount != 0)` immediately after calling `getPurchaseReturn`.
But the issue is general not only to `getPurchaseReturn`, but also with several other functions as I have mentioned above.
IMO, we should validate the reserve only in write functions (and even there, only if it seems necessary).

solidity/contracts/BancorNetwork.sol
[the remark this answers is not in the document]
barakman: @yudilevi:
I think we should keep it, because it allows us to remove the check of last-index in every iteration on the conversion path, which increases the overall gas consumption, as well as the compiler-stack usage as you may recall.

solidity/contracts/BancorNetwork.sol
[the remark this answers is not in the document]
yudilevi: I'll just rename to `handleSource/TargetToken`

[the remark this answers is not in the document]
yudilevi: done

solidity/contracts/BancorNetwork.sol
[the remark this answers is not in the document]
yudilevi: It will allow us to remove it anyway because we'll just do the check in the network after we're done with the loop

solidity/contracts/BancorNetwork.sol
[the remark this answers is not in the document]
yudilevi: renamed to `handleTargetToken`

solidity/contracts/BancorNetwork.sol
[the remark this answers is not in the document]
yudilevi: done

solidity/contracts/converter/BancorConverter.sol
yudilevi: updated, though keep in mind reserve/connectorTokenCount functions still return uint16

solidity/contracts/converter/BancorConverter.sol
yudilevi: because we don't destroy the sender balance anymore - we destroy self balance or fail if it's not already deposited

solidity/contracts/converter/BancorConverter.sol
barakman: Right... well, we can't really change that one now (without breaking the API)...

yudilevi: right

solidity/contracts/BancorNetwork.sol
```diff
@@ -218,79 +215,62 @@ contract BancorNetwork is IBancorNetwork, TokenHolder, ContractRegistryClient, F
...
-        else {
-            require(0 < _affiliateFee && _affiliateFee <= maxAffiliateFee);
-            bntToken = addressOf(BNT_TOKEN);
-        }
 
-        // iterate over the conversion path
-        for (uint256 i = 2; i <= lastIndex; i += 2) {
-            IBancorConverter converter = IBancorConverter(ISmartToken(_path[i - 1]).owner());
+        // iterate over the conversion data
+        for (uint8 i = 0; i < _data.length; i++) {
```
barakman, line 235: `uint8` ==> `uint256`

solidity/contracts/BancorNetwork.sol
```diff
@@ -450,6 +405,191 @@ contract BancorNetwork is IBancorNetwork, TokenHolder, ContractRegistryClient, F
...
+      * 
+      * @return cached conversion data to be ingested later on by the conversion flow
+    */
+    function createConversionData(IERC20Token[] _conversionPath, uint256 _minReturn, address _beneficiary, bool _affiliateFeeEnabled) private view returns (ConversionStep[]) {
+        ConversionStep[] memory data = new ConversionStep[](_conversionPath.length / 2);
+
+        bool affiliateFeeProcessed = false;
+        address bntToken = addressOf(BNT_TOKEN);
+        // iterate the conversion path and create the conversion data for each step
+        uint8 i;
```
barakman, line 506: `uint8` ==> `uint256`

solidity/contracts/BancorNetwork.sol
yudilevi: done

solidity/contracts/BancorNetwork.sol
yudilevi: done

solidity/contracts/BancorNetwork.sol
```diff
@@ -10,9 +10,14 @@ import './utility/interfaces/IContractFeatures.sol';
 import './utility/interfaces/IWhitelist.sol';
 import './token/interfaces/IEtherToken.sol';
 import './token/interfaces/ISmartToken.sol';
-import './token/interfaces/INonStandardERC20.sol';
 import './bancorx/interfaces/IBancorX.sol';
 
+// interface of older converters for backward compatibility
+contract ILegacyBancorConverter is IBancorConverter {
+    function change(IERC20Token _fromToken, IERC20Token _toToken, uint256 _amount, uint256 _minReturn) public payable returns (uint256);
```
barakman, line 17: This function is not `payable` in converter version prior to the current one.

solidity/contracts/BancorNetwork.sol
```diff
@@ -450,6 +405,191 @@ contract BancorNetwork is IBancorNetwork, TokenHolder, ContractRegistryClient, F
...
+        // network contract doesn't hold the tokens, do nothing
+        if (stepData.beneficiary != address(this))
+            return;
+
+        IERC20Token targetToken = stepData.targetToken;
+
+        // ETH / EtherToken
+        if (etherTokens[targetToken]) {
+            // newer converter should send ETH directly to the beneficiary
+            assert(!stepData.isV27OrHigherConverter);
```
barakman, line 479: I want to make sure that the assertion here is right:
If the user inputs a path which starts with EtherToken, then the owner of this EtherToken will most definitely be an older converter. Is that corrcet?
Or can the owner be a new converter?
In which case, even if you still don't want to support this scenario, you need to use `require`, because it's a legitimate user-input (not an "unreachable" scenario).

solidity/contracts/BancorNetwork.sol
```diff
@@ -450,6 +405,191 @@ contract BancorNetwork is IBancorNetwork, TokenHolder, ContractRegistryClient, F
...
+    function handleTargetToken(ConversionStep[] _data, uint256 _amount, address _beneficiary) private {
+        ConversionStep memory stepData = _data[_data.length - 1];
+
+        // network contract doesn't hold the tokens, do nothing
+        if (stepData.beneficiary != address(this))
+            return;
+
+        IERC20Token targetToken = stepData.targetToken;
+
+        // ETH / EtherToken
```
barakman, line 476: I don't think that ETH is properly supported here (as the comment `ETH / EtherToken` implies).
I'm assuming that you're counting on `if (etherTokens[targetToken])` to hold true when `targetToken == address(0)`.
Is that correct?
If yes, then calling `IEtherToken(targetToken).withdrawTo(...)` will revert.
If no, then calling `safeTransfer(targetToken, ...)` will revert.

solidity/contracts/BancorNetwork.sol
```diff
@@ -450,6 +405,191 @@ contract BancorNetwork is IBancorNetwork, TokenHolder, ContractRegistryClient, F
...
+        else
+            require(msg.value == 0);
+
+        IBancorConverter firstConverter = IBancorConverter(_smartToken.owner());
+        bool isNewerConverter = isV27OrHigherConverter(firstConverter);
+
+        // ETH / EtherToken
+        if (isSourceETH) {
+            // handle ETH
+            if (msg.value > 0) {
```
barakman, line 430: I think that this logic is wrong.
Instead of if/else on `msg.value > 0`, we should do if/else on `isNewerConverter`.
It will subsequently change the entire implementation of the current if/else clause.
Also, what if `_sourceToken == address(0)`?

solidity/contracts/BancorNetwork.sol
```diff
@@ -450,6 +405,191 @@ contract BancorNetwork is IBancorNetwork, TokenHolder, ContractRegistryClient, F
...
+    function handleSourceToken(IERC20Token _sourceToken, ISmartToken _smartToken, uint256 _amount) private {
+        bool isSourceETH = etherTokens[_sourceToken];
+
+        // validate msg.value
+        if (msg.value > 0)
+            require(msg.value == _amount && isSourceETH);
+        else
+            require(msg.value == 0);
+
+        IBancorConverter firstConverter = IBancorConverter(_smartToken.owner());
```
barakman, line 424: We don't need to execute this in all scenarios.
You can just embed `IBancorConverter(_smartToken.owner())` in each one of the two cases where it is actually needed.

solidity/contracts/converter/BancorConverter.sol
```diff
@@ -183,6 +178,18 @@ contract BancorConverter is IBancorConverter, SmartTokenController, ContractRegi
     function withdrawETH(address _to) public ownerOnly only(BANCOR_CONVERTER_UPGRADER) {
         require(hasETHReserve());
         _to.transfer(address(this).balance);
+
+        // sync the ETH reserve balance	
+        syncReserveBalance(IERC20Token(0));
```
barakman, line 183: May as well just do `reserveBalances[address(0)] = address(this).balance;`.

solidity/contracts/converter/BancorConverter.sol
```diff
@@ -554,26 +569,27 @@ contract BancorConverter is IBancorConverter, SmartTokenController, ContractRegi
...
-            require(msg.value == 0);
-            ensureTransferFrom(_reserveToken, msg.sender, this, _depositAmount);
-        }
+        else
+            require(msg.value == 0 && _reserveToken.balanceOf(this).sub(getReserveBalance(_reserveToken)) >= _depositAmount);
 
-        // issue new funds to the caller in the smart token
-        token.issue(msg.sender, amount);
+        // sync the reserve balance
+        syncReserveBalance(_reserveToken);
```
barakman, line 589: May as well embed the code of function `syncReserveBalance` into the if/else clause above.

solidity/contracts/converter/BancorConverter.sol
```diff
@@ -646,22 +666,21 @@ contract BancorConverter is IBancorConverter, SmartTokenController, ContractRegi
...
 
-        // transfer funds to the caller in the to reserve token
-        if (_toToken == IERC20Token(0)) {
-            msg.sender.transfer(amount);
-        }
-        else {
-            ensureTransferFrom(_toToken, this, msg.sender, amount);
-        }
+        // sync the reserve balances
+        syncReserveBalance(_fromToken);
```
barakman, line 676: May as well embed the code of function `syncReserveBalance` into the if/else clause above.

solidity/contracts/converter/BancorConverter.sol
```diff
@@ -809,27 +826,29 @@ contract BancorConverter is IBancorConverter, SmartTokenController, ContractRegi
...
                     etherToken.withdraw(reserveAmount);
                 }
             }
             else {
-                ensureTransferFrom(reserveToken, msg.sender, this, reserveAmount);
+                safeTransferFrom(reserveToken, msg.sender, this, reserveAmount);
             }
 
+            // sync the reserve balance	
+            syncReserveBalance(reserveToken);
```
barakman, line 850: May as well embed the code of function `syncReserveBalance` into the if/else clause above.

solidity/contracts/BancorNetwork.sol
```diff
@@ -218,79 +215,62 @@ contract BancorNetwork is IBancorNetwork, TokenHolder, ContractRegistryClient, F
...
+            if (stepData.processAffiliateFee) {
                 uint256 affiliateAmount = toAmount.mul(_affiliateFee).div(AFFILIATE_FEE_RESOLUTION);
-                require(_path[i].transfer(_affiliateAccount, affiliateAmount));
+                require(stepData.targetToken.transfer(_affiliateAccount, affiliateAmount));
                 toAmount -= affiliateAmount;
-                bntToken = address(~0);
             }
 
-            emit Conversion(_path[i - 1], _path[i - 2], _path[i], fromAmount, toAmount, msg.sender);
+            emit Conversion(stepData.smartToken, stepData.sourceToken, stepData.targetToken, fromAmount, toAmount, msg.sender);
```
barakman, line 267: I know that you've heard me saying this before, but consider removing this event because it contains redundant information (already provided in the `BancorConverter` layer), which is also partial information (does not include the fee).
Another incentive for removing this event, is the fact that it is emitted inside the loop, which can actually save quite a bit of gas.

solidity/contracts/BancorNetwork.sol
```diff
@@ -450,6 +405,191 @@ contract BancorNetwork is IBancorNetwork, TokenHolder, ContractRegistryClient, F
...
+        bool affiliateFeeProcessed = false;
+        address bntToken = addressOf(BNT_TOKEN);
+        // iterate the conversion path and create the conversion data for each step
+        uint256 i;
+        for (i = 0; i < _conversionPath.length - 1; i += 2) {
+            ISmartToken smartToken = ISmartToken(_conversionPath[i + 1]);
+            IBancorConverter converter = IBancorConverter(smartToken.owner());
+            IERC20Token targetToken = _conversionPath[i + 2];
+
+            // check if the affiliate fee should be processed in this step
```
barakman, line 512: I would change this as follows:
1. Outside the loop, declare `uint256 affiliateFeeCount = 0;`
2. Inside the loop, do:
```
if (_affiliateFeeEnabled && targetToken == bntToken)
    affiliateFeeCount += 1;
```
And then:
```
processAffiliateFee: affiliateFeeCount == 1
```
Assuming that the BNT token can occur at most once in any given path, it would yield the same efficiency (gas-cost) with a cleaner code which is easier to understand. In fact, even if the BNT token somehow appears more than once in some input path, it would still yield more or less the same efficiency.

solidity/contracts/BancorNetwork.sol
yudilevi: removed the `payable` modifier

solidity/contracts/BancorNetwork.sol
```diff
@@ -116,41 +125,30 @@ contract BancorNetwork is IBancorNetwork, TokenHolder, ContractRegistryClient, F
       * @return tokens issued in return
     */
     function convertFor2(IERC20Token[] _path, uint256 _amount, uint256 _minReturn, address _for, address _affiliateAccount, uint256 _affiliateFee) public payable returns (uint256) {
```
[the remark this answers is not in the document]
yudilevi: it's validated at the beginning of `handleSourceToken`

solidity/contracts/BancorNetwork.sol
```diff
@@ -186,28 +184,27 @@ contract BancorNetwork is IBancorNetwork, TokenHolder, ContractRegistryClient, F
         payable
```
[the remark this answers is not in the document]
yudilevi: see my comment on the same issue in `convert2`

solidity/contracts/BancorNetwork.sol
```diff
@@ -450,6 +405,191 @@ contract BancorNetwork is IBancorNetwork, TokenHolder, ContractRegistryClient, F
...
+      * @param _smartToken  smart token of the first conversion step
+      * @param _amount      amount to convert from, in the source token
+    */
+    function handleSourceToken(IERC20Token _sourceToken, ISmartToken _smartToken, uint256 _amount) private {
+        bool isSourceETH = etherTokens[_sourceToken];
+
+        // validate msg.value
+        if (msg.value > 0)
+            require(msg.value == _amount && isSourceETH);
+        else
```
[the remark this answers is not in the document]
yudilevi: you're right - removed

solidity/contracts/BancorNetwork.sol
```diff
@@ -450,6 +405,191 @@ contract BancorNetwork is IBancorNetwork, TokenHolder, ContractRegistryClient, F
...
+            require(msg.value == 0);
```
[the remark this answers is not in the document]
yudilevi: removed

solidity/contracts/BancorNetwork.sol
```diff
@@ -450,6 +405,191 @@ contract BancorNetwork is IBancorNetwork, TokenHolder, ContractRegistryClient, F
...
+
+        IBancorConverter firstConverter = IBancorConverter(_smartToken.owner());
+        bool isNewerConverter = isV27OrHigherConverter(firstConverter);
+
+        // ETH / EtherToken
```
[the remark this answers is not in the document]
yudilevi: not following you - are you referring to simply move it a few lines down?

solidity/contracts/BancorNetwork.sol
```diff
@@ -450,6 +405,191 @@ contract BancorNetwork is IBancorNetwork, TokenHolder, ContractRegistryClient, F
...
+        if (isSourceETH) {
+            // handle ETH
```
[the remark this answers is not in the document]
yudilevi: but this is already checked at the beginning of the function - you can't get here if isSourceETH is false

solidity/contracts/BancorNetwork.sol
```diff
@@ -450,6 +405,191 @@ contract BancorNetwork is IBancorNetwork, TokenHolder, ContractRegistryClient, F
...
+                safeTransferFrom(_sourceToken, msg.sender, this, _amount);
+
+                // ETH converter - withdraw the ETH
+                if (isNewerConverter)
+                    IEtherToken(_sourceToken).withdraw(_amount);
+            }
+        }
+        // other ERC20 token
+        else {
+            // newer converter - transfer the tokens from the sender directly to the converter
```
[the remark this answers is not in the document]
yudilevi: but this is already checked at the beginning of the function - msg.value can't be > 0 if isSourceETH is false

solidity/contracts/BancorNetwork.sol
yudilevi: This is the target token handling - not the source, so it's unrelated to the start of the path.
We only get to this case if the last conversion in the path sent the tokens to the network.
If the last conversion ends with ETH -
- New converter will always send that ETH to the target user (so these ETH tokens can never be currently held by the network in this case)
- Old converter will always send that ETH to the network

solidity/contracts/BancorNetwork.sol
yudilevi: in `createConversionData`, we replace whatever the user passed in the path with the correct "ether token" representation, based on the conversion version:

```
if (stepData.isV27OrHigherConverter)
    stepData.targetToken = IERC20Token(0);
    // older converter - replace the source token with the EtherToken address used by the converter
    else
        stepData.targetToken = getConverterEtherTokenAddress(stepData.converter);
```

So for an old (EtherToken) converter, `targetToken` will actually be the EtherToken used by the specific converter.

solidity/contracts/BancorNetwork.sol
yudilevi: I actually don't see an issue but let's go over it together, will be more efficient

solidity/contracts/BancorNetwork.sol
yudilevi: It's used in all the cases since this relies on it -
`bool isNewerConverter = isV27OrHigherConverter(firstConverter);`

solidity/contracts/converter/BancorConverter.sol
yudilevi: Since this function isn't gas sensitive, I prefer code reuse on adding more custom code here

solidity/contracts/converter/BancorConverter.sol
yudilevi: What is the benefit that's worth to skip code reuse here?

solidity/contracts/BancorNetwork.sol
yudilevi: I want to redesign the events to make it REALLY easier to get on-chain data since I do agree we did half a solution here, so let's keep it until we have a better alternative - this is right now helping some of the data aggregators.
Maybe we can even do it in this version, though it's getting tight :)

solidity/contracts/BancorNetwork.sol
yudilevi: nice :)  - done

solidity/contracts/BancorNetwork.sol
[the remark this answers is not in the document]
barakman: 1. Add `require(msg.value == _amount);` in line 429
2. Add `require(msg.value == 0);` in line 450
3. Get rid of the `if` code in lines 419-420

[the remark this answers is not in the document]
yudilevi: gotcha - done

solidity/contracts/BancorNetwork.sol
```diff
@@ -10,9 +10,14 @@ import './utility/interfaces/IContractFeatures.sol';
 import './utility/interfaces/IWhitelist.sol';
 import './token/interfaces/IEtherToken.sol';
 import './token/interfaces/ISmartToken.sol';
-import './token/interfaces/INonStandardERC20.sol';
 import './bancorx/interfaces/IBancorX.sol';
 
+// interface of older converters for backward compatibility
+contract ILegacyBancorConverter is IBancorConverter {
```
barakman, line 16: `is IBancorConverter` is practically redundant, since no other `IBancorConverter` is used.

barakman: @yudilevi : Fixing this myself.

solidity/contracts/BancorNetwork.sol
yudilevi: @barakman cleaned up

PR #356, opened 2020-04-20
Liquidity

barakman:


solidity/contracts/converter/BancorConverter.sol
```diff
@@ -237,7 +231,6 @@ contract BancorConverter is IBancorConverter, TokenHandler, SmartTokenController
     function acceptTokenOwnership()
         public
         ownerOnly
-        totalSupplyGreaterThanZeroOnly
```
yudilevi, line 240: kef

solidity/contracts/converter/BancorConverter.sol
```diff
@@ -789,6 +782,174 @@ contract BancorConverter is IBancorConverter, TokenHandler, SmartTokenController
...
+        payable
+        multipleReservesOnly
+    {
+        verifyLiquidityInput(_reserveTokens, _reserveAmounts, _supplyMinReturnAmount);
+
+        for (uint256 i = 0; i < _reserveTokens.length; i++)
+            if (_reserveTokens[i] == address(0))
+                require(_reserveAmounts[i] == msg.value);
+
+        if (msg.value > 0)
```
yudilevi, line 796: probably better to check this before the previous for loop to save some gas on reverts

solidity/contracts/converter/BancorConverter.sol
```diff
@@ -789,6 +782,174 @@ contract BancorConverter is IBancorConverter, TokenHandler, SmartTokenController
             safeTransferFrom(_token, _from, _to, _amount);
     }
 
+    function addLiquidity(IERC20Token[] memory _reserveTokens, uint256[] memory _reserveAmounts, uint256 _supplyMinReturnAmount)
+        public
+        payable
+        multipleReservesOnly
```
yudilevi, line 788: need to merge 0.6.0 into this branch and add the `protected` modifier

solidity/contracts/converter/BancorConverter.sol
```diff
@@ -789,6 +782,174 @@ contract BancorConverter is IBancorConverter, TokenHandler, SmartTokenController
...
+            supplyAmount = addLiquidityToEmptyPool(_reserveTokens, _reserveAmounts);
+        else
+            supplyAmount = addLiquidityToNonEmptyPool(_reserveTokens, _reserveAmounts, totalSupply);
+
+        require(supplyAmount >= _supplyMinReturnAmount);
+        token.issue(msg.sender, supplyAmount);
+    }
+
+    function removeLiquidity(IERC20Token[] memory _reserveTokens, uint256[] memory _reserveMinReturnAmounts, uint256 _supplyAmount)
+        public
```
yudilevi, line 812: need to merge 0.6.0 into this branch and add the `protected` modifier

solidity/contracts/converter/BancorConverter.sol
```diff
@@ -789,6 +782,174 @@ contract BancorConverter is IBancorConverter, TokenHandler, SmartTokenController
...
+        require(length == _reserveTokens.length);
+        require(length == _reserveAmounts.length);
+
+        for (i = 0; i < length; i++) {
+            require(reserves[_reserveTokens[i]].isSet);
+            for (j = 0; j < length; j++) {
+                if (reserveTokens[i] == _reserveTokens[j])
+                    break;
+            }
+            require(j < length);
```
yudilevi, line 841: is there any way to actually fail here given that we test the length before?

solidity/contracts/converter/BancorConverter.sol
```diff
@@ -789,6 +782,174 @@ contract BancorConverter is IBancorConverter, TokenHandler, SmartTokenController
...
+                minShare = share;
+        }
+        return minShare;
+    }
+
+    function getShare(uint256 _supply, uint256 _balance, uint256 _amount) private view returns (uint256) {
+        return _supply.mul(_amount).mul(totalReserveRatio).div(_balance.add(_amount).mul(RATIO_RESOLUTION));
+    }
+
+    function ceilLog(uint256 _x) public pure returns (uint256) {
```
yudilevi, line 932: much simpler than I thought - great

solidity/contracts/converter/BancorConverterRegistry.sol
```diff
@@ -1,9 +1,11 @@
 pragma solidity 0.4.26;
+import '../utility/TokenHandler.sol';
```
yudilevi, line 2: still need to rename TokenHandler to something more self describing like SafeTransfers

solidity/contracts/converter/BancorConverterRegistry.sol
```diff
@@ -76,14 +78,113 @@ contract BancorConverterRegistry is IBancorConverterRegistry, ContractRegistryCl
...
+        string _smartTokenSymbol,
+        uint8 _smartTokenDecimals,
+        uint32 _maxConversionFee,
+        IERC20Token _reserveToken,
+        uint32 _reserveRatio,
+        uint256 _reserveAmount
+    )
+    public payable
+    {
+        IBancorConverterFactory factory = IBancorConverterFactory(addressOf(BANCOR_CONVERTER_FACTORY));
```
yudilevi, line 102: need to move all that logic to a different contract so that we can upgrade it without upgrading the registry

solidity/contracts/converter/BancorConverterRegistry.sol
```diff
@@ -76,14 +78,113 @@ contract BancorConverterRegistry is IBancorConverterRegistry, ContractRegistryCl
...
+        SmartToken token = new SmartToken(_smartTokenName, _smartTokenSymbol, _smartTokenDecimals);
+        IBancorConverter converter = IBancorConverter(factory.createConverter(token, registry, _maxConversionFee, IERC20Token(0), 0));
+
+        converter.acceptOwnership();
+
+        if (_reserveToken != address(0)) {
+            require(msg.value == 0);
+            converter.addReserve(_reserveToken, _reserveRatio);
```
yudilevi, line 110: I believe we can actually move it to the constructor now that ether has a non zero address.
But on the other hand I'm thinking of getting rid of the 2 extra ctor arguments (reserve/weight) once this function is the main function for creating a new converter

solidity/contracts/converter/BancorConverterRegistry.sol
```diff
@@ -76,14 +78,113 @@ contract BancorConverterRegistry is IBancorConverterRegistry, ContractRegistryCl
...
+        uint32 _maxConversionFee,
+        IERC20Token _reserveToken,
+        uint32 _reserveRatio,
+        uint256 _reserveAmount
+    )
+    public payable
+    {
+        IBancorConverterFactory factory = IBancorConverterFactory(addressOf(BANCOR_CONVERTER_FACTORY));
+        SmartToken token = new SmartToken(_smartTokenName, _smartTokenSymbol, _smartTokenDecimals);
+        IBancorConverter converter = IBancorConverter(factory.createConverter(token, registry, _maxConversionFee, IERC20Token(0), 0));
```
yudilevi, line 104: any way to efficiently check the reserve configuration before creating a new converter? will reduce gas significantly on fail

solidity/contracts/converter/BancorConverterRegistry.sol
```diff
@@ -76,14 +78,113 @@ contract BancorConverterRegistry is IBancorConverterRegistry, ContractRegistryCl
...
+      * 
+      * @param _smartTokenName token name
+      * @param _smartTokenSymbol token symbol
+      * @param _smartTokenDecimals token decimals
+      * @param _maxConversionFee maximum conversion-fee
+      * @param _reserveTokens reserve tokens
+      * @param _reserveRatios reserve ratios
+      * @param _reserveAmounts reserve amounts
+    */
+    function newLiquidityPool(
```
yudilevi, line 138: love these new functions

solidity/contracts/converter/BancorConverterRegistry.sol
```diff
@@ -76,14 +78,113 @@ contract BancorConverterRegistry is IBancorConverterRegistry, ContractRegistryCl
...
+      * @dev adds an existing converter to the registry
+      * only the owner can add an existing converter to the registry, as long as the converter is active and valid
       * note that a liquidity pool converter can be added only if no converter with the same reserve-configuration is already registered
       * 
       * @param _converter converter
     */
-    function addConverter(IBancorConverter _converter) external {
-        // validate input
+    function addConverter(IBancorConverter _converter) public {
+        require(msg.sender == owner || msg.sender == address(this));
```
yudilevi, line 187: msg.sender == this actually isn't gonna work well since the sender is never address(this).
in addition, we shouldn't block others from calling it until after we gave them a grace period to adapt or all portals that do that will break immediately upon the new deployment

solidity/contracts/converter/BancorConverterRegistry.sol
```diff
@@ -315,8 +415,15 @@ contract BancorConverterRegistry is IBancorConverterRegistry, ContractRegistryCl
         // verify that the converter holds balance in each of its reserves
         uint reserveTokenCount = _converter.connectorTokenCount();
         for (uint i = 0; i < reserveTokenCount; i++) {
-            if (_converter.connectorTokens(i).balanceOf(_converter) == 0)
-                return false;
+            IERC20Token reserveToken = _converter.connectorTokens(i);
+            if (reserveToken != IERC20Token(0)) {
+                if (reserveToken.balanceOf(_converter) == 0)
```
yudilevi, line 420: the new `reserveBalance(address)` can be used here instead (I might have done this already in 0.6.0)

solidity/contracts/converter/BancorConverterUpgrader.sol
```diff
@@ -8,25 +8,6 @@ import '../utility/interfaces/IWhitelist.sol';
 import '../token/interfaces/IEtherToken.sol';
 import '../FeatureIds.sol';
 
-/*
```
yudilevi, line 11: good riddance

solidity/contracts/converter/BancorConverterRegistry.sol
barakman: @yudilevi : It implement functions `safeTransfer`, `safeTransferFrom` and `safeApprove`, so it doesn't make sense to name it `SafeTransfers`.

solidity/contracts/converter/BancorConverterRegistry.sol
barakman: @yudilevi : We most certainly need to adjust this code to the new eth-reserve address.

solidity/contracts/converter/BancorConverterRegistry.sol
barakman: @yudilevi : It is when this function is called from function `addLiquidity`.

solidity/contracts/converter/BancorConverterRegistry.sol
barakman: @yudilevi : Of course. I left it as is because I was not sure about your final version of this function. This will also resolve the eth-reserve address problem here.

solidity/contracts/converter/BancorConverter.sol
barakman: @yudilevi : We need to verify:
1. That the user has provided only the reserve tokens of the converter
2. That the user has provided all the reserve tokens of the converter
The `require` before the `for j` loop verifies #1.
The `require` after the `for j` loop verifies #2.

solidity/contracts/converter/BancorConverterRegistry.sol
yudilevi: it does cause all of these are related to safe transfers

solidity/contracts/converter/BancorConverterRegistry.sol
yudilevi: the sender will be the caller to addLiquidity - not the registry

solidity/contracts/converter/BancorConverter.sol
barakman: Up until here, it's all a bunch of `require` statements.
Does it really make any difference which one we do first?
I mean, moving this to be the first means that it will save some gas on some cases of erroneous input, and consume more gas on other cases of erroneous input.
As long as we do them all at the beginning, I don't think it matters which order we do them at (also, this is really meant for very basic errors, I don't think that we should work too hard other than making sure nothing gets executed).

yudilevi: the contract factory creation is 4M gas - that's what I think we should do after all verifications

solidity/contracts/converter/BancorConverter.sol
yudilevi: ah yeah I see it now, sababa

solidity/contracts/converter/BancorConverterRegistry.sol
yudilevi: cool

solidity/contracts/converter/BancorConverterRegistry.sol
barakman: @yudilevi : Because we're not "jumping" to another contract?
So how can we prevent non-owners from calling `addConverter`?

solidity/contracts/converter/BancorConverter.sol
barakman: @yudilevi : Sorry, but what does that have to do with `addLiquidity`?

solidity/contracts/converter/BancorConverterRegistry.sol
barakman: @yudilevi : I created this branch off of 0.5.19 (or maybe `remove_ether_token`, but most certainly not from 0.6.0). So we can do this only after merging this branch to 0.6.0.
I suggest opening a new task for this merge, and then add all of the required adjustments (mentioned in some of your comments above) in a checklist.

solidity/contracts/converter/BancorConverter.sol
yudilevi: ah sorry :) there's some storage var reads there, that's 800 gas each, that's the reason I mentioned it here

solidity/contracts/converter/BancorConverterRegistry.sol
yudilevi: in this version anyone should be able to call it anyway.
in the next version we should have a shared private function so that the factory functions can call it internally and an owner only public function can call it as well

solidity/contracts/converter/BancorConverterRegistry.sol
yudilevi: pulling the target branch into this branch doesn't require a new task, it's part of a normal task workflow - it can't be merged anyway before doing that.

solidity/contracts/converter/BancorConverter.sol
barakman: @yudilevi : please see updated code.

solidity/contracts/converter/BancorConverterRegistry.sol
barakman: @yudilevi : I'm not sure what the difference is between this comment and the one which says "the contract factory creation is 4M gas - that's what I think we should do after all verifications", but please see updated code.

solidity/contracts/converter/BancorConverterRegistry.sol
barakman: @yudilevi : please see updated code.

solidity/contracts/converter/BancorConverterRegistry.sol
barakman: @yudilevi : after merging this branch to 0.6.0, we need to remember to replace the balance checking code with calling the `reserveBalances` function. We also need to use the `protected` modifier mentioned in some of your comments above.
As far as I understand:
1. we cannot do this before merging
2. we can forget doing this after merging
That's why I suggested to list these details explicitly in a Trello task.

solidity/contracts/converter/BancorConverterRegistry.sol
barakman: @yudilevi : I meant, it doesn't make sense IMO to do something like `SafeTransfers(token).safeTransfer`.

solidity/contracts/converter/BancorConverterRegistry.sol
barakman: @yudilevi : The converter registry? I thought that was the whole idea about splitting it into a data contract and a logic contract (this contract) - in order to avoid having to update the data contract and migrate all of the data in it. Now we want to also avoid updating the logic contract?

solidity/contracts/converter/BancorConverterRegistry.sol
yudilevi: sure - we can definitely add a task for these additions

solidity/contracts/converter/BancorConverterRegistry.sol
barakman: @yudilevi : please see updated code.

solidity/contracts/converter/BancorConverterRegistry.sol
barakman: @yudilevi : please see updated code.

solidity/contracts/converter/BancorConverterRegistry.sol
barakman: @yudilevi : please see updated code; this was fixed in the (now merged) `liquidity_balances` branch.

solidity/contracts/converter/BancorConverter.sol
```diff
@@ -912,21 +933,27 @@ contract BancorConverter is IBancorConverter, TokenHandler, SmartTokenController
         private
         returns (uint256)
     {
-        uint256[] memory reserveBalances = getBalances(_reserveTokens);
+        syncReserveBalances();
+        reserves[ETH_RESERVE_ADDRESS].balance = reserves[ETH_RESERVE_ADDRESS].balance.sub(msg.value);
+
         IBancorFormula formula = IBancorFormula(addressOf(BANCOR_FORMULA));
-        uint256 supplyAmount = getMinShare(_totalSupply, reserveBalances, _reserveAmounts);
+        uint256 supplyAmount = getMinShare(_totalSupply, _reserveTokens, _reserveAmounts);
```
yudilevi, line 940: wouldn't it be easier to make sure that _reserveAmounts are already in the correct order by now and use the storage reserveTokens array instead of passing the external one around?

yudilevi, line 940: does getMinShare return the TOTAL supply should be issued?

solidity/contracts/converter/BancorConverter.sol
```diff
@@ -977,7 +998,13 @@ contract BancorConverter is IBancorConverter, TokenHandler, SmartTokenController
...
+    /**
+      * @dev calculates the number of decimal digits in a given value
+      * 
+      * @param _x   value (assumed positive)
+      * @return the number of decimal digits in the given value
+    */
+    function decimalLength(uint256 _x) public pure returns (uint256) {
         uint256 y = 0;
         while (_x > 0) {
             _x /= 10;
```
yudilevi, line 1010: getting a warning here - assigning to function argument.
also, can just do y++ instead of y += 1

solidity/contracts/converter/BancorConverter.sol
```diff
@@ -986,15 +1013,28 @@ contract BancorConverter is IBancorConverter, TokenHandler, SmartTokenController
         return y;
     }
 
+    /**
+      * @dev calculates the nearest integer to a given quotient
```
yudilevi, line 1017: I think this function should be in the bancor formula contract, seems like generic math?

solidity/contracts/converter/BancorConverter.sol
```diff
@@ -986,15 +1013,28 @@ contract BancorConverter is IBancorConverter, TokenHandler, SmartTokenController
...
         return (_n + _d / 2) / _d;
     }
 
+    /**
+      * @dev calculates the average number of decimal digits in a given list of values
+      * 
+      * @param _values  list of values (each of which assumed positive)
+      * @return the average number of decimal digits in the given list of values
+    */
     function geometricMean(uint256[] memory _values) public pure returns (uint256) {
```
yudilevi, line 1033: same for this one - generic math function

solidity/test/BancorNetworkPathFinder.js
```diff
@@ -4,18 +4,81 @@
...
+            'symbol': 'DDDAAABNTBNT', 'decimals': 18, 'fee': 1000, 'reserves': [
+                {'symbol': 'DDD', 'weight': 500000, 'balance': '1e21'},
+                {'symbol': 'AAABNTBNT', 'weight': 500000, 'balance': '1e15'},
+            ]
+        },
+    ]
+};
+
+async function getSymbol(tokenAddress) {
+    if (tokenAddress == ETH_RESERVE_ADDRESS)
```
yudilevi, line 73: you sure this comparison is fine in regards to case sensitivity?

solidity/contracts/converter/BancorConverterRegistry.sol
yudilevi: we'll figure out the names but that generic name has to be replaced by something more self explanatory

solidity/contracts/converter/BancorConverterRegistry.sol
yudilevi: discussed already, we'll open a separate task for this

solidity/contracts/converter/BancorConverter.sol
barakman: @yudilevi : No, it wouldn't. Plus, using the storage array costs more than using a local variable (or in this case, an input argument, which is the same in that sense). We've actually seen this type of optimization at UniSwap and/or Balancer. In this case it is not designated for gas optimization, but reordering the amounts to be in the same order of the reserves? It sounds like `O(N^2)` to me, and even if there's a more efficient way, it's not worth the effort when I already have the **input** reserves in the correct order (because they reflect the user's intention).

barakman: @yudilevi : It happens to do that as well. But the purpose of this function is to find the smallest share invested by the caller.

solidity/contracts/converter/BancorConverter.sol
barakman: @yudilevi:
1. With `npm run build` or `truffle compile`? Weird, I don't get that. In any case, we can either suppress that warning in your compiler settings, or declare something like `uint256 z = _x` at the beginning, and then use `z /= 10`.
2. We can; I typically prefer to avoid using `++` outside of `for` loops (and in this case I chose the `+= 1` also because of the `/= 10`).

solidity/contracts/converter/BancorConverter.sol
barakman: @yudilevi : The thing is, it's not related to BancorFormula in any way.
We do have some generic functions in BancorFormula (like `power` and `floorLog` and so on).
But those are used internally by the formula functions, and are therefore not exposed to external users.
We have a total of 3 such general math functions in BancorConverter:
- `decimalLength`
- `roundDiv`
- `geometricMean`
We could place them in a generic Math library/contract, or some Util library/contract.
But none of those should be implemented and exposed in BancorFormula IMO.

solidity/contracts/converter/BancorConverter.sol
barakman: @yudilevi : Same as above.

solidity/test/BancorNetworkPathFinder.js
barakman: @yudilevi:
- In truffle tests (web3 v0.x), all addresses are lower-cased (by either truffle or web3)
- In standalone scripts (web3 v1.x), we are using `Web3.utils.toChecksumAddress`

solidity/contracts/converter/BancorConverterRegistry.sol
barakman: @yudilevi: I already have.

yudilevi: kk

solidity/contracts/converter/BancorConverter.sol
yudilevi: k

yudilevi: Sure, just wanted to make sure that it doesn't return the supply per reserve and then we need to multiply it by the number of reserves in order to get the final supply.

solidity/contracts/converter/BancorConverter.sol
yudilevi: 1. I think newer compilers do that and that's why I'm getting in
2. ...

solidity/contracts/converter/BancorConverter.sol
barakman: @yudilevi: No, it calculates the exact amount which will be issued to a user who adds liquidity to a non-empty pool.

solidity/contracts/converter/BancorConverter.sol
yudilevi: I'm planning to rename BancorFormula to Math in the future

solidity/test/BancorNetworkPathFinder.js
yudilevi: I think that it's probably best practice to not rely on truffle for that

solidity/contracts/converter/BancorConverter.sol
yudilevi: cool

solidity/contracts/converter/BancorConverter.sol
barakman: @yudilevi:
I suggest in that case to put these functions in a library, and possibly also do that with the general math functions in BancorFormula.
The advantage of using a library is that it will not yield external function calls AFAIK.
I'm not quite sure how to link a library upon deployment and all, but I'm sure we can find out.
In either case, not sure it critical to handle this ATM.
On the other hand, using a contract might help in reducing the bytecode size of the BancorConverter contract (though I'm not sure how much of that we can get from moving those 3 functions to a separate contract).

solidity/contracts/converter/BancorConverter.sol
barakman: @yudilevi: How is it possible that you're on a newer compiler. We should both be on solc 0.4.26, and if we make sure to use our Truffle version (4.1.16), then that **will** be the case.

yudilevi: I checked, it's a linter error: "security/no-assign-params: "decimalLength": Avoid assigning to function parameters."

solidity/contracts/converter/BancorConverter.sol
yudilevi: I don't want to deal with libraries specifically because of the deployment overhead.
And yes - the reason I'm bringing it up now is because I want to keep all the error messages and we'll need to get rid of some of them if we can't get the binary size smaller.

solidity/test/BancorNetworkPathFinder.js
yudilevi: Better than nothing...

solidity/contracts/converter/BancorConverterRegistry.sol
barakman: @yudilevi: I've opened a Trello task on it, so you can resolve this comment.

solidity/contracts/converter/BancorConverter.sol
barakman: @yudilevi: So maybe change that linter-rule, because there is nothing wrong with assigning a function parameter.

yudilevi: we'll leave it for now but eventually we'll want to comply with industry standards and conventions.

barakman: @yudilevi : As far as I remember, this actually used to be the standard in an older version of solc (0.4.18 or similar), and was later removed. So if anything, the industry standard is to allow changing input arguments.

yudilevi: if that was the standard, solium linter wouldn't have pointed it out as a security issue but like I said, we can leave it there for now so feel free to resolve this one.

barakman: @yudilevi: Fixed.

PR #358, opened 2020-04-29
renamed from/to to source/target

yudilevi:


solidity/contracts/BancorNetwork.sol
```diff
@@ -56,12 +56,12 @@ contract BancorNetwork is IBancorNetwork, TokenHolder, ContractRegistryClient, F
...
       * @dev triggered when a conversion between two tokens occurs
       * 
-      * @param _smartToken      smart token governed by the converter
-      * @param _fromToken       ERC20 token converted from
-      * @param _toToken         ERC20 token converted to
-      * @param _fromAmount      amount converted, in fromToken
-      * @param _toAmount        amount returned, minus conversion fee
-      * @param _trader          wallet that initiated the trade
+      * @param _smartToken  smart token governed by the converter
+      * @param _fromToken   source ERC20 token
```
barakman, line 60: @yudilevi : Did you miss these two, or would it break current API?

solidity/contracts/converter/BancorConverter.sol
```diff
@@ -56,10 +56,10 @@ contract BancorConverter is IBancorConverter, TokenHandler, SmartTokenController
     /**
       * @dev triggered when a conversion between two tokens occurs
       * 
-      * @param _fromToken       ERC20 token converted from
-      * @param _toToken         ERC20 token converted to
+      * @param _fromToken       source ERC20 token
```
barakman, line 59: @yudilevi : Did you miss these two, or would it break current API?

solidity/contracts/BancorNetwork.sol
yudilevi: yeah, will break the API

solidity/contracts/converter/BancorConverter.sol
yudilevi: same - will break the API

PR #360, opened 2020-05-01
Gas reduction step 2

yudilevi:


solidity/contracts/converter/BancorConverter.sol
```diff
@@ -528,44 +520,45 @@ contract BancorConverter is IBancorConverter, TokenHandler, SmartTokenController
...
       * @return amount of tokens received (in units of the smart token)
     */
-    function buy(IERC20Token _reserveToken, uint256 _depositAmount, uint256 _minReturn, address _beneficiary) internal returns (uint256) {
+    function buy(IERC20Token _reserveToken, uint256 _depositAmount, address _beneficiary) internal returns (uint256) {
         (uint256 amount, uint256 feeAmount) = getPurchaseReturn(_reserveToken, _depositAmount);
 
-        // ensure the trade gives something in return and meets the minimum requested amount
-        require(amount != 0 && amount >= _minReturn);
+        // ensure the trade gives something in return
+        require(amount != 0);
```
barakman, line 561: @yudilevi : Doesn't this mean that we don't actually need to verify `minReturn > 0` everywhere?

solidity/contracts/converter/BancorConverter.sol
```diff
@@ -593,19 +586,18 @@ contract BancorConverter is IBancorConverter, TokenHandler, SmartTokenController
...
+    function sell(IERC20Token _reserveToken, uint256 _sellAmount, address _beneficiary) internal returns (uint256) {
         // ensure that the input amount was already deposited
         require(_sellAmount <= token.balanceOf(this));
 
         (uint256 amount, uint256 feeAmount) = getSaleReturn(_reserveToken, _sellAmount);
 
-        // ensure the trade gives something in return and meets the minimum requested amount
-        require(amount != 0 && amount >= _minReturn);
+        // ensure the trade gives something in return
+        require(amount != 0);
```
barakman, line 600: @yudilevi : Doesn't this mean that we don't actually need to verify minReturn > 0 everywhere?

solidity/contracts/converter/BancorConverter.sol
```diff
@@ -639,16 +631,15 @@ contract BancorConverter is IBancorConverter, TokenHandler, SmartTokenController
...
       * @return amount of tokens received (in units of the target reserve token)
     */
-    function crossConvert(IERC20Token _sourceToken, IERC20Token _targetToken, uint256 _amount, uint256 _minReturn, address _beneficiary) internal returns (uint256) {
+    function crossConvert(IERC20Token _sourceToken, IERC20Token _targetToken, uint256 _amount, address _beneficiary) internal returns (uint256) {
         (uint256 amount, uint256 feeAmount) = getCrossReserveReturn(_sourceToken, _targetToken, _amount);
 
-        // ensure the trade gives something in return and meets the minimum requested amount
-        require(amount != 0 && amount >= _minReturn);
+        // ensure the trade gives something in return
+        require(amount != 0);
```
barakman, line 642: @yudilevi : Doesn't this mean that we don't actually need to verify minReturn > 0 everywhere?

solidity/test/BancorConverter.js
```diff
@@ -5,24 +5,23 @@ const fs = require('fs');
 
 const utils = require('./helpers/Utils');
 const ContractRegistryClient = require('./helpers/ContractRegistryClient');
-
```
barakman, line 8: @yudilevi : This space is there on purpose (and if you still think that we should remove it, then it appears in a bunch of other files).

solidity/test/BancorConverter.js
```diff
@@ -721,6 +712,40 @@ contract('BancorConverter', accounts => {
         await utils.catchRevert(converter.convert(tokenAddress, reserveToken.address, 30000, 1));
     });
 
+    it('verifies that convert is allowed for a whitelisted account', async () => {
+        let converter = await initConverter(accounts, true);
+        let whitelist = await Whitelist.new();
+        await whitelist.addAddress(converter.address);
```
barakman, line 718: @yudilevi : If we have a lot of such addresses on mainnet, then perhaps we should implement a function which takes an array (instead of paying a minimum of 21000 per address).

solidity/test/BancorConverterUpgrader.js
```diff
@@ -184,10 +179,10 @@ contract('BancorConverterUpgrader', accounts => {
         describe(`${init.name}(version = ${version ? version : 'latest'}, active = ${active}):`, () => {
             it('upgrade should complete successfully', async () => {
                 const [upgrader, oldConverter] = await init(deployer, version, active);
-                const oldConverterInitialState = getConverterState(oldConverter);
+                const oldConverterInitialState = await getConverterState(oldConverter);
```
barakman, line 182: @yudilevi : Hell with this @#$%. How long has that been missing there?

solidity/contracts/BancorNetwork.sol
```diff
@@ -216,6 +215,7 @@ contract BancorNetwork is IBancorNetwork, TokenHolder, ContractRegistryClient, F
       * 
       * @param _data                conversion data, see ConversionStep struct above
       * @param _amount              amount to convert from, in the source token
+      * @param _minReturn           if the conversion results in an amount smaller than the minimum return - it is cancelled, must be nonzero
```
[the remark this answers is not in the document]
yudilevi: updated the comment.
conversion that returns 0 as a result should fail (you gave funds and got nothing in return).

solidity/contracts/BancorNetwork.sol
```diff
@@ -577,8 +577,6 @@ contract BancorNetwork is IBancorNetwork, TokenHolder, ContractRegistryClient, F
             }
         }
 
-        // the last conversion step is the only one that should check the minimum return
-        data[data.length - 1].minReturn = _minReturn;
```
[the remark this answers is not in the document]
yudilevi: just wanted to do it step by step, tons of changes as is.

solidity/contracts/converter/BancorConverter.sol
```diff
@@ -22,18 +20,18 @@ import '../bancorx/interfaces/IBancorX.sol';
...
-        uint256 balance;        // reserve balance
-        uint32 weight;          // reserve weight, represented in ppm, 1-1000000
-        bool deprecated1;       // deprecated
-        bool deprecated2;       // deprecated
-        bool isSet;             // used to tell if the mapping element is defined
+        uint256 balance;    // reserve balance
+        uint32 weight;      // reserve weight, represented in ppm, 1-1000000
+        bool deprecated1;   // deprecated
+        bool deprecated2;   // deprecated
+        bool isSet;         // used to tell if the mapping element is defined
```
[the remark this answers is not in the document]
yudilevi: updated to '`true if the reserve is valid, false if not' - good?

solidity/contracts/converter/BancorConverter.sol
```diff
@@ -528,44 +520,45 @@ contract BancorConverter is IBancorConverter, TokenHandler, SmartTokenController
...
         protected
         only(BANCOR_NETWORK)
-        greaterThanZero(_minReturn)
         returns (uint256)
     {
         require(_sourceToken != _targetToken); // validate input
 
+        // if a whitelist is set, verify that both and trader and the beneficiary are whitelisted
+        require(conversionWhitelist == address(0) ||
+                (conversionWhitelist.isWhitelisted(_trader) && conversionWhitelist.isWhitelisted(_beneficiary)));
```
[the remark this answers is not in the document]
yudilevi: we can upgrade it to receive an array but since it's only relevant if whitelist is actually defined (0 such converters in the network right now), it's not worth it at this point.

solidity/contracts/converter/BancorConverter.sol
yudilevi: we do that to save some gas before doing the conversion in case min return is 0

solidity/test/BancorConverter.js
yudilevi: I don't think there should be any space between imports

solidity/test/BancorConverter.js
yudilevi: it has an `addAddresses` function that receives an array

solidity/test/BancorConverterUpgrader.js
yudilevi: since it was created, apparently

solidity/contracts/converter/BancorConverter.sol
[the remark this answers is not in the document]
barakman: @yudilevi : `false otherwise` is the common convention afaik.

solidity/contracts/converter/BancorConverter.sol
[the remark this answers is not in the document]
barakman: @yudilevi:
1. Also removes the `&&` (which might make it worth if this is called several times during a conversion-request by the user).
2. If we are planning to redeploy the white-list contract anyway, then it is definitely worth adding it IMO.

solidity/test/BancorConverter.js
barakman: @yudilevi:
This space separates between normal `require` statements and `artifacts.require` statements.
If you still think it's unnecessary, then we should probably remove it also in all the other files.

solidity/contracts/converter/BancorConverter.sol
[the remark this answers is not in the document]
yudilevi: pretty insignificant but updated

solidity/contracts/converter/BancorConverter.sol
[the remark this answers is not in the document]
yudilevi: that contract is deployed per converter that needs it and currently there's none so there's no deployment at the moment

solidity/test/BancorConverterUpgrader.js
barakman: @yudilevi: No, only since my huge refactoring of this test on 22/03 this year.

solidity/test/BancorConverter.js
yudilevi: yeah, I don't recall any popular linter that adds spaces between imports

solidity/test/BancorConverterUpgrader.js
yudilevi: so there you go

PR #362, opened 2020-05-05
Fix balance-handling in liquidity functions

barakman:
BancorConverter functions:
- `fund`
- `liquidate`
- `addLiquidity`
- `removeLiquidity`

### DIFF
--- a/solidity/contracts/converter/BancorConverter.sol
+++ b/solidity/contracts/converter/BancorConverter.sol
@@ -805,6 +805,9 @@ contract BancorConverter is IBancorConverter, TokenHandler, SmartTokenController
         protected
         multipleReservesOnly
     {
+        syncReserveBalances();
+        reserves[ETH_RESERVE_ADDRESS].balance = reserves[ETH_RESERVE_ADDRESS].balance.sub(msg.value);
+
         uint256 supply = token.totalSupply();
         IBancorFormula formula = IBancorFormula(addressOf(BANCOR_FORMULA));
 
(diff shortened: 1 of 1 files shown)


solidity/contracts/converter/BancorConverter.sol
```diff
@@ -973,6 +979,9 @@ contract BancorConverter is IBancorConverter, TokenHandler, SmartTokenController
         for (uint256 i = 0; i < _reserveTokens.length; i++) {
             if (_reserveTokens[i] != ETH_RESERVE_ADDRESS)
                 safeTransferFrom(_reserveTokens[i], msg.sender, this, _reserveAmounts[i]);
+
+            reserves[_reserveTokens[i]].balance = _reserveAmounts[i];
```
yudilevi, line 983: what if I created the converter and transferred some funds in it before calling the add liquidity function?

barakman: @yudilevi : We've discussed that already, so here is a summary of the conclusions:

The potential problem in funding an empty pool (token supply = 0), is that if the balance of one or more reserves > 0, then BancorFormula will revert on division by 0.

However, whenever we liquidate the entire supply, we return the entire amount of each one of remaining balances, which means that "supply = 0" --> "each balance = 0".

So the scenario which can lead to this problem is not supposed to ever happen. It can happen only if somebody transfers tokens directly to the converter, when the supply = 0.

The solution: just assume that all reserve balances are 0 (as they should be).

yudilevi: that's fine, but if they aren't 0 - we're basically setting the reserve balance to the wrong value so after the operation is complete, the balance is not synced.

barakman: @yudilevi : we can replace `reserves[_reserveTokens[i]].balance = _reserveAmounts[i];` during the loop with `syncReserveBalances()` after the loop.

PR #365, opened 2020-05-09
updated restrictions when adding a converter to the converter registry

yudilevi:


solidity/contracts/converter/BancorConverterRegistry.sol
```diff
@@ -186,13 +186,12 @@ contract BancorConverterRegistry is IBancorConverterRegistry, ContractRegistryCl
...
     /**
       * @dev adds an existing converter to the registry
-      * anyone can add an existing converter to the registry, as long as the converter is valid
-      * note that a liquidity pool converter can be added only if no converter with the same reserve-configuration is already registered
+      * can only be called by the owner
       * 
       * @param _converter converter
     */
-    function addConverter(IBancorConverter _converter) public {
-        require(isConverterValid(_converter) && !isSimilarLiquidityPoolRegistered(_converter));
```
barakman, line 195: @yudilevi : There's a reason why `isSimilarLiquidityPoolRegistered` has been implemented separately from `isConverterValid`, so the fact that you removed it is troublesome. Can you please explain?

solidity/contracts/converter/BancorConverterRegistry.sol
```diff
@@ -388,16 +387,9 @@ contract BancorConverterRegistry is IBancorConverterRegistry, ContractRegistryCl
     function isConverterValid(IBancorConverter _converter) public view returns (bool) {
         // verify the the smart token has a supply and that the converter is active
         ISmartToken token = ISmartTokenController(_converter).token();
-        if (token.totalSupply() == 0 || token.owner() != address(_converter))
+        if (token.owner() != address(_converter))
```
barakman, line 390: @yudilevi : Well, the entire function can be reduced to `return token.owner() == address(_converter);`.

solidity/contracts/converter/BancorConverterRegistry.sol
yudilevi: we discussed it - feel free to resolve

solidity/contracts/converter/BancorConverterRegistry.sol
yudilevi: Isn't it the case? Do you mean one liner to get the token and compare?

barakman: @yudilevi : There's an `if`/`else` here. I say - let's go for a one-liner function.

PR #366, opened 2020-05-09
registry.newLiquidiyPool now creates an empty pool

yudilevi:


solidity/contracts/converter/BancorConverterRegistry.sol
```diff
@@ -147,15 +151,12 @@ contract BancorConverterRegistry is IBancorConverterRegistry, ContractRegistryCl
...
-        uint32[] memory _reserveWeights,
-        uint256[] memory _reserveAmounts
+        uint32[] memory _reserveWeights
     )
-    public payable
+    public payable returns (IBancorConverter)
     {
         uint256 length = _reserveTokens.length;
-        require(length == _reserveWeights.length);
-        require(length == _reserveAmounts.length);
```
barakman, line 157: @yudilevi : Why have you omitted `require(length == _reserveAmounts.length);`?

yudilevi: What do you mean? I completely removed reserve amounts from the function

PR #367, opened 2020-05-10
Zero Supply Liquid Token

barakman:
1. When purchasing at supply = 0, return the reserve amount without fee.
2. When selling the entire supply, return the entire reserve without fee.

### DIFF
--- a/solidity/contracts/converter/BancorConverter.sol
+++ b/solidity/contracts/converter/BancorConverter.sol
@@ -507,13 +507,13 @@ contract BancorConverter is IBancorConverter, TokenHandler, SmartTokenController
         validReserve(_reserveToken)
         returns (uint256, uint256)
     {
-        uint256 supply = token.totalSupply();
+        uint256 totalSupply = token.totalSupply();
 
-        if (supply == 0)
-            return (_depositAmount, _depositAmount);
+        if (totalSupply == 0)
+            return (_depositAmount, 0);
 
         uint256 amount = IBancorFormula(addressOf(BANCOR_FORMULA)).purchaseRate(
-            supply,
+            totalSupply,
             reserveBalance(_reserveToken),
             reserves[_reserveToken].weight,
             _depositAmount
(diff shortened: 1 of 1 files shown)


solidity/contracts/converter/BancorConverter.sol
```diff
@@ -507,8 +507,13 @@ contract BancorConverter is IBancorConverter, TokenHandler, SmartTokenController
         validReserve(_reserveToken)
         returns (uint256, uint256)
     {
+        uint256 supply = token.totalSupply();
+
+        if (supply == 0)
+            return (_depositAmount, _depositAmount);
```
yudilevi, line 513: I'm pretty sure the 2nd value here should be 0 (fee), no?

solidity/contracts/converter/BancorConverter.sol
```diff
@@ -536,9 +541,15 @@ contract BancorConverter is IBancorConverter, TokenHandler, SmartTokenController
         validReserve(_reserveToken)
         returns (uint256, uint256)
     {
+        uint256 supply = token.totalSupply();
+        uint256 balance = reserveBalance(_reserveToken);
+
+        if (supply == _sellAmount)
+            return (balance, balance);
```
yudilevi, line 548: same here - 2nd value should be 0, no?

solidity/contracts/converter/BancorConverter.sol
barakman: @yudilevi : No, it should be amount minus fee.

solidity/contracts/converter/BancorConverter.sol
barakman: @yudilevi : No, it should be amount minus fee.

solidity/contracts/converter/BancorConverter.sol
barakman: @yudilevi : Damn, you're right!

solidity/contracts/converter/BancorConverter.sol
barakman: @yudilevi : Damn, you're right!

barakman: @yudilevi : Fixed.

solidity/contracts/converter/BancorConverter.sol
barakman: @yudilevi : Fixed.

PR #371, opened 2020-05-16
added a path finder wrapper in the network

yudilevi:


solidity/contracts/BancorNetworkPathFinder.sol
```diff
@@ -38,7 +39,7 @@ contract BancorNetworkPathFinder is ContractRegistryClient {
       * 
       * @return a path from the source token to the target token
     */
-    function generatePath(address _sourceToken, address _targetToken) public view returns (address[] memory) {
+    function findPath(address _sourceToken, address _targetToken) public view returns (address[] memory) {
```
barakman, line 42: @yudilevi : We can technically keep this interface as is, and just connect the already deployed path-finder to the (new) network. Any special reason which justifies this API-change?

solidity/contracts/utility/ContractRegistryClient.sol
```diff
@@ -11,6 +11,7 @@ contract ContractRegistryClient is Owned, Utils {
     bytes32 internal constant BANCOR_NETWORK = "BancorNetwork";
     bytes32 internal constant BANCOR_FORMULA = "BancorFormula";
     bytes32 internal constant CONVERTER_FACTORY = "ConverterFactory";
+    bytes32 internal constant CONVERSION_PATH_FINDER = "ConversionPathFinder";
```
barakman, line 14: @yudilevi : So shouldn't we replace `BancorNetworkPathFinder` with `ConversionPathFinder` conclusively?

barakman: BTW, it is definitely on the "Network-Level", so I'm not sure we should go with "ConversionPathFinder" to begin with.

solidity/test/helpers/ContractRegistryClient.js
```diff
@@ -2,6 +2,7 @@ module.exports.CONTRACT_REGISTRY              = web3.fromAscii("ContractRegistry
 module.exports.BANCOR_NETWORK                 = web3.fromAscii("BancorNetwork");
 module.exports.BANCOR_FORMULA                 = web3.fromAscii("BancorFormula");
 module.exports.CONVERTER_FACTORY              = web3.fromAscii("ConverterFactory");
+module.exports.CONVERSION_PATH_FINDER         = web3.fromAscii("ConversionPathFinder");
```
barakman, line 5: @yudilevi : Excellent choice for placing this line!

solidity/contracts/BancorNetwork.sol
```diff
@@ -106,6 +107,20 @@ contract BancorNetwork is IBancorNetwork, TokenHolder, ContractRegistryClient {
...
+    /**
+      * @dev returns the conversion path between two tokens in the network
+      * note that this method is quite expensive in terms of gas and should generally be called off-chain
+      * 
+      * @param _sourceToken source token address
+      * @param _targetToken target token address
+      *
+      * @return conversion path between the two tokens
+    */
+    function conversionPath(IERC20Token _sourceToken, IERC20Token _targetToken) public view returns (address[]) {
```
barakman, line 119: @yudilevi : This would cost more gas (when called from a transaction) and more provider computation time (when called from an RPC) than calling it directly from the path-finder.
The only reason in favor of this as far as I can think of, is that it will allow us to change the path-finder implementation without changing the network implementation (and so if you consider redeploying the network as an API-breaking move on our behalf, then this will allow us to avoid it).
Is that your primary goal here?

solidity/contracts/BancorNetworkPathFinder.sol
yudilevi: yeah, it's part of the property style API and next time we upgrade the pathfinder, I don't want to break the network, so might as well do now, anyone who uses the current pathfinder can still keep using it, so no harm done here.

solidity/contracts/utility/ContractRegistryClient.sol
yudilevi: already opened a task for it -
https://trello.com/c/RLI1EQrC/328-ethereum-rename-bancor-network-path-finder-to-conversionpathfinder

solidity/test/helpers/ContractRegistryClient.js
yudilevi: I learned from the best

solidity/contracts/BancorNetwork.sol
yudilevi: yes, that's one reason and another reason is for users to know/worry about one less contract

solidity/contracts/BancorNetwork.sol
```diff
@@ -106,6 +107,20 @@ contract BancorNetwork is IBancorNetwork, TokenHolder, ContractRegistryClient {
...
+        IBancorNetworkPathFinder pathFinder = IBancorNetworkPathFinder(addressOf(CONVERSION_PATH_FINDER));
```
barakman, line 120: @yudilevi : Can merge these two lines into one.

solidity/test/BancorNetwork.js
```diff
@@ -122,6 +125,14 @@ async function initTokensAndConverters(accounts) {
...
 
+    let converterRegistry = await BancorConverterRegistry.new(contractRegistry.address);
+    let converterRegistryData = await BancorConverterRegistryData.new(contractRegistry.address);
+    await contractRegistry.registerAddress(ContractRegistryClient.BANCOR_CONVERTER_REGISTRY, converterRegistry.address);
+    await contractRegistry.registerAddress(ContractRegistryClient.BANCOR_CONVERTER_REGISTRY_DATA, converterRegistryData.address);
+
+    let pathFinder = await BancorNetworkPathFinder.new(contractRegistry.address);
+    await contractRegistry.registerAddress(ContractRegistryClient.CONVERSION_PATH_FINDER, pathFinder.address);
+
     bnt = await ERC20Token.new('BNT', 'BNT', 2, 10000000);
```
barakman, line 136: @yudilevi : I recommend renaming this to `bntToken`.

solidity/contracts/IBancorNetworkPathFinder.sol
```diff
@@ -0,0 +1,9 @@
+pragma solidity 0.4.26;
+import './token/interfaces/IERC20Token.sol';
+
+/*
+    Bancor Network Path Finder interface
+*/
+contract IBancorNetworkPathFinder {
```
barakman, line 7: @yudilevi : So why not `IConversionPathFinder`?

solidity/contracts/BancorNetwork.sol
yudilevi: I prefer not to have very long lines in the code and the cost is insignificant, so better keep it as is for readability.

solidity/test/BancorNetwork.js
yudilevi: it's a bit redundant since bnt is Bancor Network Token but sure, will update.

solidity/contracts/IBancorNetworkPathFinder.sol
yudilevi: for the sake of consistency - will rename these together right after the merge

solidity/test/BancorNetwork.js
yudilevi: done

PR #372, opened 2020-05-16
moved completeXConversions to the network

yudilevi:


solidity/contracts/bancorx/interfaces/IBancorX.sol
```diff
@@ -1,6 +1,8 @@
 pragma solidity 0.4.26;
+import '../../token/interfaces/IERC20Token.sol';
 
 contract IBancorX {
+    function token() public view returns (IERC20Token) {this;}
```
barakman, line 5: @yudilevi : We should try to avoid "dummy" implementation in interfaces.

solidity/test/XConversions.js
```diff
@@ -73,6 +73,7 @@ contract("XConversions", accounts => {
             const path = erc20TokenBntPath
             const amount = web3.toWei(1)
 
+            await erc20Token.approve(bancorNetwork.address, 0, { from: accounts[5] })
```
barakman, line 76: @yudilevi : This line implies a stateful test. How come the allowance is larger than 0 (to require executing this "0-approve" transaction)? We should strive for stateless tests ("it"s).

solidity/test/XConversions.js
```diff
@@ -111,17 +112,10 @@ contract("XConversions", accounts => {
...
-                1,
-                xTransferId,
-                0,
-                0,
-                utils.zeroBytes32,
-                utils.zeroBytes32,
-                { from: accounts[5] }
-            )
-
+            await bntToken.approve(bancorNetwork.address, 0, { from: accounts[5] })
```
barakman, line 115: @yudilevi : This line implies a stateful test. How come the allowance is larger than 0 (to require executing this "0-approve" transaction)? We should strive for stateless tests ("it"s).

solidity/test/XConversions.js
```diff
@@ -137,27 +131,10 @@ contract("XConversions", accounts => {
...
-                path,                                              
-                1,                                               
-                xTransferId,                                       
-                0,                                                    
-                0,                                                      
-                utils.zeroBytes32,                                                      
-                utils.zeroBytes32,                                                      
-                { from: accounts[5] }
-            )
+            await bntToken.approve(bancorNetwork.address, 0, { from: accounts[5] })
```
barakman, line 134: @yudilevi : This line implies a stateful test. How come the allowance is larger than 0 (to require executing this "0-approve" transaction)? We should strive for stateless tests ("it"s).

solidity/test/XConversions.js
```diff
@@ -175,87 +152,12 @@ contract("XConversions", accounts => {
...
-
-            const res = await bntConverter.completeXConversion2(
-                path,                                              
-                1,                                               
-                xTransferId,                                       
-                { from: accounts[5] }
-            )
-
-            const currBalance = await erc20Token.balanceOf(accounts[5])
+            await bntToken.approve(bancorNetwork.address, 0, { from: accounts[5] })
```
barakman, line 155: @yudilevi : This line implies a stateful test. How come the allowance is larger than 0 (to require executing this "0-approve" transaction)? We should strive for stateless tests ("it"s).

solidity/contracts/bancorx/interfaces/IBancorX.sol
yudilevi: we'll clean it up when we move to explicit getters

solidity/test/XConversions.js
yudilevi: this is deliberately there in order to ensure we operate in a stateless state.
so even if there are other tests that might have done it, this line will reset the approval just in case it was forgotten in a different test.

solidity/contracts/bancorx/interfaces/IBancorX.sol
barakman: @yudilevi : Why not clean it up now?

yudilevi: since it's pretty involved to do that across the board and it has no effect on public interface, so it's less urgent.

PR #380, opened 2020-06-27
Project and tests refactoring

lbeder:
- Updated `truffle` to `5.1.32`
- Updated web3.js to `1.2.9`
- Updated scripts
- Switched to `ganache-core`
- Updated to the newest `solidity-coverage` and run it as a plugin
- Added `solhint` and added npm scripts
- Added `eslint` 
- Updated tests to `expect` and `chai`
- Updated tests to use `BN` instead of `BigNumber`
- Avoided tests state sharing via `before()` callbacks
- Various tests updates and improvements
- Updated README.md


scripts/deploy-one.js
```diff
@@ -1,92 +1,105 @@
...
+const NODE_ADDRESS = process.argv[2];
+const PRIVATE_KEY = process.argv[3];
 const CONTRACT_NAME = process.argv[4];
 const CONTRACT_ARGS = process.argv.slice(5);
 
-async function scan(message) {
+async function scan (message) {
     process.stdout.write(message);
-    return await new Promise(function(resolve, reject) {
+    return await new Promise(function (resolve, reject) {
```
yudilevi, line 12: need to go over the linter rules but as an example, we prefer arrow style and no space between `function` and its args

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@bancor/contracts-solidity",
-  "version": "0.6.4",
+  "version": "0.6.5",
```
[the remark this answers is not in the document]
lbeder: I'll fix it

scripts/deploy-one.js
lbeder: This is a leftover of the previous version of the script, but I'll quickly fix it everywhere. If you see additional style violation, please let me know and I'll reformat.

PR #389, opened 2020-07-17
Add more checks to PriceOracle

lbeder:


solidity/contracts/utility/PriceOracle.sol
```diff
@@ -47,15 +39,29 @@ contract PriceOracle is IPriceOracle, Utils {
...
+    modifier vaidAddresses(address _address1, address _address2) {
+        _vaidAddresses(_address1, _address2);
+        _;
+    }
+
+    // error message binary size optimization
+    function _vaidAddresses(address _address1, address _address2) internal pure {
+        _validAddress(_address1);
+        _validAddress(_address2);
+        require(_address1 != _address2, "ERR_SAME_ADDRESSES");
```
barakman, line 52: Either `IDENTICAL_ADDRESSES` or `SAME_ADDRESS` (and why do we even check that inside a function called `_vaidAddresses`?).

solidity/contracts/utility/PriceOracle.sol
```diff
@@ -47,15 +39,29 @@ contract PriceOracle is IPriceOracle, Utils {
...
+    modifier supportedTokens(IERC20Token _tokenA, IERC20Token _tokenB) {
+        _supportedTokens(_tokenA, _tokenB);
         _;
     }
 
     // error message binary size optimization
-    function _supportedToken(IERC20Token _token) internal view {
-        require(tokensToOracles[_token] != address(0), "ERR_UNSUPPORTED_TOKEN");
+    function _supportedTokens(IERC20Token _tokenA, IERC20Token _tokenB) internal view {
+        require(_tokenA != _tokenB, "ERR_SAME_ADDRESSES");
```
barakman, line 63: Why not use `_vaidAddresses(_tokenA, _tokenB)` instead of `require(_tokenA != _tokenB, ...)`?

lbeder: Mostly to avoid the redundant check for `address(0)`. Do you think it'd more readable this way, regarding the small additional gas cost?

solidity/contracts/utility/PriceOracle.sol
lbeder: Fixed

solidity/contracts/utility/PriceOracle.sol
barakman: Depending how frequently it is used. Gas cost precedes all other considerations, that I agree.
But aren't we supposed to check that anyway (that these tokens are not zero)?

solidity/contracts/utility/PriceOracle.sol
barakman: I meant to say that the name `_vaidAddresses` was not so suitable.

solidity/contracts/utility/PriceOracle.sol
lbeder: It is explicitly checked in `tokensToOracles[_tokenA] != address(0)` below

solidity/contracts/utility/PriceOracle.sol
```diff
@@ -28,17 +28,9 @@ contract PriceOracle is IPriceOracle, Utils {
     */
     constructor(IERC20Token _tokenA, IERC20Token _tokenB, IChainlinkPriceOracle _tokenAOracle, IChainlinkPriceOracle _tokenBOracle)
         public
-        validAddress(_tokenA)
-        validAddress(_tokenB)
-        validAddress(_tokenAOracle)
-        validAddress(_tokenBOracle)
+        vaidAddresses(_tokenA, _tokenB)
```
yudilevi, line 31: typo

lbeder: 🙈 

Fixing

solidity/contracts/utility/PriceOracle.sol
yudilevi: we can use the `validAddress` modifier we already have and call it twice.
we can also add a generic modifier that ensures that 2 addresses are different.
thoughts?

PR #391, opened 2020-07-18
Update project dependencies

lbeder:


package.json
```diff
@@ -31,7 +31,7 @@
   },
   "devDependencies": {
     "@openzeppelin/test-helpers": "^0.5.6",
-    "@truffle/contract": "^4.2.10",
+    "@truffle/contract": "^4.2.13",
```
barakman, line 34: As far as I remember, we use `truffle/contract` only in one helper file which is used only in one test (for old converters).
And as far as I remember, it was suitable for `truffle/contract` v3 but not for v4.
I trust you've upgraded that helper as part of upgrading to Truffle v5, right?
Otherwise that specific test might fail (though, I guess, if we're already on `truffle/contract` v4, then it should be fine, so just making sure)...

lbeder: It's a good observation, but we have indeed solved it when we've upgraded to truffle v5

PR #396, opened 2020-07-20
Deployment gas optimization

yudilevi:


solidity/contracts/converter/types/liquidity-pool-v2/LiquidityPoolV2Converter.sol
```diff
@@ -702,6 +700,28 @@ contract LiquidityPoolV2Converter is LiquidityPoolConverter {
...
+      *
+    */
+    function createPoolTokens() internal {
+        IPoolTokensContainer container = IPoolTokensContainer(anchor);
+        if (container.poolTokens().length != 0) {
+            return;
+        }
+
+        uint256 reserveCount = reserveTokens.length;
+        for (uint256 i = 0; i < reserveCount; i++) {
```
barakman, line 716: I would change here to `uint256 i = container.poolTokens().length...` and get rid of the check above (same effect but cleaner code).

PR #397, opened 2020-07-20
Deployment gas optimization

yudilevi:


solidity/contracts/converter/types/liquidity-pool-v2/LiquidityPoolV2Converter.sol
```diff
@@ -702,6 +700,28 @@ contract LiquidityPoolV2Converter is LiquidityPoolConverter {
...
+      *
+    */
+    function createPoolTokens() internal {
+        IPoolTokensContainer container = IPoolTokensContainer(anchor);
+        if (container.poolTokens().length != 0) {
+            return;
+        }
+
+        uint256 reserveCount = reserveTokens.length;
+        for (uint256 i = 0; i < reserveCount; i++) {
```
barakman, line 716: I would change here to `uint256 i = container.poolTokens().length...` and get rid of the check above (same effect but cleaner code).

solidity/test/Converter.js
```diff
@@ -91,6 +112,24 @@ contract('Converter', accounts => {
...
+            case 0: 
+                anchor = await SmartToken.new('Token1', 'TKN1', 2);
+                break;
+
+            case 1:
+                anchor = await SmartToken.new('Pool1', 'POOL1', 2);
+                break;
+
+            case 2:
+                anchor = await PoolTokensContainer.new('Pool', 'POOL', 2);
```
barakman, line 126: Forgot `'2'` at the end of the name and at the end of the symbol.

barakman: I think...

solidity/contracts/converter/types/liquidity-pool-v2/LiquidityPoolV2Converter.sol
yudilevi: I'd rather have it explicit here, just in case

solidity/test/Converter.js
yudilevi: no `2` here - the pool tokens are gonna have 1 & 2.

solidity/contracts/converter/types/liquidity-pool-v2/LiquidityPoolV2Converter.sol
barakman: What if we allow adding tokens in the future?

yudilevi: that's pretty radical but we can change then if it's relevant

barakman: OK, so wait for @lbeder to give his opinion on this suggestion, or resolve it?

lbeder: I'd agree with @yudilevi here. The condition isn't immediately clear, so being more explicit here would be safer

solidity/test/LiquidityPoolV2Converter.js
```diff
@@ -30,20 +30,23 @@ const Whitelist = artifacts.require('Whitelist');
...
 
         now = await latest();
         await converter.setTime(now);
 
         if (activate) {
-            await anchor.transferOwnership(converter.address);
-            await converter.acceptTokenOwnership();
             await converter.activate(primaryReserveAddress, chainlinkPriceOracleA.address, chainlinkPriceOracleB.address);
+
+            poolToken1 = await SmartToken.at(await converter.poolToken.call(getReserve1Address(isETHReserve)));
```
barakman, line 48: No need to `await` on `at`.

barakman: And I just noticed that we've got 26 occurrences of that in 8 different files.

PR #403, opened 2020-07-25
Pass the rate as the amount of secondary tokens equal to one primary …

barakman:
…token.

### DIFF
--- a/solidity/contracts/converter/types/liquidity-pool-v2/LiquidityPoolV2Converter.sol
+++ b/solidity/contracts/converter/types/liquidity-pool-v2/LiquidityPoolV2Converter.sol
@@ -851,8 +851,8 @@ contract LiquidityPoolV2Converter is LiquidityPoolConverter {
             primaryStakedBalance.mul(AMPLIFICATION_FACTOR),
             primaryBalance,
             secondaryBalance,
-            _rate.d,
-            _rate.n);
+            _rate.n,
+            _rate.d);
     }
 
     /**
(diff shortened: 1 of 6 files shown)


solidity/contracts/converter/types/liquidity-pool-v2/LiquidityPoolV2Converter.sol
```diff
@@ -851,8 +851,8 @@ contract LiquidityPoolV2Converter is LiquidityPoolConverter {
             primaryStakedBalance.mul(AMPLIFICATION_FACTOR),
             primaryBalance,
             secondaryBalance,
-            _rate.d,
-            _rate.n);
+            _rate.n,
```
yudilevi, line 854: can remove the inverse comment

solidity/test/LiquidityPoolV2Converter.js
```diff
@@ -149,8 +149,8 @@ contract('LiquidityPoolV2Converter', accounts => {
                     primaryReserveData[0].mul(AMPLIFICATION_FACTOR),
                     primaryReserveData[1],
                     secondaryReserveData[1],
-                    rate.d,
-                    rate.n
+                    rate.n,
```
yudilevi, line 152: can remove the inverse comment

solidity/test/helpers/FormulaFunctions.js
```diff
@@ -129,15 +129,15 @@ const balancedWeights = (primaryReserveStakedBalance, primaryReserveBalance, sec
     }
     expect(q.gt(0) && p.gt(0)).to.be.true('ERR_INVALID_RESERVE_RATE');
 
-    const tp = t.mul(p);
-    const rq = r.mul(q);
+    const tq = t.mul(q);
```
lbeder, line 132: You should change the calls to this method in the test themselves. Otherwise, they will fail

barakman: I did not find any such call when I made this change. Did you actually see any of the tests failing because of that?

PR #405, opened 2020-07-25
Rename acceptTokenOwnership to acceptAnchorOwnership

lbeder:


solidity/contracts/converter/LiquidityPoolConverter.sol
```diff
@@ -70,9 +70,9 @@ contract LiquidityPoolConverter is ConverterBase {
       * can only be called by the contract owner
       * note that prior to version 28, you should use 'acceptTokenOwnership' instead
     */
-    function acceptTokenOwnership() public {
```
barakman, line 73: What about the tests?

lbeder: It's the other way around: this actually fixes the tests 😅

PR #406, opened 2020-07-25
Add the trivial ETH/ETH rate oracle

lbeder:


solidity/contracts/utility/ChainlinkETHToETHOracle.sol
```diff
@@ -0,0 +1,28 @@
...
+    function latestAnswer() external view returns (int256) {
+        return ETH_RATE;
+    }
+
+    /**
+      * @dev returns the trivial ETH/ETH update time.
+      *
+      * @return always returns current block's timestamp
+    */
+    function latestTimestamp() external view returns (uint256) {
```
yudilevi, line 25: this means that we'll do a rate calculation every block - might be better to use older date but that might break if we add checks for stale oracle.
how about returning now - hour or something like that? and then the timing will mostly be based on the other token's oracle if I'm not mistaken

lbeder: We'd do a rate calculation only if the rate of TKN/ETH was changed. Otherwise, it'll take `min(rate of TKN/ETH, now)` which would stay the same.

yudilevi: you're right - and are there no gas implications to the fact that the timestamp has changed?

lbeder: I don't think so, because we will always do `min(rate of TKN/ETH, now)` anyway

PR #410, opened 2020-07-28
added converter type to events

yudilevi:


solidity/contracts/converter/ConverterBase.sol
```diff
@@ -72,10 +72,11 @@ contract ConverterBase is IConverter, TokenHandler, TokenHolder, ContractRegistr
     /**
       * @dev triggered when the converter is activated
       *
+      * @param _type        converter type
       * @param _anchor      converter anchor
       * @param _activated   true if the converter was activated, false if it was deactivated
     */
-    event Activation(IConverterAnchor _anchor, bool _activated);
+    event Activation(uint16 indexed _type, IConverterAnchor indexed _anchor, bool indexed _activated);
```
barakman, line 79: I've hanged my mind - didn't realize that there were these 3 parameters (you wrote only the last two, in your question).
First anchor, then type.

PR #415, opened 2020-08-06
0.6.9

barakman:


solidity/contracts/converter/types/liquidity-pool-v2/LiquidityPoolV2Converter.sol
```diff
@@ -678,20 +700,28 @@ contract LiquidityPoolV2Converter is LiquidityPoolConverter {
         uint256 _targetAmount)
         internal view returns (uint256)
     {
-        // conversions to the primary reserve apply normal fees
-        if (_targetToken == primaryReserveToken) {
-            return super.calculateFee(_targetAmount);
+        uint256 fee;
```
yudilevi, line 703: I think the function should return super.calculateConversionFee + the adjusted fee

solidity/contracts/converter/types/liquidity-pool-v2/LiquidityPoolV2Converter.sol
```diff
@@ -156,6 +167,17 @@ contract LiquidityPoolV2Converter is LiquidityPoolConverter {
         emit Activation(converterType(), anchor, true);
     }
 
+    /**
+      * @dev updates the current dynamic fee factor
+      * can only be called by the contract owner
+      *
+      * @param _dynamicFeeFactor new dynamic fee factor, represented in ppm
+    */
+    function setDynamicFeeFactor(uint256 _dynamicFeeFactor) public ownerOnly {
```
yudilevi, line 176: limit the factor to 10%, should be enough and will prevent abuse

solidity/contracts/converter/types/liquidity-pool-v2/LiquidityPoolV2Converter.sol
barakman: Fixed.

solidity/contracts/converter/types/liquidity-pool-v2/LiquidityPoolV2Converter.sol
barakman: Worse, we must verify that `factor <= 100% / AMP` (which means 5%), otherwise the dynamic fee might be larger than 100%.

PR #420, opened 2020-08-27
Remove 'stack too deep' workarounds in LPV1/LPV2 converters

barakman:


solidity/contracts/converter/types/liquidity-pool-v1/LiquidityPoolV1Converter.sol
```diff
@@ -147,11 +146,27 @@ contract LiquidityPoolV1Converter is LiquidityPoolConverter {
...
+        uint32 sourceReserveWeight = reserves[_sourceToken].weight;
+        uint32 targetReserveWeight = reserves[_targetToken].weight;
+
         // dispatch the conversion event
         dispatchConversionEvent(_sourceToken, _targetToken, _trader, _amount, amount, fee);
 
-        // dispatch rate updates
-        dispatchRateEvents(_sourceToken, _targetToken);
+        // dispatch token rate update event
+        uint256 rateN = targetReserveBalance.mul(sourceReserveWeight);
```
yudilevi, line 159: I think it's still worth having a pure utility function that does instead of doing the calculation (and potentially missing something like the order) in every place.

solidity/contracts/converter/types/liquidity-pool-v2/LiquidityPoolV2Converter.sol
```diff
@@ -248,17 +248,6 @@ contract LiquidityPoolV2Converter is LiquidityPoolConverter {
...
     }
 
-    /**
-      * @dev returns the pool token address by the reserve token address
-      *
-      * @param _reserveToken    reserve token address
-      *
-      * @return pool token address
-    */
-    function poolToken(IERC20Token _reserveToken) public view returns (ISmartToken) {
```
yudilevi, line 258: this is a public function that we still need to expose

yudilevi (changes requested):

solidity/contracts/converter/types/liquidity-pool-v2/LiquidityPoolV2Converter.sol
barakman: Fixed.

solidity/contracts/converter/types/liquidity-pool-v1/LiquidityPoolV1Converter.sol
barakman: Fixed.

PR #421, opened 2020-08-29
Switch to real interfaces.

barakman:


solidity/contracts/converter/ConverterBase.sol
```diff
@@ -68,6 +68,9 @@ abstract contract ConverterBase is IConverter, TokenHandler, TokenHolder, Contra
     uint32 public override conversionFee = 0;           // current conversion fee, represented in ppm, 0...maxConversionFee
     bool public constant conversionsEnabled = true;     // deprecated, backward compatibility
 
+    function converterType() public virtual override pure returns (uint16);
```
yudilevi, line 71: figure out what the standard convention is and if it mentions none of that, let's move them to after the ctor

barakman: Done at https://github.com/bancorprotocol/contracts-solidity/pull/422.

PR #430, opened 2020-09-12
Recent average rate

yudilevi:


solidity/contracts/converter/types/liquidity-pool-v1/LiquidityPoolV1Converter.sol
```diff
@@ -69,6 +82,25 @@ contract LiquidityPoolV1Converter is LiquidityPoolConverter {
...
+      *
+      * @return rate (numerator)
+      * @return rate (denominator)
+    */
+    function tokensRate(IERC20Token _token1, IERC20Token _token2) public view returns (uint256, uint256) {
+        uint256 token1Balance = reserves[_token1].balance;
+        uint256 token2Balance = reserves[_token2].balance;
+        uint32 token1Weight = reserves[_token1].weight;
+        uint32 token2Weight = reserves[_token2].weight;
+
```
barakman, line 100: If called a lot, then perhaps it is better to do:
```
Reserve storage reserve1 = reserves[_token1];
Reserve storage reserve2 = reserves[_token2];
uint256 token1Balance = reserve1.balance;
uint256 token2Balance = reserve2.balance;
uint32 token1Weight = reserve1.weight;
uint32 token2Weight = reserve2.weight;
```

solidity/contracts/converter/types/liquidity-pool-v1/LiquidityPoolV1Converter.sol
```diff
@@ -506,7 +606,18 @@ contract LiquidityPoolV1Converter is LiquidityPoolConverter {
         return uint256(10) ** (roundDiv(numOfDigits, length) - 1);
     }
 
-     /**
+    /**
+      * @dev checks if the pool is a 2 reserves / 50%/50% weights pool
+      *
+      * @return true if the pool is standard, false otherwise
+    */
+    function isStandardPool() internal view returns (bool) {
```
barakman, line 614: We cannot use this concept in general, because a pool can alternate between "standard" and "non-standard" (think of v2 pools, for example).
To my understanding, we should set this as an "immutable after activation" state of the pool.
In V1 - it is pretty easy:
- Start it as false
- Set it to true upon adding the 2nd reserve, if both weights are 50%
- Set it to false upon adding the 3rd reserve

barakman (changes requested):

solidity/contracts/converter/types/liquidity-pool-v1/LiquidityPoolV1Converter.sol
```diff
@@ -9,11 +9,24 @@ import "../../../token/interfaces/ISmartToken.sol";
   * The liquidity pool v1 converter is a specialized version of a converter that manages
   * a classic bancor liquidity pool.
   *
-  * Even though classic pools can have many reserves, the most common configuration of
-  * the pool has 2 reserves with 50%/50% weights.
+  * Even though pools can have many reserves, the standard pool configuration
+  * is 2 reserves with 50%/50% weights.
 */
 contract LiquidityPoolV1Converter is LiquidityPoolConverter {
+    struct Fraction {
```
barakman, line 16: We probably wanna move this `struct Fraction` to contract `BaseConverter` (or even to a library which contract `BaseConverter` will be using).

solidity/test/LiquidityPoolV1Converter.js
```diff
@@ -668,4 +675,171 @@ contract('LiquidityPoolV1Converter', accounts => {
...
+
+            return { n: newAverageRateN, d: newAverageRateD };
+        };
+
+        const expectRatesAlmostEqual = (rate, newRate) => {
+            const rate1 = Decimal(rate.n.toString()).div(Decimal(rate.d.toString()));
+            const rate2 = Decimal(newRate.n.toString()).div(Decimal(newRate.d.toString()));
+
+            const ratio = rate1.div(rate2);
+            expect(ratio.gte(0.999998)).to.be.true(`${ratio.toString()} is below MIN_RATIO`);
```
barakman, line 708: Use strings, not (floating-point) numbers.

barakman (changes requested):

solidity/contracts/converter/types/liquidity-pool-v1/LiquidityPoolV1Converter.sol
yudilevi: done

solidity/contracts/converter/types/liquidity-pool-v1/LiquidityPoolV1Converter.sol
yudilevi: done

solidity/contracts/converter/types/liquidity-pool-v1/LiquidityPoolV1Converter.sol
yudilevi: moved into a dedicated `Types` utility contract

solidity/test/LiquidityPoolV1Converter.js
yudilevi: done

barakman: I think that you don't need `toString()` inside a `${}`.

yudilevi: done

barakman (changes requested):

solidity/contracts/converter/types/liquidity-pool-v1/LiquidityPoolV1Converter.sol
```diff
@@ -69,6 +81,43 @@ contract LiquidityPoolV1Converter is LiquidityPoolConverter {
...
     }
 
+    /**
+      * @dev defines a new reserve token for the converter
+      * can only be called by the owner while the converter is inactive
+      *
+      * @param _token   address of the reserve token
+      * @param _weight  reserve weight, represented in ppm, 1-1000000
+    */
+    function addReserve(IERC20Token _token, uint32 _weight) public override ownerOnly {
```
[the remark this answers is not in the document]
yudilevi: yeah it's enforced in the base contract

solidity/contracts/converter/types/liquidity-pool-v1/LiquidityPoolV1Converter.sol
```diff
@@ -155,6 +210,68 @@ contract LiquidityPoolV1Converter is LiquidityPoolConverter {
...
+    }
+
+    /**
+      * @dev returns the recent average rate of 1 reserve token 0 in reserve token 1 units
+      *
+      * @return recent average rate between the reserves
+    */
+    function recentAverageRate() internal view returns (Fraction memory) {
+        // get the elapsed time since the previous average rate was calculated
+        uint256 timeElapsed = time() - prevAverageRateUpdateTime;
```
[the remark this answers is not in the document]
yudilevi: why?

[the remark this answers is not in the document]
barakman: Because in `time() - prevAverageRateUpdateTime`, you are assuming that `time() >= prevAverageRateUpdateTime`.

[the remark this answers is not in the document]
yudilevi: it's good enough since we can also introduce a function that updates `prevAverageRateUpdateTime` in the tests if needed.
I actually did and then removed it - it's better to increase the time than move it back and forth.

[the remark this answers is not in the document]
barakman: So whatever function we put in the tester contract should be named `incTime` instead of `setTime`.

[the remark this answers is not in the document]
yudilevi: it's all under the test responsibility (including the custom setters) so I don't see the significance in doing that, better keep it simple in the tests.

solidity/contracts/converter/types/liquidity-pool-v1/LiquidityPoolV1Converter.sol
```diff
@@ -69,6 +81,43 @@ contract LiquidityPoolV1Converter is LiquidityPoolConverter {
...
+    /**
+      * @dev returns the rate of 1 `_token1` in `_token2` units
+      * note that the rate can only be queried for the reserves
+      *
+      * @param _token1  reserve token to get the rate of one unit of
+      * @param _token2  reserve token to get the rate of one `_token1` unit in
+      *
+      * @return rate (numerator)
+      * @return rate (denominator)
+    */
```
barakman, line 109: This function is used only when the weights are equal, so it is redundant, as we can simply use the balances.
It is also harmful in a way, because it increases the size of N and D, which we need to keep bounded.

solidity/contracts/converter/types/liquidity-pool-v1/LiquidityPoolV1Converter.sol
```diff
@@ -155,6 +210,68 @@ contract LiquidityPoolV1Converter is LiquidityPoolConverter {
...
+        // get the elapsed time since the previous average rate was calculated
+        uint256 timeElapsed = time() - prevAverageRateUpdateTime;
+
+        // if the previous average rate was calculated in the current block, return it
+        if (timeElapsed == 0) {
+            return prevAverageRate;
+        }
+
+        // get the current rate between the reserves
+        (uint256 currentRateN, uint256 currentRateD) = tokensRate(reserveTokens[0], reserveTokens[1]);
```
barakman, line 251: Can change to:
```
// get the current rate between the reserve balances
uint256 currentRateN = reserves[reserveTokens[0]].balance;
uint256 currentRateD = reserves[reserveTokens[1]].balance;
```

barakman (changes requested):

solidity/contracts/converter/types/liquidity-pool-v1/LiquidityPoolV1Converter.sol
yudilevi: need to reverse these numbers but done

solidity/contracts/converter/types/liquidity-pool-v1/LiquidityPoolV1Converter.sol
yudilevi: done

solidity/contracts/converter/types/liquidity-pool-v1/LiquidityPoolV1Converter.sol
```diff
@@ -155,6 +189,69 @@ contract LiquidityPoolV1Converter is LiquidityPoolConverter {
...
+      * @dev returns the recent average rate of 1 `_token` in the other reserve token units
+      * note that the rate can only be queried for reserves in a standard pool
+      *
+      * @param _token   token to get the rate for
+      * @return recent average rate between the reserves (numerator)
+      * @return recent average rate between the reserves (denominator)
+    */
+    function recentAverageRate(IERC20Token _token) external view returns (uint256, uint256) {
+        // if the pool isn't standard, return 0, 0
+        if (!isStandardPool) {
```
barakman, line 202: Do we really want to do a storage read just in order to return this "invalid-anyway" value?
Why don't we just add a comment in the Natspec documentation above, stating that this function is designated only for standard pools?

solidity/contracts/helpers/TestLiquidityPoolV1Converter.sol
```diff
@@ -16,4 +18,12 @@ contract TestLiquidityPoolConverter is LiquidityPoolV1Converter {
...
     }
+
+    function time() internal view override returns (uint256) {
+        return currentTime != 0 ? currentTime : now;
+    }
+
+    function setTime(uint256 _currentTime) public {
+        currentTime = _currentTime;
+    }
 }
```
barakman, line 29: Due to `uint256 timeElapsed = time() - prevAverageRateUpdateTime` in the real contract, we would probably be better off here with:
```
    function increaseTime(uint256 _seconds) public {
        currentTime += _seconds;
    }
```
Because setting the current time to a value lower than `prevAverageRateUpdateTime` would cause the statement above to underflow.

barakman (changes requested):

solidity/contracts/converter/types/liquidity-pool-v1/LiquidityPoolV1Converter.sol
yudilevi: for the external version, I think it's better, otherwise you get undefined behavior

solidity/contracts/helpers/TestLiquidityPoolV1Converter.sol
yudilevi: unless the tests also update `prevAverageRateUpdateTime` accordingly - better keep this flexible.

solidity/contracts/converter/types/liquidity-pool-v1/LiquidityPoolV1Converter.sol
barakman: @yudilevi : So perhaps `require(isStandardPool, "ERR_MSG");` is better than returning rate = 0/0 (which is also pretty undefined).

barakman (changes requested):

solidity/contracts/converter/types/liquidity-pool-v1/LiquidityPoolV1Converter.sol
yudilevi: done

PR #432, opened 2020-09-13
Add function `addLiquidityCost`, for calculating the minimum amount o…

barakman:
Add function `addLiquidityCost`, for calculating the minimum amount of reserve tokens required of each reserve, given the amount of one of the reserves.

### DIFF
--- a/solidity/contracts/converter/types/liquidity-pool-v1/LiquidityPoolV1Converter.sol
+++ b/solidity/contracts/converter/types/liquidity-pool-v1/LiquidityPoolV1Converter.sol
@@ -436,6 +436,35 @@ contract LiquidityPoolV1Converter is LiquidityPoolConverter {
     }
 
     /**
+      * @dev given the amount of one of the reserve tokens to add liquidity of,
+      * calculates the required amount of each one of the other reserve tokens
+      * since an empty pool can be funded with any list of non-zero input amounts,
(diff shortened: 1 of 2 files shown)


solidity/contracts/converter/types/liquidity-pool-v1/LiquidityPoolV1Converter.sol
```diff
@@ -435,6 +435,35 @@ contract LiquidityPoolV1Converter is LiquidityPoolConverter {
...
+      * since an empty pool can be funded with any list of non-zero input amounts,
+      * this function assumes that the pool is not empty (has already been funded)
+      *
+      * @param _reserveTokens       address of each reserve token
+      * @param _reserveTokenIndex   index of the relevant reserve token
+      * @param _reserveAmount       amount of the relevant reserve token
+      *
+      * @return the required amount of each one of the reserve tokens
+    */
+    function addLiquidityCost(IERC20Token[] memory _reserveTokens, uint256 _reserveTokenIndex, uint256 _reserveAmount)
```
yudilevi, line 450: what if the pool is empty?

barakman: Then there is no restriction on the input amounts, hence no point in calling this function.
As a side-note - this function will revert upon calling the formula function with supply = 0.

PR #440, opened 2020-10-13
Fix internal function `recentAverageRate`

barakman:
If the previous average rate was never calculated, the average rate is equal to the current rate.
Note that after a pool is liquidated completely, the previous average rate may return to a 'never calculated' state, but there is no problem with that.


solidity/contracts/converter/types/liquidity-pool-v1/LiquidityPoolV1Converter.sol
```diff
@@ -242,6 +242,11 @@ contract LiquidityPoolV1Converter is LiquidityPoolConverter {
         // calculate the numerator and the denumerator of the new rate
         Fraction memory prevAverage = prevAverageRate;
 
+        // if the previous average rate was never calculated, the average rate is equal to the current rate
+        if (prevAverage.n == 0 && prevAverage.d == 0) {
```
yudilevi, line 246: better use || instead of && here

PR #448, opened 2020-10-25
0.6.23

barakman:


yudilevi (approved):

solidity/contracts/liquidity-protection/LiquidityProtection.sol
```diff
@@ -655,35 +657,11 @@ contract LiquidityProtection is TokenHandler, ContractRegistryClient, Reentrancy
...
-
-        // calculate the amount of pool tokens required for liquidation
-        // note that the amount is doubled since it's not possible to liquidate one reserve only
-        Fraction memory poolRate = poolTokenRate(liquidity.poolToken, liquidity.reserveToken);
-        uint256 poolAmount = targetAmount.mul(poolRate.d).mul(2).div(poolRate.n);
-
-        // limit the amount of pool tokens by the amount the system/caller holds
-        uint256 availableBalance = store.systemBalance(liquidity.poolToken).add(liquidity.poolAmount);
-        poolAmount = poolAmount > availableBalance ? availableBalance : poolAmount;
+        // calculate the amount of base tokens required for liquidation
```
yudilevi, line 660: I don't understand this comment - base tokens can't be liquidated, only pool tokens can

solidity/contracts/liquidity-protection/LiquidityProtection.sol
```diff
@@ -852,6 +815,68 @@ contract LiquidityProtection is TokenHandler, ContractRegistryClient, Reentrancy
...
     }
 
+    /**
+      * @dev returns the amount of tokens required for liquidation
+      *
+      * @param _poolToken       pool token
+      * @param _reserveToken    reserve token
+      * @param _poolAmount      pool token amount when the liquidity was added
+      * @param _targetAmount    target token amount received in return
+      * @param _isBaseAmount    true if the required amount is of the based token
```
yudilevi, line 825: the required amount is in pool tokens so this comment is confusing as there's no required amount in base tokens

solidity/contracts/liquidity-protection/LiquidityProtection.sol
```diff
@@ -852,6 +815,68 @@ contract LiquidityProtection is TokenHandler, ContractRegistryClient, Reentrancy
...
+    */
+    function liquidationAmount(
+        IDSToken _poolToken,
+        IERC20Token _reserveToken,
+        uint256 _poolAmount,
+        uint256 _targetAmount,
+        bool _isBaseAmount)
+        internal view returns (uint256)
+    {
+        // calculate the amount of pool tokens required for liquidation
```
yudilevi, line 836: this is the same as the function description so it's redundant. instead I'd just give context to the next line.

solidity/contracts/liquidity-protection/LiquidityProtection.sol
```diff
@@ -852,6 +815,68 @@ contract LiquidityProtection is TokenHandler, ContractRegistryClient, Reentrancy
...
+        uint256 poolAmount = _targetAmount.mul(poolRate.d).div(poolRate.n / 2);
+
+        // limit the amount of pool tokens by the amount that the system holds
+        uint256 systemBalance = store.systemBalance(_poolToken).add(_poolAmount);
+        poolAmount = poolAmount > systemBalance ? systemBalance : poolAmount;
+
+        if (_isBaseAmount) {
+            // calculate the base token amount received by liquidating the pool tokens
+            // note that the amount is divided by 2 since the pool amount represents both reserves
+            return poolAmount.mul(poolRate.n / 2).div(poolRate.d);
```
yudilevi, line 848: seems like the function return base token amount in this case and not pool token amount while it does return pool token amount in the last line (and that's what the description says).
I think we should aim for more consistency here, 2 different return values that also represent different meanings is a bit too much.

solidity/contracts/liquidity-protection/LiquidityProtection.sol
```diff
@@ -852,6 +815,68 @@ contract LiquidityProtection is TokenHandler, ContractRegistryClient, Reentrancy
...
+        bool _isBaseAmount)
+        internal view returns (uint256)
+    {
+        // calculate the amount of pool tokens required for liquidation
+        // note that the amount is doubled since it's not possible to liquidate one reserve only
+        Fraction memory poolRate = poolTokenRate(_poolToken, _reserveToken);
+        uint256 poolAmount = _targetAmount.mul(poolRate.d).div(poolRate.n / 2);
+
+        // limit the amount of pool tokens by the amount that the system holds
+        uint256 systemBalance = store.systemBalance(_poolToken).add(_poolAmount);
```
yudilevi, line 842: I understand why we add the pool amount to the system balance here but only because I'm familiar with the previous version of the code - it's really unclear in the context of the function itself - we take the amount that was initially added (based on the description) and we add it to the system balance and then limit the pool amount by that value? why would we limit it by a number that - in the context of the function - looks like is higher than the total number that the system holds?

solidity/contracts/liquidity-protection/LiquidityProtection.sol
```diff
@@ -751,16 +729,8 @@ contract LiquidityProtection is TokenHandler, ContractRegistryClient, Reentrancy
...
-
         // calculate the amount of pool tokens required for liquidation
-        // note that the amount is doubled since it's not possible to liquidate one reserve only
-        Fraction memory poolRate = poolTokenRate(liquidity.poolToken, liquidity.reserveToken);
-        uint256 poolAmount = targetAmount.mul(poolRate.d).mul(2).div(poolRate.n);
-
-        // limit the amount of pool tokens by the amount the system holds
-        uint256 systemBalance = store.systemBalance(liquidity.poolToken);
-        poolAmount = poolAmount > systemBalance ? systemBalance : poolAmount;
+        uint256 poolAmount = liquidationAmount(liquidity.poolToken, liquidity.reserveToken, 0, targetAmount, false);
```
yudilevi, line 733: notice that in the previous version of this function, we didn't add the pool amount to the system balance, while `liquidationAmount` does it, which will produce different results.

yudilevi (changes requested):

solidity/contracts/liquidity-protection/LiquidityProtection.sol
barakman: Can you please fix this comment then?

solidity/contracts/liquidity-protection/LiquidityProtection.sol
barakman: Can you please fix this comment then?

solidity/contracts/liquidity-protection/LiquidityProtection.sol
barakman: These are not two different return values.
It's the same return value depicted in different units.
The input argument `_isBaseAmount` indicates whether the return value should be in base token units or not (in which case, they should be in pool token units).

yudilevi: it's highly unconventional nonetheless, and even got me confused

solidity/contracts/liquidity-protection/LiquidityProtection.sol
barakman: This nevertheless does the same thing that the previous code version did, with the intention of reusing code in favor of reducing code size. If there is no functional error in this new version, can you please point out what exactly it is that you want to fix, or better yet (assuming it's essentially cosmetic) - fix it yourself? Thanks.

solidity/contracts/liquidity-protection/LiquidityProtection.sol
barakman: We actually did. Please check the previous version, you'll se it there.

solidity/contracts/liquidity-protection/LiquidityProtection.sol
yudilevi: check the `removeLiquidity` function in the previous version, it doesn't add the pool amount to the system balance -

```
// limit the amount of pool tokens by the amount the system holds
uint256 systemBalance = store.systemBalance(liquidity.poolToken);
poolAmount = poolAmount > systemBalance ? systemBalance : poolAmount;
```

functional error as far as I can tell.

solidity/contracts/liquidity-protection/LiquidityProtection.sol
yudilevi: see my comment above re. the functional error.

solidity/contracts/liquidity-protection/LiquidityProtection.sol
barakman: It does in the other place, and I am passing `0` in this case.

solidity/contracts/liquidity-protection/LiquidityProtection.sol
barakman: See mine too.

solidity/contracts/converter/ConverterBase.sol
```diff
@@ -430,23 +423,13 @@ abstract contract ConverterBase is IConverter, TokenHandler, TokenHolder, Contra
     */
     function reserveBalance(IERC20Token _reserveToken)
         public
```
yudilevi, line 425: how come this isn't an override anymore? it should be a part of the interface as far as I can tell

solidity/contracts/converter/interfaces/IConverter.sol
```diff
@@ -20,16 +19,13 @@ interface IConverter is IOwned {
                      address _trader,
                      address payable _beneficiary) external payable returns (uint256);
 
-    function conversionWhitelist() external view returns (IWhitelist);
     function conversionFee() external view returns (uint32);
     function maxConversionFee() external view returns (uint32);
-    function reserveBalance(IERC20Token _reserveToken) external view returns (uint256);
```
yudilevi, line 26: this should be part of the interface even though it's not used internally by the contracts - only reason it's not is the backward compatibility

solidity/contracts/liquidity-protection/LiquidityProtection.sol
```diff
@@ -21,7 +21,7 @@ interface ILiquidityPoolV1Converter is IConverter {
 }
 
 /**
-  * @dev Liquidity Protection
+  * @dev This contract provides the liquidity protection mechanism.
```
yudilevi, line 24: `implements` the liquidity protection mechanism?

solidity/contracts/liquidity-protection/LiquidityProtection.sol
```diff
@@ -65,7 +65,7 @@ contract LiquidityProtection is TokenHandler, ContractRegistryClient, Reentrancy
     // number of seconds from liquidation to full network token release
     uint256 public lockDuration = 24 hours;
 
-    // maximum deviation of the average rate from the actual rate
+    // maximum deviation of the average rate from the spot rate
     uint32 public averageRateMaxDeviation = 20000; // PPM units
```
yudilevi, line 69: changed to 0.5% in the LP branch?

solidity/contracts/liquidity-protection/LiquidityProtection.sol
```diff
@@ -1018,28 +1034,48 @@ contract LiquidityProtection is TokenHandler, ContractRegistryClient, Reentrancy
...
+      * @dev returns whether or not the deviation of the average rate from the spot rate is within range
+      *
+      * @param _spotRateN       spot rate numerator
+      * @param _spotRateD       spot rate denominator
+      * @param _averageRateN    average rate numerator
+      * @param _averageRateD    average rate denominator
+      * @param _maxDeviation    the maximum permitted deviation of the average rate from the spot rate
+    */
+    function averageRateInRange(
+        uint256 _spotRateN,
```
yudilevi, line 1068: can use Fraction for both here

barakman: Aborting this, right?

yudilevi: yeah

PR #449, opened 2020-10-27
Revise function `removeLiquidityTargetAmount`.

barakman:


solidity/contracts/liquidity-protection/LiquidityProtection.sol
```diff
@@ -845,27 +849,30 @@ contract LiquidityProtection is TokenHandler, ContractRegistryClient, Reentrancy
...
     {
-        // get the adjusted amount of reserve tokens based on the exposure and rate changes
-        uint256 outputAmount = adjustedAmount(_poolToken, _reserveToken, _poolAmount, _addRate, _removeRate);
+        // get the rate between the reserves upon adding liquidity and now
+        Fraction memory addSpotRate = Fraction({n: _packedRates.addSpotRateN, d: _packedRates.addSpotRateD});
+        Fraction memory removeSpotRate = Fraction({n: _packedRates.removeSpotRateN, d: _packedRates.removeSpotRateD});
+        Fraction memory removeAverageRate = Fraction({n: _packedRates.removeAverageRateN, d: _packedRates.removeAverageRateD});
+
+        // calculate the entitled amount of reserve tokens before compensation
+        uint256 entitlement = entitledAmount(_poolToken, _reserveToken, _poolAmount, addSpotRate, removeSpotRate);
```
yudilevi, line 863: output amount instead of entitlement? totalAmount? reserveAmountPlusFees?

solidity/contracts/liquidity-protection/LiquidityProtection.sol
```diff
@@ -1040,6 +1045,17 @@ contract LiquidityProtection is TokenHandler, ContractRegistryClient, Reentrancy
       * @param _reserveToken    reserve token
     */
     function reserveTokenAverageRate(IDSToken _poolToken, IERC20Token _reserveToken) internal view returns (Fraction memory) {
+        (, , uint256 averageRateN, uint256 averageRateD) = reserveTokenRates(_poolToken, _reserveToken);
```
yudilevi, line 1048: this will increase gas costs

solidity/contracts/liquidity-protection/LiquidityProtection.sol
```diff
@@ -1052,7 +1068,39 @@ contract LiquidityProtection is TokenHandler, ContractRegistryClient, Reentrancy
...
+    /**
+      * @dev returns a packed object containing the reserves rates
+      *
+      * @param _poolToken       pool token
+      * @param _reserveToken    reserve token
+      * @param _addSpotRateN    add spot rate numerator
+      * @param _addSpotRateD    add spot rate denominator
+      * @return packed object containing the reserves rates
+    */
+    function getPackedRates(
```
yudilevi, line 1083: `get` again

solidity/contracts/liquidity-protection/LiquidityProtection.sol
```diff
@@ -1163,16 +1211,16 @@ contract LiquidityProtection is TokenHandler, ContractRegistryClient, Reentrancy
     }
 
     /**
-      * @dev returns the adjusted amount of reserve tokens based on the exposure and rate changes
+      * @dev returns the entitled amount of reserve tokens before compensation
```
yudilevi, line 1214: if this returns the initial reserve amount + the fees, that's probably how we should define/describe it

solidity/contracts/liquidity-protection/LiquidityProtection.sol
```diff
@@ -1181,12 +1229,17 @@ contract LiquidityProtection is TokenHandler, ContractRegistryClient, Reentrancy
         internal view returns (uint256)
     {
         Fraction memory poolRate = poolTokenRate(_poolToken, _reserveToken);
-        Fraction memory poolFactor = poolTokensFactor(_addRate, _removeRate);
+        uint256 n = Math. ceilSqrt(_addRate.d.mul(_removeRate.n)).mul(poolRate.n);
```
yudilevi, line 1232: extra space between Math. and ceilSqrt

solidity/contracts/liquidity-protection/LiquidityProtection.sol
```diff
@@ -1181,12 +1229,17 @@ contract LiquidityProtection is TokenHandler, ContractRegistryClient, Reentrancy
...
+        uint256 d = Math.floorSqrt(_addRate.n.mul(_removeRate.d)).mul(poolRate.d);
 
-        (uint256 poolRateN, uint256 poolRateD) = Math.reducedRatio(_poolAmount.mul(poolRate.n), poolRate.d, MAX_UINT128);
-        (uint256 poolFactorN, uint256 poolFactorD) = Math.reducedRatio(poolFactor.n, poolFactor.d, MAX_UINT128);
+        uint256 x = n * _poolAmount;
```
yudilevi, line 1235: can't overflow?

solidity/contracts/liquidity-protection/LiquidityProtection.sol
```diff
@@ -1181,12 +1229,17 @@ contract LiquidityProtection is TokenHandler, ContractRegistryClient, Reentrancy
...
+        if (x / n == _poolAmount) {
```
yudilevi, line 1236: `n` can't be 0 (because of rounding for example)?

solidity/contracts/liquidity-protection/LiquidityProtection.sol
```diff
@@ -1181,12 +1229,17 @@ contract LiquidityProtection is TokenHandler, ContractRegistryClient, Reentrancy
...
+            return x / d;
```
yudilevi, line 1237: `d` can't be 0 (because of rounding for example)?

solidity/contracts/liquidity-protection/LiquidityProtection.sol
```diff
@@ -1181,12 +1229,17 @@ contract LiquidityProtection is TokenHandler, ContractRegistryClient, Reentrancy
...
+        }
 
-        return poolRateN.mul(poolFactorN).div(poolRateD.mul(poolFactorD));
+        (uint256 hi, uint256 lo) = n > _poolAmount ? (n, _poolAmount) : (_poolAmount, n);
+        (uint256 p, uint256 q) = Math.reducedRatio(hi, d, uint256(-1) / lo);
```
yudilevi, line 1241: `q` can't be 0 (because of rounding for example)?

yudilevi (changes requested):

solidity/contracts/liquidity-protection/LiquidityProtection.sol
```diff
@@ -834,8 +839,7 @@ contract LiquidityProtection is TokenHandler, ContractRegistryClient, Reentrancy
       * @param _reserveToken    reserve token
       * @param _poolAmount      pool token amount when the liquidity was added
       * @param _reserveAmount   reserve token amount that was added
-      * @param _addRate         rate of 1 reserve token in the other reserve token units when the liquidity was added
-      * @param _removeRate      rate of 1 reserve token in the other reserve token units when the liquidity is removed
+      * @param _packedRates     packed object containing the reserves rates
```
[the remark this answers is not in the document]
barakman: There are now 3 rates here, given by 6 variables.
How exactly do you want to accomplish this?

solidity/contracts/liquidity-protection/LiquidityProtection.sol
barakman: A local variable by the name `totalAmount` is used in various places, so we cannot use this one.
Please specify what name you want for the variable and what name you want for the function.

solidity/contracts/liquidity-protection/LiquidityProtection.sol
barakman: How so? The other two variables (spot rate) are used internally in order to verify the average rate's maximum permitted deviation. This is exactly why I've split that function into 2 or 3 functions.

solidity/contracts/liquidity-protection/LiquidityProtection.sol
barakman: Desired name please (function and variable, where the function is called).

solidity/contracts/liquidity-protection/LiquidityProtection.sol
barakman: Of course it can. That's why we have that `if` statement which follows.

solidity/contracts/liquidity-protection/LiquidityProtection.sol
barakman: No.
Functions `floorSqrt` and `ceilSqrt` are implemented for positive input only (and documented as such).
For the smallest legal input (1), the output is larger than 0 (also 1).
So a divide-by-zero revert (invalid opcode) can take place here only if the input (one of the rates) is 0, which we know is not the case (and if it was the case, then the transaction would revert at a much earlier point).

solidity/contracts/liquidity-protection/LiquidityProtection.sol
barakman: No.
Functions `floorSqrt` and `ceilSqrt` are implemented for positive input only (and documented as such).
For the smallest legal input (1), the output is larger than 0 (also 1).
So a divide-by-zero revert (invalid opcode) can take place here only if the input (one of the rates) is 0, which we know is not the case (and if it was the case, then the transaction would revert at a much earlier point).

solidity/contracts/liquidity-protection/LiquidityProtection.sol
barakman: No.
Function `reducedRatio` returns a pair of positive integers.

solidity/contracts/liquidity-protection/LiquidityProtection.sol
barakman: Was hoping you'd forget about it; fixed.

solidity/contracts/liquidity-protection/LiquidityProtection.sol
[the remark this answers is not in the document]
yudilevi: they all go in the same direction

solidity/contracts/liquidity-protection/LiquidityProtection.sol
```diff
@@ -38,6 +38,15 @@ contract LiquidityProtection is TokenHandler, ContractRegistryClient, Reentrancy
         uint256 timestamp;          // timestamp
     }
 
+    struct PackedRates {
+        uint128 addSpotRateN;
```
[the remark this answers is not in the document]
yudilevi: did you consider using uint128 for fractions?

solidity/contracts/liquidity-protection/LiquidityProtection.sol
yudilevi: reserveAmountPlusFees then

solidity/contracts/liquidity-protection/LiquidityProtection.sol
yudilevi: the cost of `reserveTokenAverageRate` is increased since we call `reserveTokenRates` that gets both versions of the rates and we only use one

solidity/contracts/liquidity-protection/LiquidityProtection.sol
yudilevi: figure it out according to our standards

solidity/contracts/liquidity-protection/LiquidityProtection.sol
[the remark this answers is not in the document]
barakman: We can't do that of course, because we're using `uint256` based Fractions everywhere.

solidity/contracts/liquidity-protection/LiquidityProtection.sol
barakman: It uses both, please see my comment above.

solidity/contracts/liquidity-protection/LiquidityProtection.sol
barakman: Named it `protectedAmountPlusFee`.

solidity/contracts/liquidity-protection/LiquidityProtection.sol

solidity/contracts/liquidity-protection/LiquidityProtection.sol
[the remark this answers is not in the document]
yudilevi: the Fraction struct is only used inside contracts, not between contracts and also not in storage.
we could choose to use 128 bit based fraction in the liquidity protection which would have saved us some headaches, but only if it would actually work, since I know also multiply them by one another etc.
so not sure if it's practical but just a suggestion.

solidity/contracts/liquidity-protection/LiquidityProtection.sol
yudilevi: the reason why this isn't an issue is that when we only want one of them (spot OR average - just like in this specific case), we anyway need to get both in order to verify the max deviation.

solidity/contracts/liquidity-protection/LiquidityProtection.sol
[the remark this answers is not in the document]

solidity/contracts/liquidity-protection/LiquidityProtection.sol
barakman: `reserveTokenRates` is called once from the `protectLiquidity` flow (via `reserveTokenAverageRate`), and once from the `removeLiquidity` flow (via `packRates`).

solidity/contracts/liquidity-protection/LiquidityProtection.sol

solidity/contracts/liquidity-protection/LiquidityProtection.sol
```diff
@@ -1052,7 +1069,39 @@ contract LiquidityProtection is TokenHandler, ContractRegistryClient, Reentrancy
...
+        IERC20Token _reserveToken,
+        uint256 _addSpotRateN,
+        uint256 _addSpotRateD)
+        internal view returns (PackedRates memory)
+    {
+        (uint256 removeSpotRateN, uint256 removeSpotRateD, uint256 removeAverageRateN, uint256 removeAverageRateD) = reserveTokenRates(_poolToken, _reserveToken);
+
+        require(_addSpotRateN <= MAX_UINT128 && _addSpotRateD <= MAX_UINT128, "ERR_INVALID_RATE");
+        require(removeSpotRateN <= MAX_UINT128 && removeSpotRateD <= MAX_UINT128, "ERR_INVALID_RATE");
+        require(removeAverageRateN <= MAX_UINT128 && removeAverageRateD <= MAX_UINT128, "ERR_INVALID_RATE");
```
lbeder, line 1095: Perhaps combining this into a single `require` will save some gas?

solidity/test/LiquidityProtectionEdgeCases.js
```diff
@@ -44,7 +44,7 @@ const CONFIGURATIONS = [
 ];
 
 const NUM_OF_DAYS = [30, 100];
-const DECIMAL_COMBINATIONS = cartesian([12, 25], [12, 25], [15, 22], [15, 22]);
+const DECIMAL_COMBINATIONS = cartesian([12, 24], [12, 24], [15, 21], [15, 21]);
```
lbeder, line 47: @barakman can you also add the edge case of 6 decimals (USDT, USDC) trading against 18 decimals (BNT)?

barakman: I can, but it's gonna be a little complicated, because I'd still need to avoid the combination of 6 and 24, which will surely not work.

solidity/contracts/liquidity-protection/LiquidityProtection.sol
lbeder: It won't be a significant save, but since the revert message is the same for all of the cases - it won't hurt

solidity/test/LiquidityProtectionEdgeCases.js
lbeder: You can add an `if` to skip this combination or to push it explicitly to `DECIMAL_COMBINATIONS` afterwards

solidity/contracts/liquidity-protection/LiquidityProtection.sol
yudilevi: yup, just a bit cleaner and 2 seconds to do

solidity/test/LiquidityProtectionEdgeCases.js
yudilevi: any of these options work

barakman: This requires a more sophisticated (or at least a different) test.
Just changing the numbers from "12,24" to "6,18" is not going to achieve a lot.

At present, this test does the following:
- Add an initial amount of 10^12 or 10^24 BNT and 10^12 or 10^24 TKN
- Protect an additional amount of 10^15 or 10^21 BNT and 10^15 or 10^21 TKN
- Increase the rate of BNT or TKN or neither (via converting in one direction)
- Generate a fee of BNT or TKN or neither (via converting back and forth)
- Wait for 30 or 100 days
- Call `removeLiquidityReturn` and make sure that the result makes sense

The level of extremeness is generally determined by **the ratio** between the amounts.
So reducing each one of them by a magnitude of 10^6 is not going to make a big impact.

PR #456, opened 2020-11-07
Reduce gas-cost of adding/removing liquidity

barakman:
At the end of function `verifyLiquidityInput`, we rearrange the input arrays according to the converter's array of reserve tokens.
Then, all subsequent functions can assume that the ID of the 1st input token is 1 and the ID of the 2nd input token is 2.

### DIFF
--- a/solidity/contracts/converter/types/standard-pool/StandardPoolConverter.sol
+++ b/solidity/contracts/converter/types/standard-pool/StandardPoolConverter.sol
@@ -858,11 +858,9 @@ contract StandardPoolConverter is ConverterVersion, IConverter, TokenHandler, To
 
         // iterate through the reserve tokens and transfer a percentage equal to the weight between
         // _amount and the total supply in each reserve from the caller to the converter
-        uint256 reserveCount = __reserveTokens.length;
-        for (uint256 i = 0; i < reserveCount; i++) {
+        for (uint256 i = 0; i < 2; i++) {
             IERC20Token reserveToken = __reserveTokens[i];
-            uint256 reserveId = __reserveIds[reserveToken];
-            uint256 rsvBalance = reserveBalance(reserveId);
+            uint256 rsvBalance = reserveBalance(i + 1);
             uint256 reserveAmount = fundCost(supply, rsvBalance, _amount);
 
             // transfer funds from the caller in the reserve token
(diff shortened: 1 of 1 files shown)


solidity/contracts/converter/types/standard-pool/StandardPoolConverter.sol
```diff
@@ -1067,14 +1061,13 @@ contract StandardPoolConverter is ConverterVersion, IConverter, TokenHandler, To
...
             IERC20Token reserveToken = _reserveTokens[i];
-            uint256 reserveId = __reserveIds[reserveToken];
             uint256 reserveAmount = _reserveAmounts[i];
 
             if (reserveToken != ETH_RESERVE_ADDRESS) { // ETH has already been transferred as part of the transaction
                 safeTransferFrom(reserveToken, msg.sender, address(this), reserveAmount);
             }
 
-            setReserveBalance(reserveId, reserveAmount);
+            setReserveBalance(i + 1, reserveAmount);
```
yudilevi, line 1070: we could set the reserve balance for both at the same time

solidity/contracts/converter/types/standard-pool/StandardPoolConverter.sol
```diff
@@ -880,7 +878,7 @@ contract StandardPoolConverter is ConverterVersion, IConverter, TokenHandler, To
 
             // sync the reserve balance
             uint256 newReserveBalance = rsvBalance.add(reserveAmount);
-            setReserveBalance(reserveId, newReserveBalance);
+            setReserveBalance(i + 1, newReserveBalance);
```
yudilevi, line 881: we could set the reserve balance for both at the time time

solidity/contracts/converter/types/standard-pool/StandardPoolConverter.sol
```diff
@@ -1107,8 +1100,7 @@ contract StandardPoolConverter is ConverterVersion, IConverter, TokenHandler, To
 
         for (uint256 i = 0; i < _reserveTokens.length; i++) {
```
yudilevi, line 1101: it's time to merge the add liquidity to empty/non empty pool functions - they are basically the same function, the only difference is the calculated balance - this is where we should have the differentiation.

solidity/contracts/converter/types/standard-pool/StandardPoolConverter.sol
```diff
@@ -1179,9 +1171,8 @@ contract StandardPoolConverter is ConverterVersion, IConverter, TokenHandler, To
             uint256 reserveAmount = reserveAmounts[i];
             require(reserveAmount >= _reserveMinReturnAmounts[i], "ERR_ZERO_TARGET_AMOUNT");
 
-            uint256 reserveId = __reserveIds[reserveToken];
-            uint256 newReserveBalance = reserveBalance(reserveId).sub(reserveAmount);
-            setReserveBalance(reserveId, newReserveBalance);
+            uint256 newReserveBalance = reserveBalance(i + 1).sub(reserveAmount);
```
yudilevi, line 1174: can read both reserve balances in a single call

PR #465, opened 2020-11-12
Function `removeLiquidityTargetAmount`

barakman:
Add an extensive test for internal function `removeLiquidityTargetAmount`, and subsequently:
- In internal function `protectedAmountPlusFee`, avoid possible division by zero in extreme cases
- In internal function `compensationAmount`, avoid possible multiplication overflow in extreme cases


solidity/contracts/liquidity-protection/LiquidityProtection.sol
```diff
@@ -1361,8 +1369,13 @@ contract LiquidityProtection is TokenHandler, ContractRegistryClient, Reentrancy
         Fraction memory _loss,
         Fraction memory _level
     ) internal pure returns (uint256) {
-        (uint256 lossN, uint256 lossD) = Math.reducedRatio(_loss.n, _loss.d, MAX_UINT128);
-        return _total.mul(lossD.sub(lossN)).div(lossD).add(_amount.mul(lossN.mul(_level.n)).div(lossD.mul(_level.d)));
+        uint256 levelN = _level.n.mul(_amount);
+        uint256 levelD = _level.d;
+        uint256 max = _total;
+        if (max < levelN) max = levelN;
```
lbeder, line 1375: We really want to avoid this style. It's harder to read and easy to miss out

solidity/test/LiquidityProtectionTargetAmount.js
```diff
@@ -0,0 +1,181 @@
+const { expect } = require('chai');
```
lbeder, line 1: Can you move this to one of the existing test files? There are already so many LP test files and it makes it harder to test

solidity/test/LiquidityProtectionTargetAmount.js
```diff
@@ -0,0 +1,181 @@
...
+            for (const poolTokenRateD of amounts) {
+                for (const poolAmount of amounts) {
+                    for (const reserveAmount of amounts) {
+                        for (const addSpotRateN of amounts) {
+                            for (const addSpotRateD of amounts) {
+                                for (const removeSpotRateN of amounts) {
+                                    for (const removeSpotRateD of amounts) {
+                                        for (const removeAverageRateN of amounts) {
+                                            for (const removeAverageRateD of amounts) {
+                                                for (const timeElapsed of durations) {
```
lbeder, line 76: Mother of all tests!

barakman: MOAT???

solidity/contracts/liquidity-protection/LiquidityProtection.sol
barakman: Yeah, we technically need a `max` function in this file (useful 2 times in this function, and 1 more time in another function).

solidity/test/LiquidityProtectionTargetAmount.js
barakman: Couldn't find anyone really suitable.

solidity/contracts/liquidity-protection/LiquidityProtection.sol
barakman: But I wanted to minimize the number of visible changes in this PR, in order to make it clearer.
Then, adding a `max` function separately would be a pretty cosmetic change by itself.

lbeder: I meant that same-line bracketless `if` 😅

solidity/test/LiquidityProtectionTargetAmount.js
lbeder: Definitely a moat...

solidity/contracts/liquidity-protection/LiquidityProtection.sol
barakman: Yeah, me too. Those two lines should be reformatted as `uint256 maxVal = max(max(levelN, levelD), _total);`.

barakman: Fixed; please review.

drumenbassik (approved):

drumenbassik (approved):

drumenbassik (approved):

drumenbassik (approved):

drumenbassik (approved):

solidity/test/LiquidityProtectionTargetAmount.js
barakman: In addition to that, this test relies on a very simple one-time initialization step, in order to speed up testing of the actual feature at stake, whereas all other test files of this contract already utilize more complex initialization which would be redundant in this case.
Finally, I don't see any real benefit in having fewer files, where each file is larger. If anything, it makes it harder to maintain them (just like you would try to avoid writing too long functions, classes, etc).
Can you please explain the benefit of pouring the contents of this file into another file in this case?

lbeder: You can use scopes to have dozens of different initialization logic in the same file. The rationale behind keeping the same component tests in the same file (other than being the de-facto standard) is that it allows much easier management (e.g., running all X tests at the same time, configuring watchers that if Y changes - run X again, etc.). In addition to that, the same component tests can usually reuse similar code (which is heavily duplicated today).

solidity/contracts/liquidity-protection/LiquidityProtection.sol
```diff
@@ -1388,8 +1386,15 @@ contract LiquidityProtection is TokenHandler, ContractRegistryClient, Reentrancy
...
 
         (uint256 hi, uint256 lo) = n > _poolAmount ? (n, _poolAmount) : (_poolAmount, n);
-        (uint256 p, uint256 q) = Math.reducedRatio(hi, d, uint256(-1) / lo);
-        return (p * lo) / q;
+        (uint256 p, uint256 q) = Math.reducedRatio(hi, d, MAX_UINT256 / lo);
+
+        if (q != 0) {
+            return p * lo / q;
+        }
+        if (p == 0) {
```
yudilevi, line 1394: can you explain this change? it seems like a logical change in the function return value

solidity/contracts/liquidity-protection/LiquidityProtection.sol
```diff
@@ -1471,6 +1485,16 @@ contract LiquidityProtection is TokenHandler, ContractRegistryClient, Reentrancy
         return 0;
     }
 
+    /**
+     * @dev returns the larger of two values
+     *
+     * @param _val1 the first value
+     * @param _val2 the second value
+     */
+    function max(uint256 _val1, uint256 _val2) internal pure returns (uint256) {
```
yudilevi, line 1494: this should be in Math.sol

solidity/contracts/utility/Math.sol
```diff
@@ -56,8 +56,10 @@ library Math {
         uint256 _d,
         uint256 _max
     ) internal pure returns (uint256, uint256) {
-        if (_n > _max || _d > _max) return normalizedRatio(_n, _d, _max);
-        return (_n, _d);
+        (uint256 n, uint256 d) = (_n, _d);
+        if (n > _max || d > _max) (n, d) = normalizedRatio(n, d, _max);
```
yudilevi, line 60: 1. what is this logical change?
2. missing curly brackets etc.

solidity/test/LiquidityProtectionTargetAmount.js
yudilevi: @barakman move to one of the other files, too many of them, it's becoming unmanageable to run all and only liquidity protection tests

yudilevi (changes requested):

solidity/contracts/liquidity-protection/LiquidityProtection.sol
barakman: Yes:
- The `if (q != 0)` part is obvious (avoid possible division by zero and revert without gas refund)
- The `if (p == 0)` part, now that `q == 0` is used here in order to emulate `0/0 == 1`. In other words, when `p == q`, we can just return `lo`
- The default return statement takes place when `p > 0` and `q == 0`, in which case, we divide the positive number `p * lo` (it is positive because we also know that `lo > 0`) by `0`, hence returning the closest thing to infinity, which is `MAX_UINT256`

solidity/contracts/liquidity-protection/LiquidityProtection.sol
barakman: True, but we will have no `min` function there.

solidity/contracts/utility/Math.sol
barakman: 1. This is just used as a preliminary step to the logical change in this function, of returning the smallest possible result when the numerator and the denominator are equal (i.e., `1, 1`).

2. Was also missing before. I wanted to keep it that way in order to minimize visible changes during code review.

solidity/contracts/liquidity-protection/LiquidityProtection.sol
yudilevi: but in case of any issue, returning `MAX_UINT256` from `protectedAmountPlusFee` is going to lead to a disaster, so why would we fallback to that value?

solidity/contracts/liquidity-protection/LiquidityProtection.sol
yudilevi: and why is that an issue? we don't need `min` right now

barakman: Fixed

solidity/contracts/liquidity-protection/LiquidityProtection.sol

solidity/test/LiquidityProtectionTargetAmount.js
barakman: Done.

solidity/test/LiquidityProtectionTargetAmount.js
barakman: Now it's even moater...

solidity/contracts/liquidity-protection/LiquidityProtection.sol
barakman: Fixed again.

yudilevi (approved):

PR #469, opened 2020-11-13
Standard Pool

barakman:


solidity/contracts/helpers/TestConverterRegistry.sol
```diff
@@ -7,8 +7,11 @@ import "../converter/ConverterRegistry.sol";
 */
 contract TestConverterRegistry is ConverterRegistry {
     IConverter public createdConverter;
+    bool public typeChangingEnabled;
 
-    constructor(IContractRegistry _registry) public ConverterRegistry(_registry) {}
+    constructor(IContractRegistry _registry) public ConverterRegistry(_registry) {
+        typeChangingEnabled = true;
```
lbeder, line 13: You can set `typeChangingEnabled` to true directly in its definition above

solidity/contracts/helpers/TestStandardPoolConverter.sol
```diff
@@ -0,0 +1,30 @@
+// SPDX-License-Identifier: SEE LICENSE IN LICENSE
+pragma solidity 0.6.12;
+import "../converter/types/standard-pool/StandardPoolConverter.sol";
+
+contract TestStandardPoolConverter is StandardPoolConverter {
+    uint256 public currentTime = 1;
```
lbeder, line 6: We would need to refactor it with the `TestTime` change once it's merged

solidity/contracts/liquidity-protection/LiquidityProtection.sol
```diff
@@ -30,6 +30,18 @@ interface ILiquidityPoolV1Converter is IConverter {
         uint256[] memory _reserveMinReturnAmounts
     ) external;
 
+    function addLiquidity(
+        IERC20Token[2] memory _reserveTokens,
+        uint256[2] memory _reserveAmounts,
+        uint256 _minReturn
+    ) external payable;
```
lbeder, line 37: It's important that both public `addLiquidity` and `removeLiquidity` methods returns positions ids for the LM implementation

solidity/contracts/liquidity-protection/LiquidityProtection.sol
```diff
@@ -1295,16 +1307,28 @@ contract LiquidityProtection is TokenHandler, ContractRegistryClient, Reentrancy
...
+            amounts[1] = _reserveAmount2;
+            _converter.addLiquidity{ value: _value }(reserveTokens, amounts, 1);
+        } else {
+            IERC20Token[] memory reserveTokens = new IERC20Token[](2);
+            uint256[] memory amounts = new uint256[](2);
+            reserveTokens[0] = _reserveToken1;
+            reserveTokens[1] = _reserveToken2;
+            amounts[0] = _reserveAmount1;
+            amounts[1] = _reserveAmount2;
+            _converter.addLiquidity{ value: _value }(reserveTokens, amounts, 1);
```
lbeder, line 1328: This block looks like a code duplication

solidity/contracts/liquidity-protection/LiquidityProtection.sol
```diff
@@ -1324,16 +1348,28 @@ contract LiquidityProtection is TokenHandler, ContractRegistryClient, Reentrancy
...
+            minReturns[1] = 1;
+            converter.removeLiquidity(_poolAmount, reserveTokens, minReturns);
+        } else {
+            IERC20Token[] memory reserveTokens = new IERC20Token[](2);
+            uint256[] memory minReturns = new uint256[](2);
+            reserveTokens[0] = _reserveToken1;
+            reserveTokens[1] = _reserveToken2;
+            minReturns[0] = 1;
+            minReturns[1] = 1;
+            converter.removeLiquidity(_poolAmount, reserveTokens, minReturns);
```
lbeder, line 1369: Same here

lbeder (changes requested):

solidity/contracts/helpers/TestConverterRegistry.sol
barakman: True, but I figured it's a bad practice when a non-default (empty) constructor is implemented, since it is not clear when the initialization of the state-variable takes place.
Recommendation?

solidity/contracts/helpers/TestStandardPoolConverter.sol
barakman: True.

solidity/contracts/helpers/TestConverterRegistry.sol
lbeder: I'd just initialize it in the definition and leave the constructor empty

solidity/contracts/liquidity-protection/LiquidityProtection.sol
barakman: And is it important to do this prior the LM implementation, in order to avoid breaking the API?
In this case, is LM designated for SP converters only? Because we already have many other converters deployed, so if this is not the case, then we'll be breaking the API anyway.

barakman: Ah wait, this is the LP contract; ignore...

solidity/contracts/helpers/TestConverterRegistry.sol
barakman: But the constructor is not empty, it needs to call the parent's constructor.

lbeder: Empty as in `constructor(IContractRegistry _registry) public ConverterRegistry(_registry) {}`

solidity/contracts/converter/types/standard-pool/StandardPoolConverter.sol
```diff
@@ -0,0 +1,1386 @@
...
+    uint256 public averageRateInfo;
+
+    /**
+     * @dev triggered when the converter is activated
+     *
+     * @param _type        converter type
+     * @param _anchor      converter anchor
+     * @param _activated   true if the converter was activated, false if it was deactivated
+     */
+    event Activation(uint16 indexed _type, IConverterAnchor indexed _anchor, bool indexed _activated);
```
lbeder, line 58: Since you don't inherit from `ConverterBase`, you can still move shared definitions to `IConverter.sol` (for example)

lbeder: Like many of these events

solidity/contracts/converter/types/standard-pool/StandardPoolConverter.sol
```diff
@@ -0,0 +1,1386 @@
...
+        IConverterAnchor _anchor,
+        IContractRegistry _registry,
+        uint32 _maxConversionFee
+    ) public ContractRegistryClient(_registry) validAddress(address(_anchor)) validConversionFee(_maxConversionFee) {
+        anchor = _anchor;
+        maxConversionFee = _maxConversionFee;
+    }
+
+    // ensures that the converter is active
+    modifier active() {
```
lbeder, line 149: You can also consider creating another base class for this, thus preventing the duplications (but it's not a priority, of course)

solidity/contracts/converter/types/standard-pool/StandardPoolConverter.sol
```diff
@@ -0,0 +1,1386 @@
...
+        return reserveBalance(reserveId);
+    }
+
+    /**
+     * @dev returns the balances of both reserve tokens
+     *
+     * @return the balances of both reserve tokens
+     */
+    function reserveBalances() public view returns (uint256, uint256) {
+        return reserveBalances(1, 2);
```
lbeder, line 413: Maybe 0 and 1 instead?

solidity/contracts/converter/types/standard-pool/StandardPoolConverter.sol
```diff
@@ -0,0 +1,1386 @@
...
+
+    /**
+     * @dev stores the stored reserve balance for a given reserve id
+     *
+     * @param _reserveId       reserve id
+     * @param _reserveBalance  reserve balance
+     */
+    function setReserveBalance(uint256 _reserveId, uint256 _reserveBalance) internal {
+        require(_reserveBalance <= MAX_UINT128, "ERR_RESERVE_BALANCE_OVERFLOW");
+        uint256 otherBalance = decodeReserveBalance(__reserveBalances, 3 - _reserveId);
```
lbeder, line 480: re: `3 - ...` - can you refactor all these reserves IDs magic numbers (1, 2, and 3) to some helper functions? Otherwise, it's hard to read and error prone

solidity/contracts/converter/ConverterUpgrader.sol
```diff
@@ -165,10 +171,6 @@ contract ConverterUpgrader is IConverterUpgrader, ContractRegistryClient {
             if (reserveAddress == ETH_RESERVE_ADDRESS) {
                 _newConverter.addReserve(ETH_RESERVE_ADDRESS, weight);
             }
-            // Ether reserve token
-            else if (reserveAddress == etherToken) {
```
lbeder, line 169: Do we need to have `EtherToken` as part of the repo?

solidity/contracts/converter/types/standard-pool/StandardPoolConverter.sol
```diff
@@ -0,0 +1,1386 @@
...
+        address _trader,
+        uint256 _amount,
+        uint256 _returnAmount,
+        uint256 _feeAmount
+    ) internal {
+        // fee amount is converted to 255 bits -
+        // negative amount means the fee is taken from the source token, positive amount means its taken from the target token
+        // currently the fee is always taken from the target token
+        // since we convert it to a signed number, we first ensure that it's capped at 255 bits to prevent overflow
+        assert(_feeAmount < 2**255);
```
lbeder, line 568: This should be something like 2**127 now instead, right?

solidity/contracts/converter/types/standard-pool/StandardPoolConverter.sol
```diff
@@ -0,0 +1,1386 @@
...
+     * note that prior to version 28, you should use 'fund' instead
+     *
+     * @param _reserveTokens   address of each reserve token
+     * @param _reserveAmounts  amount of each reserve token
+     * @param _minReturn       token minimum return-amount
+     *
+     * @return amount of pool tokens issued
+     */
+    function addLiquidity(
+        IERC20Token[2] memory _reserveTokens,
```
lbeder, line 748: I think that declaring them as `calldata` will result in slightly cheaper gas costs (since `calldata` args are immutable)

solidity/contracts/liquidity-protection/LiquidityProtection.sol
barakman: It is not, because `IERC20Token[2]` and `IERC20Token[]` are two different types with no common "ancestor".
Same goes for `uint256[2]` and `uint256[]`.
If the language supported templates, then we could unify these two blocks.
Otherwise, I don't see how this can be done.

solidity/contracts/liquidity-protection/LiquidityProtection.sol
barakman: It is not, because `IERC20Token[2]` and `IERC20Token[]` are two different types with no common "ancestor".
Same goes for `uint256[2]` and `uint256[]`.
If the language supported templates, then we could unify these two blocks.
Otherwise, I don't see how this can be done.

solidity/contracts/converter/types/standard-pool/StandardPoolConverter.sol
barakman: @yudilevi : Move all events from `ConverterBase` to `IConverter`?

solidity/contracts/converter/types/standard-pool/StandardPoolConverter.sol
barakman: With `mapping(IERC20Token => uint256) private __reserveIds`, we are assuming that our two reserve tokens are mapped to 1 and 2, and all other possible addresses are mapped to 0.

solidity/contracts/converter/types/standard-pool/StandardPoolConverter.sol
barakman: Yes. 3 is their sum of course, but you're right.

solidity/contracts/converter/ConverterUpgrader.sol
barakman: I have asked the same question.
It is still being used at the `BancorNetwork` level.
And generally speaking - this change is unrelated of the PR (as you may have realized by now).

solidity/contracts/converter/types/standard-pool/StandardPoolConverter.sol
barakman: Hmmmm, probably, yes. Not sure in what way this "thing" is still relevant to us.

solidity/contracts/helpers/TestConverterRegistry.sol
barakman: Fixed.

solidity/contracts/converter/types/standard-pool/StandardPoolConverter.sol
barakman: We later use `(_id - 1) * 128` which relies on the fact that this ID can be either `1` or `2`.
Any suggestion on how to generalize this?
Just declaring constants `1` and `2` implies that one can change them to any other values.

lbeder: I see. Let's leave it for now

solidity/contracts/converter/types/standard-pool/StandardPoolConverter.sol
yudilevi: @barakman sure

solidity/contracts/converter/ConverterUpgrader.sol
yudilevi: let's open a separate task for this

solidity/contracts/converter/types/standard-pool/StandardPoolConverter.sol
yudilevi: that check isn't relevant in this pool.

solidity/contracts/converter/ConverterUpgrader.sol
barakman: Move this fix out of this branch?
Or just fix the `BancorNetwork` level in a new branch?

yudilevi: leave the branch as is and open another task for it (not to be fixed now)

solidity/contracts/helpers/TestStandardPoolConverter.sol
barakman: Done

solidity/contracts/converter/types/standard-pool/StandardPoolConverter.sol
barakman: Done

solidity/contracts/converter/types/standard-pool/StandardPoolConverter.sol
barakman: Will be done separately, as part of a "clean" task.

solidity/contracts/converter/types/standard-pool/StandardPoolConverter.sol
barakman: Fixed.

solidity/contracts/converter/types/standard-pool/StandardPoolConverter.sol
barakman: I don't understand.
First, they are not immutable at this point, because function `VerifyInput` inverts the order if necessary.
So at most, we can apply it in each one of the underlying (private) functions.
Second - this is a broad change, which should be applied globally, so why only on this function?

solidity/contracts/liquidity-protection/LiquidityProtection.sol
```diff
@@ -1006,23 +1018,35 @@ contract LiquidityProtection is TokenHandler, Utils, Owned, ReentrancyGuard, Tim
...
-        reserveTokens[0] = _reserveToken1;
-        reserveTokens[1] = _reserveToken2;
-        amounts[0] = _reserveAmount1;
-        amounts[1] = _reserveAmount2;
-
         // ensure that the contract can receive ETH
         updatingLiquidity = true;
-        _converter.addLiquidity{ value: _value }(reserveTokens, amounts, 1);
+
+        if (isStandardPool(_converter)) {
```
yudilevi, line 1031: this needs to be removed, right?

barakman: @yudilevi : Yes, this is part of the task on Trello.

yudilevi (approved):

yudilevi (approved):

PR #470, opened 2020-11-13
Standard Pool Extension

barakman:


drumenbassik (approved):

drumenbassik (approved):

solidity/contracts/converter/types/standard-pool/StandardPoolConverter.sol
```diff
@@ -795,9 +795,9 @@ contract StandardPoolConverter is
      */
     function removeLiquidity(
         uint256 _amount,
-        IERC20Token[] memory _reserveTokens,
-        uint256[] memory _reserveMinReturnAmounts
-    ) public protected active returns (uint256[] memory) {
+        IERC20Token[2] memory _reserveTokens,
+        uint256[2] memory _reserveMinReturnAmounts
+    ) public protected active returns (uint256[2] memory) {
```
barakman, line 800: This function may change the order of the input reserve tokens, thus may return the output reserve amounts in the opposite order; must be fixed.

barakman: Fixed.

PR #473, opened 2020-11-15
Configure ganache-core to avoid disk I/O

barakman:


lbeder (approved):
👍 👍 👍

truffle-config.js
```diff
@@ -24,6 +26,7 @@ module.exports = {
             gasPrice: 20000000000,
             gas: 9500000,
             provider: ganache.provider({
+                db: memdown(),
```
lbeder, line 29: You should also add this to `.solcover.js`

barakman: Done.

PR #475, opened 2020-11-15
High tier pools

yudilevi:


solidity/contracts/liquidity-protection/LiquidityProtection.sol
```diff
@@ -61,21 +61,30 @@ contract LiquidityProtection is TokenHandler, ContractRegistryClient, Reentrancy
...
+    }
+
     IERC20Token internal constant ETH_RESERVE_ADDRESS = IERC20Token(0xEeeeeEeeeEeEeeEeEeEeeEEEeeeeEeeeeeeeEEeE);
     uint32 internal constant PPM_RESOLUTION = 1000000;
     uint256 internal constant MAX_UINT128 = 2**128 - 1;
 
     // the address of the whitelist administrator
     address public whitelistAdmin;
 
+    // list of pools with less minting restrictions and mapping of pool anchor address -> index in the list of pools for quick access
```
barakman, line 76: Split into two separate comments:
1. list of pools with fewer minting restrictions
2. mapping of each pool anchor address to its index in the list of pools (for quick access)

solidity/contracts/liquidity-protection/LiquidityProtection.sol
```diff
@@ -369,6 +378,77 @@ contract LiquidityProtection is TokenHandler, ContractRegistryClient, Reentrancy
         else store.removePoolFromWhitelist(_poolAnchor);
     }
 
+    /**
+     * @dev adds a high tier pool
+     * can only be called by the contract owner
```
barakman, line 383: IMO `can be called only by the contract owner`.

solidity/contracts/liquidity-protection/LiquidityProtection.sol
```diff
@@ -61,21 +61,30 @@ contract LiquidityProtection is TokenHandler, ContractRegistryClient, Reentrancy
...
     IERC20Token internal constant ETH_RESERVE_ADDRESS = IERC20Token(0xEeeeeEeeeEeEeeEeEeEeeEEEeeeeEeeeeeeeEEeE);
     uint32 internal constant PPM_RESOLUTION = 1000000;
     uint256 internal constant MAX_UINT128 = 2**128 - 1;
 
     // the address of the whitelist administrator
     address public whitelistAdmin;
 
+    // list of pools with less minting restrictions and mapping of pool anchor address -> index in the list of pools for quick access
+    IConverterAnchor[] private _highTierPools;
+    mapping(IConverterAnchor => PoolIndex) private highTierPoolIndices;
```
lbeder, line 78: Should this state variable also start with an underscore (like `_highTierPools`)?

solidity/contracts/liquidity-protection/LiquidityProtection.sol
```diff
@@ -369,6 +378,77 @@ contract LiquidityProtection is TokenHandler, ContractRegistryClient, Reentrancy
         else store.removePoolFromWhitelist(_poolAnchor);
     }
 
+    /**
+     * @dev adds a high tier pool
+     * can only be called by the contract owner
+     *
+     * @param _poolAnchor pool anchor
+     */
+    function addHighTierPool(IConverterAnchor _poolAnchor)
```
lbeder, line 387: Perhaps it's worth to add add/remove events for this (or since it's owner-only, it's not important)?

solidity/contracts/liquidity-protection/LiquidityProtection.sol
```diff
@@ -369,6 +378,77 @@ contract LiquidityProtection is TokenHandler, ContractRegistryClient, Reentrancy
...
+     */
+    function removeHighTierPool(IConverterAnchor _poolAnchor)
+        external
+        ownerOnly
+        validAddress(address(_poolAnchor))
+        notThis(address(_poolAnchor))
+    {
+        // validate input
+        PoolIndex storage poolIndex = highTierPoolIndices[_poolAnchor];
+        require(poolIndex.isValid, "ERR_POOL_DOESNT_EXIST");
```
barakman, line 416: `DOES_NOT`

solidity/contracts/liquidity-protection/LiquidityProtection.sol
```diff
@@ -369,6 +378,77 @@ contract LiquidityProtection is TokenHandler, ContractRegistryClient, Reentrancy
...
+    /**
+     * @dev checks whether a given pool is a high tier one
+     *
+     * @param _poolAnchor pool anchor
+     * @return true if the given pool is a high tier one, false otherwise
+     */
+    function isHighTierPool(IConverterAnchor _poolAnchor) public view returns (bool) {
+        return highTierPoolIndices[_poolAnchor].isValid;
+    }
+
```
barakman, line 451: Missing:
```
    function highTierPoolCount() external view returns (uint256) {
        return _highTierPools;
    }
```

lbeder (approved):

barakman (changes requested):

solidity/contracts/liquidity-protection/LiquidityProtection.sol
yudilevi: I just did it cause we have such a function name

solidity/contracts/liquidity-protection/LiquidityProtection.sol
yudilevi: yeah I thought it's not important for now since it's a temporary solution - once we have the full tier system, we'll have events for all of these (we need to add these pools to the high tier list with every upgrade for now)

solidity/contracts/liquidity-protection/LiquidityProtection.sol
yudilevi: done

solidity/contracts/liquidity-protection/LiquidityProtection.sol
yudilevi: copied from any other such comment we have

solidity/contracts/liquidity-protection/LiquidityProtection.sol
yudilevi: done

solidity/contracts/liquidity-protection/LiquidityProtection.sol
yudilevi: yeah didn't want it there - you can get the array and check its length

solidity/contracts/liquidity-protection/LiquidityProtection.sol
lbeder: Gotcha

solidity/contracts/liquidity-protection/LiquidityProtection.sol
lbeder: 👍

solidity/contracts/liquidity-protection/LiquidityProtection.sol
barakman: Quoting you: "the fact that we have some 'bad' stuff in previous code doesn't mean that we should port it to new code".

solidity/contracts/liquidity-protection/LiquidityProtection.sol
barakman: Makes me wonder if we even need all of those `Count` methods that we've added up until now in various places.
Perhaps it becomes useful once the array is extremely large, and a node might reject your attempt to read it in full.
@lbeder : maybe you have an idea?

lbeder: `*Count` methods are indeed very useful when the array is very large (not because it rejects, but to save undeeded resources using web3.js)

solidity/contracts/liquidity-protection/LiquidityProtection.sol
yudilevi: and it's true, but this is a comment

solidity/contracts/liquidity-protection/LiquidityProtection.sol
yudilevi: yeah it's not relevant in this case, temp solution for very small number of pools.

PR #477, opened 2020-11-17
Skip average-rate validation in read-only functions

barakman:
In the read-only functions `poolROI` and `removeLiquidityReturn`, avoid reverting when the maximum permitted deviation of the average rate from the spot rate is exceeded.

### DIFF
--- a/solidity/contracts/liquidity-protection/LiquidityProtection.sol
+++ b/solidity/contracts/liquidity-protection/LiquidityProtection.sol
@@ -1117,7 +1117,7 @@ contract LiquidityProtection is TokenHandler, ContractRegistryClient, Reentrancy
         uint256 _poolAmount,
         uint256 _reserveAmount
     ) internal returns (uint256) {
-        Fraction memory rate = reserveTokenAverageRate(_poolToken, _reserveToken);
+        Fraction memory rate = reserveTokenAverageRate(_poolToken, _reserveToken, true);
         return
             store.addProtectedLiquidity(
                 _provider,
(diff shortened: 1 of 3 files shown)


solidity/contracts/liquidity-protection/LiquidityProtection.sol
```diff
@@ -1258,9 +1270,7 @@ contract LiquidityProtection is TokenHandler, ContractRegistryClient, Reentrancy
         uint256 _averageRateD,
         uint32 _maxDeviation
     ) internal pure returns (bool) {
-        uint256 min = _spotRateN.mul(_averageRateD).mul(PPM_RESOLUTION - _maxDeviation).mul(
-            PPM_RESOLUTION - _maxDeviation
-        );
+        uint256 min = _spotRateN.mul(_averageRateD).mul(PPM_RESOLUTION - _maxDeviation).mul(PPM_RESOLUTION - _maxDeviation);
```
lbeder, line 1273: This is a prettier change, so it'll revert later when someone uses formating.

barakman: Yup, so it turns out that the prettier doesn't always actually prettifies things.
IMO, the previous (red) form makes it harder to read and comprehend the function, which is somewhat arithmetically complicated. I believe that it is better in the new (green) form.
Is there anything that we can do in order to tell it to skip this one?

yudilevi: @barakman we'll handle prettier rules when we have some time to breath, it's a simple change that will auto-format so no need to worry about it

solidity/contracts/liquidity-protection/LiquidityProtection.sol
```diff
@@ -1163,17 +1171,18 @@ contract LiquidityProtection is TokenHandler, ContractRegistryClient, Reentrancy
         view
         returns (Fraction memory)
     {
-        (, , uint256 averageRateN, uint256 averageRateD) = reserveTokenRates(_poolToken, _reserveToken);
+        (, , uint256 averageRateN, uint256 averageRateD) = reserveTokenRates(_poolToken, _reserveToken, true);
```
yudilevi, line 1174: a bit confusing that the function always passes `true` here - we can't really guarantee that it's only used in write functions (as opposed to read-only). think we should do anything about it?

barakman: Depending on the purpose of the function.
For this specific function, you are probably right, i.e., it should take that flag as input.

yudilevi (approved):

PR #478, opened 2020-11-18
Token Time-Weighted Average

lbeder:


solidity/contracts/converter/types/liquidity-pool-v1/LiquidityPoolV1Converter.sol
```diff
@@ -697,6 +670,33 @@ contract LiquidityPoolV1Converter is LiquidityPoolConverter {
         return reserveAmounts;
     }
 
+    /**
```
barakman, line 673: Is this just the result of moving a function from one place in the file to another?

solidity/contracts/helpers/TestTokenTimeWeightedAverage.sol
```diff
@@ -0,0 +1,18 @@
+// SPDX-License-Identifier: SEE LICENSE IN LICENSE
+pragma solidity 0.6.12;
+
+import "../liquidity-protection/TokenTimeWeightedAverage.sol";
+
+contract TestTokenTimeWeightedAverage is TokenTimeWeightedAverage {
+    uint256 public currentTime = 1;
```
barakman, line 7: We already have this infrastructure in about 3 or 4 different places; might wanna consider reusing it somehow.

solidity/contracts/liquidity-protection/TokenTimeWeightedAverage.sol
```diff
@@ -0,0 +1,312 @@
...
+        uint256 lastSampleTime;
+        uint256 prevAccumulatorTime;
+        Fraction prevAccumulator;
+        mapping(uint256 => Fraction) accumulators;
+        mapping(uint256 => bool) timestamps;
+    }
+
+    uint256 private constant INITIAL_SAMPLE_N = 0;
+    uint256 private constant INITIAL_SAMPLE_D = 1;
+    uint256 private constant MAX_UINT128 = 2**128 - 1;
```
barakman, line 30: Missing spaces before and after `**` (probably my fault in the original commit).

solidity/test/TokenTimeWeightedAverage.js
```diff
@@ -0,0 +1,373 @@
...
+                }),
+                (time) => time.add(duration.days(1))
+            );
+        });
+
+        it('should properly accumulate infinitesimal values', async () => {
+            await test(
+                (value) => ({
+                    n: new BN(1),
+                    d: value.d.add(new BN(10 ** 10).mul(new BN(10).pow(new BN(18))))
```
barakman, line 367: Perhaps just `value.d.add(new BN(10).pow(new BN(28))))`?

solidity/contracts/helpers/TestTokenTimeWeightedAverage.sol
lbeder: Perhaps. I'll add it to my todos

solidity/contracts/liquidity-protection/TokenTimeWeightedAverage.sol
lbeder: Prettier... nothing to do about that, other than to chillax.

solidity/test/TokenTimeWeightedAverage.js
lbeder: I wanted to be very explicit about the meaning of the numbers

solidity/contracts/helpers/TestTokenTimeWeightedAverage.sol
lbeder: Done

solidity/contracts/helpers/TestLiquidityPoolV1Converter.sol
```diff
@@ -12,19 +13,15 @@ contract TestLiquidityPoolV1Converter is LiquidityPoolV1Converter {
...
-
     function removeLiquidityTest(
         uint256 _amount,
         IERC20Token[] memory _reserveTokens,
         uint256[] memory _reserveMinReturnAmounts
     ) public {
         reserveAmountsRemoved = removeLiquidity(_amount, _reserveTokens, _reserveMinReturnAmounts);
     }
+
+    function time() public view virtual override(Time, TestTime) returns (uint256) {
```
barakman, line 24: I wonder if there's an OO pattern which would allow us to get rid of this function in this contract.
After all, it is already declared in contract `TestTime`, which this contract inherits from.
I guess that there might be a diamond issue here, because this contract also inherits from contract `Time` which also implements this function...

solidity/contracts/liquidity-protection/TokenTimeWeightedAverage.sol
barakman: +1 for chillax

solidity/contracts/liquidity-protection/TokenTimeWeightedAverage.sol
```diff
@@ -0,0 +1,375 @@
...
+     * @param _time the sampling timestamp
+     */
+    function addPastSample(
+        IERC20Token _token,
+        uint256 _n,
+        uint256 _d,
+        uint256 _time
+    ) external override validAddress(address(_token)) greaterThanZero(_d) initialized(_token) {
+        require(hasRole(ROLE_SEEDER, msg.sender), "ERR_ACCESS_DENIED");
+        require(_time <= time(), "ERR_INVALID_TIME");
```
barakman, line 131: Minor: If `_time == time()` then it's not a "past sample" (and also - the user can/should call function `addSample` in this case).
But in either case, see my comment above about renaming this function to `addSample` and calling it from the other `addSample`.

solidity/contracts/liquidity-protection/TokenTimeWeightedAverage.sol
```diff
@@ -0,0 +1,375 @@
...
+        uint256 length = _ns.length;
+        require(length == _ds.length && length == _times.length, "ERR_INVALID_LENGTH");
+
+        for (uint256 i = 0; i < length; ++i) {
+            uint256 n = _ns[i];
+            uint256 d = _ds[i];
+            uint256 t = _times[i];
+
+            _greaterThanZero(d);
+            require(t <= time(), "ERR_INVALID_TIME");
```
barakman, line 162: Again, `time()` is a present-sample, so you should use `<` instead of `<=` here.

And you may as well just call `addPastSample` instead of `addSample` here (or if you accept my comments above, then call the `addSample` function which takes `_time` as an additional input argument).

solidity/contracts/liquidity-protection/TokenTimeWeightedAverage.sol
```diff
@@ -0,0 +1,375 @@
...
+
+    /**
+     * @dev calculated the TWA from a specific sample time
+     *
+     * @param _token the token the data is accumulated for
+     * @param _startTime the sampling starting time
+     *
+     * @return TWA's numerator and denominator
+     */
+    function timeWeightedAverage(IERC20Token _token, uint256 _startTime)
```
barakman, line 176: Code reuse: you can simply call the other `timeWeightedAverage`, passing `tokenData.lastSampleTime` as `_endTime`.

solidity/contracts/liquidity-protection/TokenTimeWeightedAverage.sol
```diff
@@ -0,0 +1,375 @@
...
+
+    /**
+     * @dev adds a new sample to the accumulator
+     * can only be called by an owner
+     *
+     * @param _token the token the data is accumulated for
+     * @param _n ratio numerator
+     * @param _d ratio denominator
+     */
+    function addSample(
```
barakman, line 105: Code reuse: you can simply call `addPastSample(_n, _d, time())` here.
I would rename `addPastSample` to `addSample` (i.e., function-overload with this one), because `time()` means that we're adding a present-sample, not a past-sample (I apologize if it sounds like an English grammar lesson).

solidity/contracts/liquidity-protection/TokenTimeWeightedAverage.sol
```diff
@@ -0,0 +1,375 @@
...
+    function timeWeightedAverage(IERC20Token _token, uint256 _startTime)
+        external
+        view
+        override
+        validAddress(address(_token))
+        returns (uint256, uint256)
+    {
+        TokenData storage tokenData = data[_token];
+        require(tokenData.timestamps[_startTime], "ERR_NO_DATA");
+
```
barakman, line 185: No need for `require(_startTime < tokenData.lastSampleTime, "ERR_INVALID_TIME");` here?
If you accept my comment above, then you get this additional statement "for free".

barakman (changes requested):

solidity/contracts/helpers/TestLiquidityPoolV1Converter.sol
lbeder: Yeah. I couldn't find an easier solution :(

solidity/contracts/liquidity-protection/TokenTimeWeightedAverage.sol
lbeder: I'll fix it to a `<`

solidity/contracts/liquidity-protection/TokenTimeWeightedAverage.sol
lbeder: In both of these cases, I wanted to avoid triggering the input validating modifiers twice, so both reuse the same `private` implementation

solidity/contracts/liquidity-protection/TokenTimeWeightedAverage.sol
lbeder: In both of these cases, I wanted to avoid triggering the input validating modifiers twice, so both reuse the same `private` implementation

solidity/contracts/liquidity-protection/TokenTimeWeightedAverage.sol
lbeder: In both of these cases, I wanted to avoid triggering the input validating modifiers twice, so both reuse the same `private` implementation

solidity/contracts/liquidity-protection/TokenTimeWeightedAverage.sol
lbeder: Similarly to the above, it's the same "keep all checks outside of the shared private method" approach

solidity/contracts/liquidity-protection/TokenTimeWeightedAverage.sol
lbeder: Fixed

solidity/contracts/liquidity-protection/TokenTimeWeightedAverage.sol
```diff
@@ -0,0 +1,375 @@
...
+        Fraction memory endAccumulator = _tokenData.accumulators[_endTime];
+
+        // if we have received only a single sample - just return it
+        if (_tokenData.firstSampleTime == _endTime) {
+            return (endAccumulator.n, endAccumulator.d);
+        }
+
+        Fraction memory startAccumulator = _tokenData.accumulators[_startTime];
+
+        // TWA = (endAccumulator - startAccumulator) / (_endTime - _startTime)
```
barakman, line 234: Better yet:
```
uint256 n = startAccumulator.d.mul(endAccumulator.n).sub(startAccumulator.n.mul(endAccumulator.d));
uint256 d = startAccumulator.d.mul(endAccumulator.d).mul(_endTime.sub(_startTime));
```
No need to lose precision on division here.

solidity/contracts/liquidity-protection/TokenTimeWeightedAverage.sol
```diff
@@ -0,0 +1,375 @@
...
+            return;
+        }
+
+        Fraction memory lastAccumulator = tokenData.accumulators[lastSampleTime];
+
+        // update the previous accumulator value once per-block. we would need it in order to accumulate
+        // same-block changes
+        if (_time > tokenData.prevAccumulatorUpdateTime) {
+            tokenData.prevAccumulatorUpdateTime = _time;
+            tokenData.prevAccumulator = lastAccumulator;
```
barakman, line 337: I always prefer:
```
if (tokenData.prevAccumulatorUpdateTime < _time) {
    tokenData.prevAccumulatorUpdateTime = _time;
    ...
}
```
One can see the logic more clearly IMO.

solidity/contracts/liquidity-protection/TokenTimeWeightedAverage.sol
```diff
@@ -0,0 +1,375 @@
...
+     * @param _token the token the data is accumulated for
+     *
+     * @return whether the accumulator is initialized for a specific token
+     */
+    function isInitialized(IERC20Token _token) private view returns (bool) {
+        TokenData storage tokenData = data[_token];
+        uint256 firstSampleTime = tokenData.firstSampleTime;
+
+        Fraction memory firstAccumulator = tokenData.accumulators[firstSampleTime];
+        return firstSampleTime > 0 && firstAccumulator.n == INITIAL_SAMPLE_N && firstAccumulator.d == INITIAL_SAMPLE_D;
```
barakman, line 373: Is it possible for `firstAccumulator.n` and `firstAccumulator.d` to be anything other than 0 when `firstSampleTime` is 0?
Otherwise, it might be useful to avoid the related storage read in this case.

barakman (changes requested):

solidity/contracts/liquidity-protection/TokenTimeWeightedAverage.sol
```diff
@@ -0,0 +1,375 @@
...
+        }
+
+        // if we already have a sample for this timestamp - use the backup of the previous accumulator and use it in
+        // combination with the new sample
+        if (_time == lastSampleTime) {
+            lastAccumulator = tokenData.prevAccumulator;
+        }
+
+        // accumulate the current value in combination with the last accumulator using the TWA formuls:
+        //  ACC[time] = ACC[time - timeDiff] + value * timeDiff
```
barakman, line 347: I think that you should change this comment to:
```
// accumulator = lastAccumulator + smaple * timeDelta
```
If you don't think so, then just note the extra space after the slashes.

solidity/contracts/utility/Time.sol
```diff
@@ -0,0 +1,16 @@
...
+import "./interfaces/ITime.sol";
+
+/*
+    Time implementing contract
+*/
+contract Time is ITime {
+    /**
+     * @dev returns the current time
+     */
+    function time() public view virtual override returns (uint256) {
```
barakman, line 13: Shouldn't this function be `internal`?

solidity/contracts/utility/interfaces/ITime.sol
```diff
@@ -0,0 +1,12 @@
...
+pragma solidity 0.6.12;
+
+/*
+    Time implementing contract interface
+*/
+interface ITime {
+    /**
+     * @dev returns the current time
+     */
+    function time() external view returns (uint256);
```
barakman, line 11: Ahhh, it can't be `internal`.
But why do we need an `interface` here?
Contract `Time` should be an inherited extension (kinda like `RegistryClient`), no?

solidity/test/TokenTimeWeightedAverage.js
```diff
@@ -0,0 +1,502 @@
...
+        accumulators: {
+            [start]: Decimal(0)
+        },
+        lastAccumulator: Decimal(0),
+        lastAccumulatorTime: Decimal(0)
+    };
+};
+
+const addTWASample = (acc, n, d, time) => {
+    const sampleTime = time.toNumber();
```
barakman, line 28: Not sure what the type of `time` is in this function, and why you chose to convert it to `Number`.
But since you're converting it to `Decimal` below, it seems pointless to convert it to `Number` here.
For example, what if it represents an integer value larger than `Number.MAX_SAFE_INTEGER`?
If the issue is that `time` is a `BN` object, then the typical solution is:
```
const sampleTime = Decimal(time.toString());
```

solidity/test/TokenTimeWeightedAverage.js
```diff
@@ -0,0 +1,502 @@
...
+
+describe('TokenTimeWeightedAverage', () => {
+    const owner = defaultSender;
+    const seeder = accounts[1];
+    const nonOwner = accounts[5];
+    let twa;
+    let now;
+
+    beforeEach(async () => {
+        now = await latest();
```
barakman, line 63: Since we're using `TestTokenTimeWeightedAverage`, which inherits from `TestTime`, my recommendation is to avoid using the actual blockchain time throughout the tests, as it just adds an undetermined factor into them (i.e., the possibility of different behavior on different executions).
You can just use any value larger than 0, or rely on the fact that `TestTime` already does this for you (that's why I've explicitly used `uint256 public currentTime = 1`), which will relieve you of the need to call `twa.setTime(now)`.

barakman (changes requested):

solidity/contracts/liquidity-protection/TokenTimeWeightedAverage.sol
lbeder: 👍

solidity/contracts/liquidity-protection/TokenTimeWeightedAverage.sol
lbeder: 👍

solidity/contracts/liquidity-protection/TokenTimeWeightedAverage.sol
lbeder: Nope, they shouldn't. I'll remove it

solidity/contracts/liquidity-protection/TokenTimeWeightedAverage.sol
lbeder: Sure, why not?

solidity/contracts/utility/interfaces/ITime.sol
lbeder: I'll wait for your PR

solidity/contracts/liquidity-protection/TokenTimeWeightedAverage.sol
barakman: Only reason not to is because I'm wrong.

solidity/test/TokenTimeWeightedAverage.js
lbeder: 👍

solidity/test/TokenTimeWeightedAverage.js
lbeder: My original intention was to mix "real" timestamps to the calculation, but since it's using a delta anyway - it doesn't matter. I'll fix it

solidity/contracts/liquidity-protection/TokenTimeWeightedAverage.sol
barakman: As you can understand, I only moved the "div" part from the numerator to be a "mul" part in the denominator.
You may notice a few other changes in each one of them, but those are only cosmetic (OCD-related) reorders of the multiplication operands.

solidity/contracts/utility/interfaces/ITime.sol
barakman: Done

PR #479, opened 2020-11-19
Use test-env for tests and fix legacy tests

lbeder:


chai-local.js
```diff
@@ -0,0 +1,14 @@
+/* eslint-disable import/no-extraneous-dependencies */
+const chai = require('chai');
+chai.use(require('chai-as-promised'))
+    .use(require('chai-bn')(require('bn.js')))
+    .use(require('chai-string'))
+    .use(require('dirty-chai'))
+    .expect();
+
+const Decimal = require('decimal.js');
+Decimal.set({ precision: 100, rounding: Decimal.ROUND_DOWN, toExpPos: 40 });
```
barakman, line 10: I don't think we need to `toExpPos: 40` part (can simply use `toFixed(40)` where necessary, if necessary).
I know it was there before, but wondering why you added it in the first place...

lbeder: We did. It was a global setting in truffle-config.js

yudilevi (approved):

PR #481, opened 2020-11-20
Update prettier and remove the explicit ganache-core dependency 

lbeder:


solidity/contracts/BancorNetwork.sol
```diff
@@ -1,781 +1,797 @@
...
-        address _affiliateAccount,
-        uint256 _affiliateFee
-    )
-        public
-        returns (uint256)
-    {
-        return convertByPath(_path, _amount, _minReturn, _beneficiary, _affiliateAccount, _affiliateFee);
-    }
-}
+// SPDX-License-Identifier: SEE LICENSE IN LICENSE
```
yudilevi, line 1: is this only a prettier change?

lbeder: Yeah. Apparently, the old version couldn't parse this file properly, so didn't format it

yudilevi (approved):

PR #482, opened 2020-11-20
Migrate LP settings to an external contract

lbeder:


solidity/contracts/liquidity-protection/LiquidityProtectionSettings.sol
```diff
@@ -0,0 +1,443 @@
...
+    /**
+     * @dev adds a pool to the whitelist
+     * can only be called by the contract owner
+     *
+     * @param _poolAnchor pool anchor
+     */
+    function addPoolToWhitelist(IConverterAnchor _poolAnchor)
+        external
+        override
+        onlyWhitelistAdmin
```
yudilevi, line 184: that's a logical change - was previously whitelist admin or owner - intentional?

lbeder: You're right. Yes, I wanted to be less verbose, since the owner can always set himself as a whitelist admin

yudilevi: ok, cool

yudilevi (approved):

PR #490, opened 2020-11-24
Ensure at least one block between creating a position and removing it

barakman:
Ensure at least 1 block between creating a position and removing it (or a part of it).


yudilevi (changes requested):
no tests added

lbeder (approved):

lbeder (changes requested):
Waiting for tests as well

PR #492, opened 2020-11-24
Patch OZ Test Environment

barakman:
OZ test environment currently does not support configuring ganache to avoid disk I/O, which significantly increases test execution time.

This commit adds a "hot-fix" of OZ source code (editing of an internal source file at the end of `npm install`), which forces the above configuration.

It is of course not guaranteed to be backward-compatible and work correctly on the next version of the OZ Test Environment, but hopefully it will remain so (backward-compatible) until they fix this issue (i.e., before they change the file being patched here).

We may require that NodeJS will be configured with `--max-old-space-size=4096` when `npm run qtest` is executed (i.e., we may need to update the corresponding entry in our `package.json`).

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,7 +16,7 @@
         "npm": "^3.0.0"
     },
     "scripts": {
-        "install": "node scripts/fix-modules.js",
+        "postinstall": "node scripts/fix-modules.js",
         "build": "truffle compile",
         "test": "npm run build && npm run qtest",
         "qtest": "mocha",
--- a/package.json
+++ b/package.json
@@ -16,6 +16,7 @@
         "npm": "^3.0.0"
     },
     "scripts": {
+        "install": "node scripts/fix-modules.js",
         "build": "truffle compile",
         "test": "npm run build && npm run qtest",
         "qtest": "mocha",
--- a/scripts/fix-modules.js
+++ b/scripts/fix-modules.js
@@ -1,18 +1,18 @@
-const fs = require("fs");
+const fs = require('fs');
 
 function fix(fileName, tokens) {
-    console.log("Fixing " + fileName);
+    console.log('Fixing ' + fileName);
     try {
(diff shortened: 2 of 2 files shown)


package.json
```diff
@@ -16,6 +16,7 @@
         "npm": "^3.0.0"
     },
     "scripts": {
+        "install": "node scripts/fix-modules.js",
```
lbeder, line 19: It needs to be `postinstall`

scripts/fix-modules.js
```diff
@@ -0,0 +1,18 @@
+const fs = require("fs");
```
lbeder, line 1: You should prettify it. For example, we don't use double quotes in js files

lbeder (changes requested):
This is well-intended, but I'm really against patching it this way. Please consider waiting for their response or fork `test-env` and open a PR for OZ to allow this configuration.

package.json
barakman: It's the same according to the npm official documentation AFAIK.

scripts/fix-modules.js
barakman: That true for the entire repo though, so we probably wanna fix it separately.
In any case, I applied it to this file only.
First, by just replacing double quotes with single quotes, which forced me to use backslashes for a couple of single quotes **within** one of the strings.
Then, by just running the prettifier on this file.
Note that in this case, the single quotes inside the string have caused the prettifier to violate its own rules, by overriding my backslash fix inside the string, and using double quotes for the string itself.
So the file now has an equally bad violation of our coding-convention.
we might be able to workaround this with ticks or some other JS syntax.

lbeder:
All other files were already prettified

lbeder:
Yes, but it’s being deprecated

package.json
barakman: Fixed

barakman:
Fixed in `openzeppelin/test-environment` v0.1.7.

PR #494, opened 2020-11-25
Upgrade to OZ Test Environment v0.1.7

barakman:
This new version supports a broader configuration of ganache.
Following this upgrade, configure ganache to avoid disk I/O.
Note that we might need to commit `package-lock.json` too.


lbeder (changes requested):
You should also commit package-json.lock

barakman:
> You should also commit package-json.lock

Done

lbeder (approved):

PR #495, opened 2020-12-01
Remove transferLiquidity, add addLiquidityFor and checkpointing mechanism

lbeder:


solidity/test/LiquidityProtection.js
```diff
@@ -144,7 +145,7 @@ describe('LiquidityProtection', () => {
         return new BN(transaction.gasPrice).mul(new BN(txResult.receipt.cumulativeGasUsed));
     };
 
-    const expectAlmostEqual = (amount1, amount2, maxError = '0.01') => {
+    const expectAlmostEqual = (amount1, amount2, maxError = 0.01) => {
```
barakman, line 148: Using native floating-point numbers here is wrong.
We cast this string to `Decimal`.
When you cast `0.01` instead of `'0.01'`, you may get a different value.
This is also true for anywhere we call this function and pass this value explicitly (instead of leaving it as default).
Especially since we actually pass very small values in several different places.
The usage of strings for floating-point values is intentional.

solidity/test/LiquidityProtection.js
```diff
@@ -542,7 +585,7 @@ describe('LiquidityProtection', () => {
         await poolToken.approve(liquidityProtection.address, balance);
         await liquidityProtection.protectLiquidity(poolToken.address, '200');
 
-        liquidityProtection.setTime(now.add(duration.seconds(1)));
```
barakman, line 545: Holy c#$%p, I've placed several of those in the code. How did this even work???

solidity/contracts/liquidity-protection/CheckpointStore.sol
```diff
@@ -0,0 +1,115 @@
...
+     * @param _addresses the addresses we're collecting the data for
+     * @param _times the checkpoints
+     */
+    function addPastCheckpoints(address[] calldata _addresses, uint256[] calldata _times) external override {
+        require(hasRole(ROLE_SEEDER, msg.sender), "ERR_ACCESS_DENIED");
+
+        uint256 length = _addresses.length;
+        require(length == _times.length, "ERR_INVALID_LENGTH");
+
+        for (uint256 i = 0; i < length; ++i) {
```
barakman, line 79: Minor: we usually do `i++` (x5 different places in this PR).

barakman (changes requested):

solidity/contracts/liquidity-protection/interfaces/ICheckpointStore.sol
```diff
@@ -0,0 +1,15 @@
+// SPDX-License-Identifier: SEE LICENSE IN LICENSE
+pragma solidity 0.6.12;
+
+/**
+ * @dev Checkpoint store contract interface
+ */
+interface ICheckpointStore {
+    function addCheckpoint(address _address) external;
```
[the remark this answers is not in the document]
lbeder: The rationale is that batch update is only required (and should be used) for "seeding" past checkpoints, while new checkpoints, from this point forward, will added on time

solidity/test/CheckpointStore.js
```diff
@@ -0,0 +1,206 @@
...
+
+describe('CheckpointStore', () => {
+    const owner = defaultSender;
+    const seeder = accounts[1];
+    const nonOwner = accounts[5];
+    const user = accounts[6];
+    const user2 = accounts[7];
+
+    let checkpointStore;
+    let now = new BN(1606121626);
```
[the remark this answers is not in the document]
lbeder: Just a semi-random value. I can change it to `await latest()` or to `new BN(10000)` for all we care. What do you prefer?

solidity/test/LiquidityProtection.js
lbeder: Gotcha. I'll revert.

solidity/test/LiquidityProtection.js
lbeder: Not sure why it got removed. I'll restore it

lbeder: Ah, I've replaced them with a global `setTime` that handles other contracts as well

solidity/contracts/liquidity-protection/CheckpointStore.sol
lbeder: Old Hobbits die hard. In C/C++, `++i` is more efficient than `i++`. I'll change it

solidity/test/CheckpointStore.js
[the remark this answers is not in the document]
barakman: Not critical.
I guess that if we use `await latest()` elsewhere in the code, then we can use it here too (though it generally introduces non-deterministic-ness into our tests, so I would actually avoid using that altogether and everywhere).

barakman (changes requested):

solidity/test/LiquidityProtection.js
barakman: Yeah, you actually fixed that here (made me fix it in the other PR, so we'll probably have some merge-conflicts.

solidity/contracts/liquidity-protection/CheckpointStore.sol
barakman: Hobbies (killing hobbits is quite cruel).
BTW, `++i` is more efficient also in Solidity if I remember correctly (because just like in C/C++, there is no need to store a temporary value and return it).

solidity/test/LiquidityProtection.js
lbeder: I've merged and retested everything

solidity/contracts/liquidity-protection/CheckpointStore.sol
lbeder: Probably not a biggy

solidity/test/CheckpointStore.js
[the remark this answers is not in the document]
lbeder: Fixed anyway

barakman (approved):

solidity/test/LiquidityProtectionSettings.js
```diff
@@ -118,15 +118,31 @@ describe('LiquidityProtectionSettings', () => {
 
         it('should succeed when an admin attempts to add a whitelisted pool', async () => {
             expect(await settings.isPoolWhitelisted.call(poolToken.address)).to.be.false();
+            expect(await settings.poolWhitelist.call()).to.be.ofSize(0);
+
             await settings.addPoolToWhitelist(poolToken.address, { from: admin });
+
             expect(await settings.isPoolWhitelisted.call(poolToken.address)).to.be.true();
+            expect(await settings.poolWhitelist.call()).to.be.equalTo([poolToken.address]);
```
barakman, line 126: Shouldn't this be `deepEqualTo` or something similar?

solidity/test/LiquidityProtectionStateless.js
```diff
@@ -31,29 +34,33 @@ function assertAlmostEqual(actual, expected) {
...
 
         const govToken = await DSToken.new('vBNT', 'vBNT', 18);
         const govTokenGovernance = await TokenGovernance.new(govToken.address);
-        await govTokenGovernance.grantRole(roles.ROLE_GOVERNOR, governor);
+        await govTokenGovernance.grantRole(ROLE_GOVERNOR, governor);
         await govToken.transferOwnership(govTokenGovernance.address);
         await govTokenGovernance.acceptTokenOwnership();
 
+        const checkpointStore = await CheckpointStore.new({ from: owner });
         liquidityProtection = await LiquidityProtection.new(
```
barakman, line 53: Sender (`from`) not important here?

solidity/test/helpers/Constants.js
```diff
@@ -19,7 +19,8 @@ module.exports = {
         ROLE_OWNER: web3.utils.keccak256('ROLE_OWNER'),
         ROLE_GOVERNOR: web3.utils.keccak256('ROLE_GOVERNOR'),
         ROLE_MINTER: web3.utils.keccak256('ROLE_MINTER'),
-        ROLE_WHITELIST_ADMIN: web3.utils.keccak256('ROLE_WHITELIST_ADMIN')
+        ROLE_WHITELIST_ADMIN: web3.utils.keccak256('ROLE_WHITELIST_ADMIN'),
+        ROLE_SEEDER: web3.utils.keccak256('ROLE_SEEDER')
     },
 
     ETH_RESERVE_ADDRESS: '0xEeeeeEeeeEeEeeEeEeEeeEEEeeeeEeeeeeeeEEeE'
```
barakman, line 26: Maybe it's time we added this one in some common helper file.

barakman (changes requested):

solidity/test/LiquidityProtectionSettings.js
lbeder: `equalTo` is already a special/new method from `chai-arrays`.

solidity/test/LiquidityProtectionStateless.js
lbeder: Not really. I mean, it's a matter of style. I like to always explicitly specify who is sending a transaction when it matters/permissions, but in both of the cases, since the owner is the defaultSender anyway - it won't matter

solidity/test/helpers/Constants.js
lbeder: This is the common helper file, no?

solidity/contracts/liquidity-protection/CheckpointStore.sol
```diff
@@ -0,0 +1,115 @@
...
+    }
+
+    /**
+     * @dev returns the store value for a specific address
+     *
+     * @param _address the address we're collecting the data for
+     *
+     * @return the checkpoint
+     */
+    function checkpoint(address _address) external view override validAddress(_address) returns (uint256) {
```
yudilevi, line 97: do we actually need the address check here?

solidity/contracts/liquidity-protection/LiquidityProtection.sol
```diff
@@ -85,12 +87,14 @@ contract LiquidityProtection is TokenHandler, Utils, Owned, ReentrancyGuard, Tim
      * @param _store liquidity protection store
      * @param _networkTokenGovernance network token governance
      * @param _govTokenGovernance governance token governance
+     * @param _lastRemoveCheckpointStore last liquidity removal/transfer/unprotection checkpoints store
```
yudilevi, line 90: transfer isn't relevant here anymore

solidity/contracts/liquidity-protection/LiquidityProtection.sol
```diff
@@ -255,6 +263,34 @@ contract LiquidityProtection is TokenHandler, Utils, Owned, ReentrancyGuard, Tim
         store.withdrawTokens(liquidity1.poolToken, msg.sender, liquidity1.poolAmount.add(liquidity2.poolAmount));
     }
 
+    /**
+     * @dev adds protected liquidity to a pool for a specific recipient
+     * also mints new governance tokens for the caller if the caller adds network tokens
+     *
+     * @param _recipient       protected liquidity recipient
```
yudilevi, line 270: owner?

solidity/contracts/liquidity-protection/LiquidityProtection.sol
```diff
@@ -255,6 +263,34 @@ contract LiquidityProtection is TokenHandler, Utils, Owned, ReentrancyGuard, Tim
...
+        external
+        payable
+        protected
+        validAddress(_recipient)
+        poolSupported(_poolAnchor)
+        poolWhitelisted(_poolAnchor)
+        greaterThanZero(_amount)
+        returns (uint256)
+    {
+        return addLiquidity(_recipient, _poolAnchor, _reserveToken, _amount);
```
yudilevi, line 291: any reason not to have the addLiquidity logic here?

solidity/contracts/liquidity-protection/LiquidityProtectionSettings.sol
```diff
@@ -209,23 +198,7 @@ contract LiquidityProtectionSettings is ILiquidityProtectionSettings, AccessCont
...
-        uint256 lastIndex = length - 1;
-        if (index < lastIndex) {
-            IConverterAnchor lastAnchor = poolWhitelist[lastIndex];
-            poolWhitelistIndices[lastAnchor].value = index;
-            poolWhitelist[index] = lastAnchor;
-        }
-
-        poolWhitelist.pop();
-        delete poolWhitelistIndices[_poolAnchor];
+        require(_poolWhitelist.remove(address(_poolAnchor)), "ERR_POOL_NOT_WHITELISTED");
```
yudilevi, line 201: wow :) so much nicer

solidity/contracts/liquidity-protection/LiquidityProtectionSettings.sol
```diff
@@ -237,7 +210,21 @@ contract LiquidityProtectionSettings is ILiquidityProtectionSettings, AccessCont
...
-        return poolWhitelistIndices[_poolAnchor].isValid;
+        return _poolWhitelist.contains(address(_poolAnchor));
+    }
+
+    /**
+     * @dev returns pools whitelist
+     *
+     * @return pools whitelist
+     */
+    function poolWhitelist() external view returns (address[] memory) {
```
yudilevi, line 221: can break if the list becomes long but I guess it's easy to upgrade

solidity/contracts/liquidity-protection/CheckpointStore.sol
lbeder: Hmmm... not really. If someone wants to checkpoint the 0x0 address - it's totally fine. I'll remove it.

solidity/contracts/liquidity-protection/LiquidityProtection.sol
lbeder: 👍

solidity/contracts/liquidity-protection/LiquidityProtection.sol
lbeder: 👍

solidity/contracts/liquidity-protection/LiquidityProtectionSettings.sol
lbeder: 💯

solidity/contracts/liquidity-protection/LiquidityProtection.sol
lbeder: I wanted to avoid double input verification here.

solidity/contracts/liquidity-protection/LiquidityProtectionSettings.sol
lbeder: Yeah

solidity/contracts/liquidity-protection/LiquidityProtection.sol
yudilevi: oh right, makes sense

solidity/test/helpers/Constants.js
barakman: Yes, sorry :)

yudilevi (approved):

barakman (approved):

PR #506, opened 2020-12-15
LiquidityProtectionStore Upgrade & Migration

barakman:


barakman:
1. Reduce gas consumption
2. Extend event information
3. Add migration infrastructure

PR #507, opened 2020-12-17
Add function `poolAvailableSpace`

barakman:


solidity/contracts/liquidity-protection/LiquidityProtection.sol
```diff
@@ -451,6 +451,51 @@ contract LiquidityProtection is TokenHandler, Utils, Owned, ReentrancyGuard, Tim
...
 
+    /**
+     * @dev returns the single-side staking limits of a given pool
+     *
+     * @param _poolAnchor   anchor of the pool
+     * @return base maximum amount of base token liquidity that can be single-sided staked in the pool
+     * @return network maximum amount of base token liquidity that can be single-sided staked in the pool
+     */
+    function poolAvailableSpace(IConverterAnchor _poolAnchor) external view returns (uint256, uint256) {
+        // get the pool token
```
lbeder, line 462: I think that this method should have the `poolSupported(_poolAnchor)` and `poolWhitelisted(_poolAnchor)` checks, so that it won't cause unexpected results for other converters

solidity/contracts/liquidity-protection/LiquidityProtection.sol
```diff
@@ -451,6 +451,51 @@ contract LiquidityProtection is TokenHandler, Utils, Owned, ReentrancyGuard, Tim
...
+        ILiquidityPoolConverter converter = ILiquidityPoolConverter(payable(ownedBy(_poolAnchor)));
+
+        // get the base token
+        IERC20Token baseToken = converter.connectorTokens(0);
+        if (baseToken == networkToken) {
+            baseToken = converter.connectorTokens(1);
+        }
+
+        // get the reserve balances
+        (uint256 reserveBalanceBase, uint256 reserveBalanceNetwork) =
```
lbeder, line 475: Mayme `baseTokenReserveBalance` and `networkTokenReserveBalance`?

solidity/test/LiquidityProtection.js
```diff
@@ -680,6 +684,38 @@ describe('LiquidityProtection', () => {
                 expect(newBalance).to.be.bignumber.equal(new BN(0));
             });
 
+            for (let x = 24; x < 30; x++) {
```
lbeder, line 687: Great tests! 💪

lbeder (changes requested):

solidity/contracts/liquidity-protection/LiquidityProtection.sol
barakman: We have the same naming issue cross system, sometimes using 'reserve', sometimes using `token` and sometimes using 'reserveToken'.
Due to ETH_RESERVE, it is probably wrong to use `token` in the general case.
Anyway, I don't think it's the right time and place to fix this issue, and fixing it here will just be a drop in the ocean. I actually copied that line from the function which imposes the limit (`addBaseTokenLiquidity`).

solidity/test/LiquidityProtection.js
barakman: TX.
Should probably be a lot more thorough, with two more loop scopes - one for the initial balance of the pool in the base token, and one for the initial balance of the pool in the network token.
Unfortunately, the way that our `beforeEach` clause and `initPool` functions are written, it's quite a "heavy" task to change those initial balances (unless we do the whole thing in a separate test file).
In addition to all of the above, it would be useful to choose to use in those loops various amounts which are not necessarily divisible by each other. If the `poolAvailableSpace` function is called under some configuration of the pool, and misses by a single wei, a user transaction based on its return value will revert.

solidity/contracts/liquidity-protection/LiquidityProtection.sol
```diff
@@ -451,6 +451,67 @@ contract LiquidityProtection is TokenHandler, Utils, Owned, ReentrancyGuard, Tim
...
+        uint256 mintingLimit = settings.networkTokenMintingLimits(_poolAnchor);
+        if (mintingLimit == 0) {
+            mintingLimit = settings.defaultNetworkTokenMintingLimit();
+        }
+
+        // get the amount of network tokens minted for the pool
+        uint256 networkTokensMinted = settings.networkTokensMinted(_poolAnchor);
+
+        // return the maximum amount of base token liquidity that can be single-sided staked in the pool
+        return mintingLimit.sub(networkTokensMinted).mul(reserveBalanceBase).div(reserveBalanceNetwork);
```
yudilevi, line 495: limit might be lower than the minted amount which will result in an exception.

solidity/contracts/liquidity-protection/LiquidityProtection.sol
```diff
@@ -451,6 +451,67 @@ contract LiquidityProtection is TokenHandler, Utils, Owned, ReentrancyGuard, Tim
...
+        return (baseTokenAvailableSpace(_poolAnchor), networkTokenTokenAvailableSpace(_poolAnchor));
+    }
+
+    /**
+     * @dev returns the base-token staking limits of a given pool
+     *
+     * @param _poolAnchor   anchor of the pool
+     * @return maximum amount of base tokens that can be single-side staked in the pool
+     */
+    function baseTokenAvailableSpace(IConverterAnchor _poolAnchor) internal view returns (uint256) {
```
yudilevi, line 471: might as well make these functions public to increase flexility.
also, since they are separated, we can re-use them to save some code (unless it's not worth it gas-wise).

yudilevi (changes requested):

solidity/contracts/liquidity-protection/LiquidityProtection.sol
barakman: Of course, I think it is actually the current state on the ETHBNT pool.
But what do you wanna do about it, return 0 if `mintingLimit < networkTokensMinted`?

yudilevi: yup

solidity/contracts/liquidity-protection/LiquidityProtection.sol
barakman: I've considered that (using each one of the getter functions in the corresponding setter function), however, it would actually increase the gas-cost of each one of the setter functions, since some of the storage read operations and the external function calls would actually be duplicated (i.e., both in the getter functions and in the setter functions).

We can find a way to flatten all of that (e.g., using several private functions which will receive the common data and process it), but it will uglify the code significantly, and we will possibly find ourselves dealing with 'stack too deep' and other issues.

In short, it might require a lot of work to do code-reuse here, which won't reduce the contract bytecode size (if that's what you're aiming at), and can even pose a certain risk - we can live with a bug in the getter function, but not if it's used in by a setter function.

barakman: `might as well make these functions public to increase flexibility` - done.

solidity/contracts/liquidity-protection/LiquidityProtection.sol
barakman: Done.

yudilevi (approved):

solidity/contracts/liquidity-protection/LiquidityProtection.sol
barakman: Done

yudilevi (approved):

lbeder (approved):

PR #509, opened 2020-12-23
Upgrader Converter Upgrader

barakman:
Upon upgrading a converter, withdraw the ether or tokens of each reserve only if the balance of the converter in that reserve is larger than zero.


yudilevi (commented):
missing test

barakman:
> missing test

What exactly do you wanna test here?
Our tokens allow `transfer(0)`, so such test would pass with and without this fix.

yudilevi (approved):

PR #512, opened 2020-12-31
Improve the test-deployment script

barakman:
In `test_deployment.js`, handle standard errors in a more efficient manner (i.e., automatic retry).


barakman:
Needs some testing first, to see that it works as intended.

PR #514, opened 2021-01-03
Migrate liquidity protection store

barakman:
Please review only contract changes for now.


solidity/contracts/liquidity-protection/LiquidityProtection.sol
```diff
@@ -72,6 +73,7 @@ contract LiquidityProtection is TokenHandler, Utils, Owned, ReentrancyGuard, Tim
 
     ILiquidityProtectionSettings public immutable settings;
     ILiquidityProtectionStore public immutable store;
+    ILiquidityProtectionStats public immutable stats;
```
yudilevi, line 76: can use `using SafeERC20 for IERC20;` with OZ's SafeERC20, cleaner

solidity/contracts/liquidity-protection/LiquidityProtectionStore.sol
```diff
@@ -11,51 +11,73 @@ import "../utility/Utils.sol";
...
         uint256 amount; // amount of network tokens
         uint256 expirationTime; // lock expiration time
     }
 
-    // protected liquidity by provider
-    uint256 private nextProtectedLiquidityId;
-    mapping(address => uint256[]) private protectedLiquidityIdsByProvider;
-    mapping(uint256 => ProtectedLiquidity) private protectedLiquidities;
+    // position by provider
+    uint256 private nextPositionId;
```
yudilevi, line 47: still need to change the conventions in the entire contract to OZ's

solidity/contracts/liquidity-protection/LiquidityProtectionStore.sol
```diff
@@ -11,51 +11,73 @@ import "../utility/Utils.sol";
...
     mapping(address => LockedBalance[]) private lockedBalances;
 
     // system balances
     mapping(IERC20Token => uint256) private systemBalances;
 
-    // total protected pool supplies / reserve amounts
-    mapping(IDSToken => uint256) private totalProtectedPoolAmounts;
-    mapping(IDSToken => mapping(IERC20Token => uint256)) private totalProtectedReserveAmounts;
+    // allows execution by the owner only
+    modifier ownerOnly {
```
yudilevi, line 58: missing binary size optimizations that we use for modifiers + modifiers before ctor?

solidity/contracts/liquidity-protection/LiquidityProtection.sol
```diff
@@ -86,29 +88,34 @@ contract LiquidityProtection is TokenHandler, Utils, Owned, ReentrancyGuard, Tim
      *
      * @param _settings liquidity protection settings
      * @param _store liquidity protection store
+     * @param _store liquidity protection stats
```
lbeder, line 91: Should be _stats

solidity/contracts/liquidity-protection/LiquidityProtection.sol
```diff
@@ -86,29 +88,34 @@ contract LiquidityProtection is TokenHandler, Utils, Owned, ReentrancyGuard, Tim
...
     )
         public
         validAddress(address(_settings))
         validAddress(address(_store))
+        validAddress(address(_stats))
         validAddress(address(_networkTokenGovernance))
         validAddress(address(_govTokenGovernance))
         notThis(address(_settings))
         notThis(address(_store))
+        notThis(address(_stats))
```
lbeder, line 112: Really not important (and you weren't aware of it), but you can now use the new `validExternalAddress` modifier which combines `validAddress` and `notThis`

solidity/contracts/liquidity-protection/LiquidityProtectionStats.sol
```diff
@@ -0,0 +1,131 @@
+// SPDX-License-Identifier: SEE LICENSE IN LICENSE
+pragma solidity 0.6.12;
+import "./interfaces/ILiquidityProtectionStats.sol";
+import "@openzeppelin/contracts/access/AccessControl.sol";
```
lbeder, line 4: Ideally, contracts from external repos should come first, thus helping us detect issues when we're trying to redefine (or in conflict with) external changes

solidity/contracts/liquidity-protection/LiquidityProtectionStore.sol
```diff
@@ -251,112 +287,94 @@ contract LiquidityProtectionStore is ILiquidityProtectionStore, Owned, TokenHand
...
-        IERC20Token reserveToken = liquidity.reserveToken;
-        uint256 prevPoolAmount = liquidity.poolAmount;
-        uint256 prevReserveAmount = liquidity.reserveAmount;
-        liquidity.poolAmount = _newPoolAmount;
-        liquidity.reserveAmount = _newReserveAmount;
+        IDSToken poolToken = pos.poolToken;
+        IERC20Token reserveToken = pos.reserveToken;
+        uint256 prevPoolAmount = uint256(pos.poolAmount);
+        uint256 prevReserveAmount = uint256(pos.reserveAmount);
+        pos.poolAmount = toUint128(_newPoolAmount);
```
yudilevi, line 334: if we only support uint128, why not pass in uint128 as the arg types?

solidity/contracts/liquidity-protection/LiquidityProtectionStats.sol
```diff
@@ -0,0 +1,131 @@
...
+     * @param _poolToken        pool token address
+     * @param _reserveToken     reserve token address
+     */
+    function resetTotalAmounts(
+        address _provider,
+        IDSToken _poolToken,
+        IERC20Token _reserveToken
+    ) external override ownerOnly {
+        totalPoolAmount[_poolToken] = 0;
+        totalReserveAmount[_poolToken][_reserveToken] = 0;
```
lbeder, line 99: This looks like a bug. You shouldn't set total amounts to 0 if a single provider removes his liquidity. Do you have a test for this?

solidity/contracts/liquidity-protection/LiquidityProtectionStats.sol
```diff
@@ -0,0 +1,131 @@
...
+        address[] memory _tokens,
+        address[] memory _reserve0s,
+        address[] memory _reserve1s,
+        address[] memory _providers,
+        uint256[] memory _poolAmounts,
+        uint256[] memory _reserve0Amounts,
+        uint256[] memory _reserve1Amounts,
+        uint256[] memory _provider0Amounts,
+        uint256[] memory _provider1Amounts
+    ) external seederOnly {
```
lbeder, line 113: Using `calldata` will reduce gas costs by a (little) bit.

solidity/contracts/liquidity-protection/LiquidityProtectionStore.sol
```diff
@@ -527,28 +545,120 @@ contract LiquidityProtectionStore is ILiquidityProtectionStore, Owned, TokenHand
...
-     * @param _reserveToken    reserve token address
-     * @return total protected amount
-     */
-    function totalProtectedReserveAmount(IDSToken _poolToken, IERC20Token _reserveToken)
-        external
-        view
-        returns (uint256)
-    {
-        return totalProtectedReserveAmounts[_poolToken][_reserveToken];
+    function encodeReserveRateInfo(
```
yudilevi, line 553: need to add a public utility function for decoding reserve rate info to ease on integrations

solidity/contracts/liquidity-protection/LiquidityProtectionStats.sol
```diff
@@ -0,0 +1,131 @@
...
+        uint256 length = _tokens.length;
+        require(length == _reserve0s.length);
```
lbeder, line 115: Please don't add `require`s without a message :(

solidity/contracts/liquidity-protection/LiquidityProtectionStore.sol
```diff
@@ -527,28 +545,120 @@ contract LiquidityProtectionStore is ILiquidityProtectionStore, Owned, TokenHand
...
+
+    function decodeReserveRateT(uint256 _reserveRateInfo) private pure returns (uint256) {
+        return _reserveRateInfo >> 224;
+    }
+
+    function seedNextPositionId(uint256 _nextPositionId) external seederOnly {
+        nextPositionId = _nextPositionId;
+    }
+
+    function seedPositions(
```
yudilevi, line 578: it's a monster! :D
technically we could have used a separate contract for migration to keep this one cleaner but it will increase gas costs.
in this specific case, since it will probably only increase gas costs by roughly 1-2%, we should consider it

solidity/contracts/liquidity-protection/LiquidityProtectionStats.sol
```diff
@@ -0,0 +1,131 @@
...
+        require(length == _reserve1s.length);
+        require(length == _providers.length);
+        require(length == _poolAmounts.length);
+        require(length == _reserve0Amounts.length);
+        require(length == _reserve1Amounts.length);
+        require(length == _provider0Amounts.length);
+        require(length == _provider1Amounts.length);
```
lbeder, line 122: You can do all these checks in a single `require` (especially if there is no/same error message)

yudilevi (changes requested):

solidity/contracts/liquidity-protection/LiquidityProtectionStats.sol
```diff
@@ -0,0 +1,131 @@
...
+        for (uint256 i = 0; i < length; i++) {
+            totalPoolAmount[IDSToken(_tokens[i])] = _poolAmounts[i];
```
lbeder, line 124: Caching all `xxx[i]` might reduce gas by a little bit

solidity/contracts/liquidity-protection/LiquidityProtectionStore.sol
```diff
@@ -1,7 +1,7 @@
 // SPDX-License-Identifier: SEE LICENSE IN LICENSE
 pragma solidity 0.6.12;
 import "./interfaces/ILiquidityProtectionStore.sol";
-import "../utility/Owned.sol";
+import "@openzeppelin/contracts/access/AccessControl.sol";
```
lbeder, line 4: Same as above

solidity/contracts/liquidity-protection/LiquidityProtectionStore.sol
```diff
@@ -527,28 +545,120 @@ contract LiquidityProtectionStore is ILiquidityProtectionStore, Owned, TokenHand
...
+    function seedPositions(
+        uint256[] memory _ids,
+        address[] memory _providers,
+        address[] memory _poolTokens,
+        address[] memory _reserveTokens,
+        uint256[] memory _poolAmounts,
+        uint256[] memory _reserveAmounts,
+        uint256[] memory _reserveRateNs,
+        uint256[] memory _reserveRateDs,
+        uint256[] memory _timestamps
```
lbeder, line 587: Same as above re: `calldata`

solidity/contracts/liquidity-protection/LiquidityProtectionStore.sol
```diff
@@ -527,28 +545,120 @@ contract LiquidityProtectionStore is ILiquidityProtectionStore, Owned, TokenHand
...
+        uint256 length = _ids.length;
+        require(length == _providers.length);
+        require(length == _poolTokens.length);
+        require(length == _reserveTokens.length);
+        require(length == _poolAmounts.length);
+        require(length == _reserveAmounts.length);
+        require(length == _reserveRateNs.length);
+        require(length == _reserveRateDs.length);
+        require(length == _timestamps.length);
+        for (uint256 i = 0; i < length; i++) {
```
lbeder, line 598: Same as above re: error message + single require + caching

lbeder (changes requested):

solidity/contracts/liquidity-protection/LiquidityProtectionStore.sol
barakman: @yudilevi : The original intention (before it was decided to redeploy the main contract), was to avoid breaking the API of the store contract.

solidity/contracts/liquidity-protection/LiquidityProtection.sol

solidity/contracts/liquidity-protection/LiquidityProtection.sol
barakman: @yudilevi : What's the verdict? Use `validExternalAddress`, or we scrap of the `notThis` verification and use `validAddress` (which is what I currently changed the code to, due to a 'stack-too-deep' error)?

solidity/contracts/liquidity-protection/LiquidityProtectionStats.sol
barakman: I usually go by the order of inheritance in the contract declaration (where I always place the contract's own interface first), but OK.

solidity/contracts/liquidity-protection/LiquidityProtectionStore.sol
barakman: I don't understand this requirement. The reserve-rate info is a field in a private mapping.
Function `position(uint256 _id)` already returns its decoded values (N, D and timestamp).
There is no way to access this field from outside of the contract, so you won't have any proper input to pass to such utility function.

solidity/contracts/liquidity-protection/LiquidityProtectionStats.sol
barakman: @lbeder : Fixed; can you please check?

solidity/contracts/liquidity-protection/LiquidityProtectionStats.sol
barakman: Fixed.

solidity/contracts/liquidity-protection/LiquidityProtectionStats.sol
barakman: Fixed.

solidity/contracts/liquidity-protection/LiquidityProtectionStore.sol
barakman: We don't have a binary-size problem here.
This optimization is on the expanse of higher gas-cost during runtime, so we're probably better off without it.

solidity/contracts/liquidity-protection/LiquidityProtectionStore.sol
barakman: Another contract which would be the seeder of this contract?
That would increase the gas cost by a lot more than 1-2% IMO, plus, it won't even make that "monster" any less of a monster than it already is. Instead of input arrays, we will have input variables, and the `for` loop will be removed. That's very little reduction in text size, plus, why do we care about this text in the contract anyway?
On top of all of that, adding another mediator here (the seeding contract) will just add a lot more complexity to the already-complicated migration process.

solidity/contracts/liquidity-protection/LiquidityProtectionStats.sol
barakman: Fixed.

solidity/contracts/liquidity-protection/LiquidityProtectionStore.sol
barakman: Fixed.

solidity/contracts/liquidity-protection/LiquidityProtection.sol
barakman: Changing to from `TokenHandler` to `SafeERC20` requires changing from `IERC20Token` to `IERC20`, which in turn requires changes all over the place (LP contract, LP test contract, LP interface, `DSToken` contract, `DSToken` interface and probably any other contract or interface which relies on this interface).
We shouldn't apply this fix as part of the migration task, but in a separate task (and on a separate branch).

barakman:
This PR is no longer relevant, but the branch may still be relevant, so keeping it for now.

PR #516, opened 2021-01-13
LP notifications and preparation for integration

lbeder:
This PR contains the following changes:
1. Making the project integrable with other projects.
2. Replace internal SafeMath with OZ and rename Math to MathEx in order to prevent conflict with the LM repo.
3. Add LP notifications.
4. Missing formatting (for some reason).


solidity/contracts/liquidity-protection/LiquidityProtection.sol
```diff
@@ -612,6 +656,18 @@ contract LiquidityProtection is ILiquidityProtection, TokenHandler, Utils, Owned
             );
         }
 
+        // notify event subscribers
+        if (address(eventsSubscriber) != address(0)) {
```
yudilevi, line 660: called after removing liquidity (but before stats are updated) - intentional?

solidity/contracts/liquidity-protection/LiquidityProtection.sol
```diff
@@ -856,6 +912,11 @@ contract LiquidityProtection is ILiquidityProtection, TokenHandler, Utils, Owned
         uint256 _poolAmount,
         uint256 _reserveAmount
     ) internal returns (uint256) {
+        // notify event subscribers
+        if (address(eventsSubscriber) != address(0)) {
```
yudilevi, line 916: called before provider pool is set - does it not need to know the provider has liquidity in that pool at this point?

lbeder: It doesn't, since its major goal is to only "sync" the previous state

solidity/contracts/liquidity-protection/LiquidityProtection.sol
lbeder: Hmmm... it'd be cleaner to move it before the liquidity is removed, although it won't affect the current implementation

yudilevi (approved):

PR #523, opened 2021-01-18
Liquidity protection store upgrade

barakman:
Handled:
- System Store
- Token Holder

Not Handled:
- User Store


yudilevi:
@barakman should we close this one?

PR #524, opened 2021-01-19
Add static-pool

barakman:


solidity/contracts/converter/types/static-pool/StaticPoolConverter.sol
```diff
@@ -0,0 +1,42 @@
+// SPDX-License-Identifier: SEE LICENSE IN LICENSE
+pragma solidity 0.6.12;
+
+import "../standard-pool/StandardPoolConverter.sol";
+
+/**
+ * @dev This contract is a specialized version of the converter, which is
```
yudilevi, line 7: missing dedicated comment

barakman: Fixed.

PR #526, opened 2021-01-28
Add function `moveLiquidity`

barakman:


yudilevi:
this should go live with the next version of the contracts so we need to spend some time to clear up the design

PR #537, opened 2021-02-15
Cross reserve source amount

barakman:


solidity/contracts/converter/types/standard-pool/StandardPoolConverter.sol
```diff
@@ -1111,6 +1153,22 @@ contract StandardPoolConverter is
...
     }
 
+    function crossReserveSourceAmount(
+        uint256 _sourceReserveBalance,
+        uint256 _targetReserveBalance,
+        uint256 _amount
+    ) private pure returns (uint256) {
+        // validate input
+        require(_sourceReserveBalance > 0, "ERR_INVALID_RESERVE_BALANCE");
+        require(_amount < _targetReserveBalance, "ERR_INVALID_AMOUNT");
```
yudilevi, line 1163: we're comparing amount in source token with the target reserve balance - mistake?

yudilevi (approved):

PR #538, opened 2021-02-15
Switch to yarn and upgrade dependencies

lbeder:


package.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "@bancor/contracts-solidity",
-    "version": "0.6.32",
+    "version": "1.22.10",
```
yudilevi, line 3: this is the contracts version, no? shouldn't it stay as 0.6.32 (or 0.6.33)?

lbeder: Oh, my bad. I'll revert

yudilevi (approved):

PR #540, opened 2021-02-20
record spot rate on deposit instead of avg. rate

barakman:


solidity/contracts/liquidity-protection/LiquidityProtection.sol
```diff
@@ -963,23 +965,6 @@ contract LiquidityProtection is ILiquidityProtection, TokenHandler, Utils, Owned
...
     }
 
-    /**
-     * @dev returns the average rate of 1 reserve token in the other reserve token units
-     *
-     * @param _poolToken            pool token
-     * @param _reserveToken         reserve token
-     * @param _validateAverageRate  true to validate the average rate; false otherwise
-     */
-    function reserveTokenAverageRate(
```
yudilevi, line 973: average rate should still be used when calculating the IL (not when calculating the fees).

yudilevi (changes requested):

solidity/contracts/helpers/TestLiquidityProtection.sol
```diff
@@ -52,8 +52,8 @@ contract TestLiquidityProtection is LiquidityProtection, TestTime {
     }
 
     function averageRateTest(IDSToken _poolToken, IERC20Token _reserveToken) external view returns (uint256, uint256) {
-        Fraction memory rate = reserveTokenAverageRate(_poolToken, _reserveToken, true);
-        return (rate.n, rate.d);
+        (, , uint256 rateN, uint256 rateD) = reserveTokenRates(_poolToken, _reserveToken, true);
```
yudilevi, line 55: I don't quite get this - it's the average rate test function that calls the non average rate, no?

barakman: Function `reserveTokenRates` returns the spot rate (2 values) and the average rate (2 values).
Here, we ignore the first 2 values and use the last 2 values.

yudilevi (approved):

PR #541, opened 2021-02-20
Replace ERC20Token and TokenHandler with OZ

lbeder:


solidity/contracts/converter/ConverterUpgrader.sol
```diff
@@ -134,9 +133,8 @@ contract ConverterUpgrader is IConverterUpgrader, ContractRegistryClient {
         // new converter - get the type from the converter itself
         if (isV28OrHigherConverter(_oldConverter))
             newType = _oldConverter.converterType();
-        // old converter - if it has 1 reserve token, the type is a liquid token, otherwise the type liquidity pool
-        else if (reserveTokenCount > 1)
-            newType = 1;
+            // old converter - if it has 1 reserve token, the type is a liquid token, otherwise the type liquidity pool
```
barakman, line 136: This looks like a bad linter decision.
Adding curly braces will probably solve this.

barakman (changes requested):

solidity/contracts/converter/ConverterUpgrader.sol
lbeder: Yeah, you're right. I'm trying to fix these cases here and there but missed this one. Good catch!

lbeder:
@barakman I've fixed the issue above and also merged the latest rate changes to this branch. FYI

barakman (approved):

PR #544, opened 2021-02-22
Add support for function `poweredRatio`

barakman:


solidity/contracts/utility/MathEx.sol
```diff
@@ -34,6 +38,42 @@ library MathEx {
...
+     * @param _d   ratio denominator
+     * @param _exp ratio exponent
+     *
+     * @return powered ratio's numerator and denominator
+     */
+    function poweredRatio(
+        uint256 _n,
+        uint256 _d,
+        uint256 _exp
+    ) internal pure returns (uint256, uint256) {
```
lbeder, line 54: This can be a good opportunity to convert the whole contract (or only the new function) to the new coding conventions we're using with new code

solidity/contracts/utility/MathEx.sol
```diff
@@ -34,6 +38,42 @@ library MathEx {
...
+        require(_exp <= MAX_EXP, "ERR_EXP_TOO_LARGE");
+
+        uint256[MAX_EXP_BIT_LEN] memory ns;
+        uint256[MAX_EXP_BIT_LEN] memory ds;
+
+        (ns[0], ds[0]) = reducedRatio(_n, _d, MAX_UINT128);
+        for (uint256 i = 0; (_exp >> i) > 1; i++) {
```
lbeder, line 61: Since `i` is incrementing by 1 every time, couldn't you keep shifting `_exp` once as well and just cache the previous result? Might save a wee bit of gas

solidity/test/helpers/MathUtils.js
```diff
@@ -20,6 +21,11 @@ function ceilSqrt(n) {
     return Decimal(n).sqrt().ceil().toFixed();
 }
 
+function poweredRatio(a, b, exp) {
```
lbeder, line 24: We generally prefer arrow notation (const powerRation = (a, b, exp) =>) in js as well

solidity/contracts/utility/MathEx.sol
barakman: Variable `i` is used also inside the loop, so we cannot get rid of it.
Hence, the only thing we can do here is to replace the condition `(_exp >> i) > 1` with `_exp > 1`, and then inside the loop - shift `_exp` by 1 and check `(_exp & 1) > 0` instead of checking `((_exp >> i) & 1) > 0`.
In short, we will only be replacing 2 shifts with 1 shift and 1 assignment. I doubt that it would actually save any gas.

lbeder: I didn't mean to get rid of `i`, but rather to do `_exp >> 1` inside the loop every step. Otherwise, it'll currently do `_exp >> 1`, then `_exp >> 2`, then `_exp >> 3`, and so on.

barakman: You mean `_exp >>= 1`, which consists of 1 load, 1 shift and 1 store.

In the current implementation, `_exp >> i` consists of 2 loads and 1 shift.

Pretty much the same, no?

lbeder: Hmmm... yes, you're right

lbeder (approved):

yudilevi (approved):

yudilevi:
Is this ready for merge?

barakman:
> Is this ready for merge?

As far as I'm concerned - yes (it's just an additional function (currently not used anywhere) + tests.

yudilevi:
So what is it waiting for? Was approved by all reviewers

PR #546, opened 2021-02-25
Remove affiliate-fee.

barakman:


solidity/contracts/BancorNetwork.sol
```diff
@@ -207,48 +194,34 @@ contract BancorNetwork is TokenHolder, ContractRegistryClient, ReentrancyGuard {
...
      * @param _amount              amount to convert from, in the source token
      * @param _minReturn           if the conversion results in an amount smaller than the minimum return - it is cancelled, must be greater than zero
      * @param _beneficiary         account that will receive the conversion result or 0x0 to send the result to the sender account
-     * @param _affiliateAccount    wallet address to receive the affiliate fee or 0x0 to disable affiliate fee
-     * @param _affiliateFee        affiliate fee in PPM or 0 to disable affiliate fee
      *
      * @return amount of tokens received from the conversion
      */
-    function convertByPath(
+    function convertByPath2(
```
lbeder, line 206: Why do we need to rename this method?

solidity/contracts/BancorNetwork.sol
```diff
@@ -594,11 +517,7 @@ contract BancorNetwork is TokenHolder, ContractRegistryClient, ReentrancyGuard {
 
             // first check if the converter in this step is newer as older converters don't even support the beneficiary argument
             if (stepData.isV28OrHigherConverter) {
-                // if affiliate fee is processed in this step, beneficiary is the network contract
-                if (stepData.processAffiliateFee)
-                    stepData.beneficiary = payable(address(this));
-                    // if it's the last step, beneficiary is the final beneficiary
-                else if (i == data.length - 1)
+                if (i == data.length - 1)
```
lbeder, line 520: Could you add brackets here?

solidity/contracts/BancorNetwork.sol
```diff
@@ -723,10 +656,10 @@ contract BancorNetwork is TokenHolder, ContractRegistryClient, ReentrancyGuard {
         address[] memory _path,
         uint256 _amount,
         uint256 _minReturn,
-        address _affiliateAccount,
-        uint256 _affiliateFee
+        address /* _affiliateAccount */,
+        uint256 /* _affiliateFee */
```
lbeder, line 660: @yudilevi do we really need to keep these interfaces?

lbeder (changes requested):

solidity/contracts/BancorNetwork.sol
barakman: Because it is part of the API, and we don't want to break any application which relies on this.

solidity/contracts/BancorNetwork.sol
barakman: It is part of the API, and we don't want to break any application which relies on this.

solidity/contracts/BancorNetwork.sol
barakman: Done all across the file; Can you please re-review?

solidity/contracts/BancorNetwork.sol
lbeder: 🦁 🦁 🦁

lbeder (approved):

yudilevi (approved):

PR #547, opened 2021-02-26
Remove ether token.

barakman:


solidity/contracts/token/EtherToken.sol
```diff
@@ -1,118 +0,0 @@
-// SPDX-License-Identifier: SEE LICENSE IN LICENSE
```
yudilevi, line 1: move to the legacy contracts repo

solidity/contracts/BancorNetwork.sol
```diff
@@ -456,19 +414,8 @@ contract BancorNetwork is TokenHolder, ContractRegistryClient, ReentrancyGuard {
...
-            assert(!stepData.isV28OrHigherConverter);
-
-            // EtherToken converter - withdraw the ETH and transfer to the beneficiary
-            IEtherToken(address(targetToken)).withdrawTo(_beneficiary, _amount);
-        }
-        // other ERC20 token
-        else {
-            targetToken.safeTransfer(_beneficiary, _amount);
-        }
+        assert(targetToken != ETH_RESERVE_ADDRESS);
```
lbeder, line 417: Perhaps a `require`?

solidity/contracts/BancorNetwork.sol
```diff
@@ -502,50 +449,23 @@ contract BancorNetwork is TokenHolder, ContractRegistryClient, ReentrancyGuard {
...
-            }
-        }
-
         // set the beneficiary for each step
         for (i = 0; i < data.length; i++) {
-            stepData = data[i];
-
             // check if the converter in this step is newer as older converters don't even support the beneficiary argument
-            if (stepData.isV28OrHigherConverter) {
+            if (data[i].isV28OrHigherConverter) {
```
lbeder, line 455: Caching `data[i]` can slightly reduce gas

lbeder (changes requested):

solidity/contracts/BancorNetwork.sol
barakman: If this condition is not met, then it's because of a bug in the contract, not because of a wrong user input.
So IMO, `assert` is appropriate here, while `require` is not.

solidity/contracts/token/EtherToken.sol
barakman: Done.

solidity/contracts/BancorNetwork.sol
barakman: Done.

lbeder (approved):

yudilevi (approved):

PR #549, opened 2021-03-02
Network settings contract

barakman:


solidity/test/NetworkSettings.js
```diff
@@ -0,0 +1,87 @@
+const { contract, accounts } = require('@openzeppelin/test-environment');
+const { BN, constants, expectRevert } = require('@openzeppelin/test-helpers');
+const { expect } = require('../../chai-local');
+
+const NetworkSettings = contract.fromArtifact('NetworkSettings');
+
+const PPM_RESOLUTION = new BN(1000000);
+
+describe('NetworkSettings', () => {
```
yudilevi, line 9: missing permission checks (!!!)

solidity/contracts/NetworkSettings.sol
```diff
@@ -0,0 +1,65 @@
...
+import "./utility/Owned.sol";
+import "./utility/Utils.sol";
+
+/**
+ * @dev This contract maintains the network settings.
+ */
+contract NetworkSettings is INetworkSettings, Owned, Utils {
+    uint32 private constant PPM_RESOLUTION = 1000000;
+
+    address private _feeWallet;
```
yudilevi, line 14: even though these aren't called from a contract, we should still use names that provide the context.
we might introduce other types of fees like affiliate fee here.
I'd also prefer to use just fee and not "portion", I think it's easier to understand and in line with the (conversion) fee in the pools.

yudilevi (changes requested):

solidity/test/NetworkSettings.js
barakman: Which ones???

solidity/contracts/NetworkSettings.sol
barakman: So what name do you wanna use?

solidity/test/NetworkSettings.js
yudilevi: all of them? owner only calls?

solidity/contracts/NetworkSettings.sol
yudilevi: networkFee
networkFeeWallet

solidity/contracts/NetworkSettings.sol
```diff
@@ -0,0 +1,65 @@
...
+    uint32 private _feePortion;
+
+    // ensures that the portion is valid
+    modifier validPortion(uint32 _portion) {
+        _validPortion(_portion);
+        _;
+    }
+
+    // error message binary size optimization
+    function _validPortion(uint32 _portion) internal pure {
```
lbeder, line 24: We use these modifiers in 3 places already, so perhaps it should go to Utils?

solidity/test/NetworkSettings.js
yudilevi: review issue, forget it

yudilevi (changes requested):

solidity/contracts/NetworkSettings.sol
```diff
@@ -0,0 +1,65 @@
...
+pragma solidity 0.6.12;
+
+import "./INetworkSettings.sol";
+import "./utility/Owned.sol";
+import "./utility/Utils.sol";
+
+/**
+ * @dev This contract maintains the network settings.
+ */
+contract NetworkSettings is INetworkSettings, Owned, Utils {
```
lbeder, line 11: @yudilevi @barakman When do we want to add it to the test deployment script and the client registry? In this PR or the PR which will be actually using it?

solidity/test/NetworkSettings.js
```diff
@@ -0,0 +1,87 @@
...
+    it('should revert when creating a contract with a too-small fee portion', async () => {
+        await expectRevert(NetworkSettings.new(validFeeWallet1, tooSmallFeePortion), 'ERR_INVALID_PORTION');
+    });
+
+    it('should revert when creating a contract with a too-large fee portion', async () => {
+        await expectRevert(NetworkSettings.new(validFeeWallet1, tooLargeFeePortion), 'ERR_INVALID_PORTION');
+    });
+
+    it('should revert when setting an invalid fee wallet', async () => {
+        let networkSettings = await NetworkSettings.new(validFeeWallet1, validFeePortion1);
```
lbeder, line 42: `neworkSettings` in this test and the tests below can be DRYed into a `beforeEach` callback

solidity/test/NetworkSettings.js
```diff
@@ -0,0 +1,87 @@
...
+        await expectRevert(networkSettings.setFeeWallet(invalidFeeWallet), 'ERR_INVALID_ADDRESS');
+        await expectReturn(networkSettings.feeParams(), {0: validFeeWallet1, 1: validFeePortion1});
```
lbeder, line 44: This is a weird syntax, but I do understand why you're doing this. I'd personally compare the returned array explicitly, but it's fine

lbeder (changes requested):

solidity/contracts/NetworkSettings.sol
barakman: Done.

solidity/contracts/NetworkSettings.sol
barakman: Yeah, same goes for `uint32 internal constant PPM_RESOLUTION = 1000000;` IMO.

solidity/test/NetworkSettings.js
barakman: Not really. See the first few `it` clauses.

solidity/test/NetworkSettings.js
barakman: We don't have a unified set of rules for comparisons in our tests, so I just went along with this option...

solidity/contracts/NetworkSettings.sol
barakman: Can do it in this PR.

solidity/contracts/NetworkSettings.sol
lbeder: 👍

solidity/contracts/NetworkSettings.sol
lbeder: 👍

solidity/test/NetworkSettings.js
lbeder: I'm not referring to the "bad" cases, but only to the "good" cases below this line. This is also why I usually wrap the "good" cases in an explicit `describe`, so that you could reuse stuff.

solidity/contracts/NetworkSettings.sol
barakman: Done.

solidity/contracts/NetworkSettings.sol
barakman: Second thought, it looks pointless (redundant) in this PR; will do it when it is actually needed (i.e., in the test-deployment script, in the registry-client contract and perhaps also in one or more tests).

solidity/test/NetworkSettings.js
barakman: So split into two describes - one for testing construction with illegal input, and one for all the other tests, and then add a `beforeEach` clause in the 2nd describe?
It will overall yield more text in this specific test, and of course, will not speed up anything.

lbeder: It's not supposed to speed anything... it's removing code duplication by moving shared code into a single locaiton

lbeder: And also separating semantically different test contexts

solidity/contracts/NetworkSettings.sol
lbeder: Alright

solidity/contracts/NetworkSettings.sol
lbeder: Can you also remove it from other contracts which reference it?

barakman: Reference what?

lbeder: Mi scusi!

solidity/test/NetworkSettings.js
barakman: Done.

lbeder: 💪 💪 💪

solidity/contracts/utility/TokenHolder.sol
```diff
@@ -35,6 +37,11 @@ contract TokenHolder is ITokenHolder, Owned, Utils {
         address _to,
         uint256 _amount
     ) public virtual override ownerOnly validAddress(address(_token)) validAddress(_to) notThis(_to) {
```
lbeder, line 39: Just an FYI, but we have a `validExternalAddress` modifier that combines both of these checks together

barakman: We don't need the 'notThis' check IMO, it's a pure waste of gas (and bytecode size).

lbeder: In this case, 100%

lbeder (approved):

solidity/contracts/NetworkSettings.sol
```diff
@@ -0,0 +1,52 @@
...
+    address private _networkFeeWallet;
```
yudilevi, line 12: still need specific getters for these, along with a description for what which one is (network fee = portion that's taken from all swap fees etc.)

solidity/contracts/utility/Utils.sol
```diff
@@ -1,10 +1,15 @@
...
 pragma solidity 0.6.12;
 
+import "@openzeppelin/contracts/token/ERC20/IERC20.sol";
+
 /**
  * @dev Utilities & Common Modifiers
  */
 contract Utils {
+    uint32 internal constant PPM_RESOLUTION = 1000000;
+    IERC20 internal constant ETH_RESERVE_ADDRESS = IERC20(0xEeeeeEeeeEeEeeEeEeEeeEEEeeeeEeeeeeeeEEeE);
```
yudilevi, line 11: rename to `NATIVE_TOKEN_ADDRESS`

yudilevi (changes requested):

solidity/contracts/NetworkSettings.sol
barakman: We have a getter, it's called `feeParams`.

yudilevi: I know, I want dedicated getters as well

barakman: Done.

solidity/contracts/utility/Utils.sol
barakman: Done.

solidity/contracts/NetworkSettings.sol
```diff
@@ -30,23 +30,41 @@ contract NetworkSettings is INetworkSettings, Owned, Utils {
         return (_networkFeeWallet, _networkFee);
     }
 
+    /**
+     * @dev returns the network fee wallet
+     *
+     * @return network fee wallet
+     */
+    function networkFeeWallet() external view override returns (address) {
```
lbeder, line 38: Perhaps `feeParams` should be renamed `networkFeeParams` so that it's consistent with the new getters?

yudilevi: makes sense

barakman: In that case, we may as well name the contract `NetworkFee`.

yudilevi: it's gonna hold other global network settings, so don't think so. but feeParams should be renamed.

yudilevi (changes requested):

solidity/contracts/NetworkSettings.sol
```diff
@@ -0,0 +1,84 @@
...
+        return _networkFeeWallet;
+    }
+
+    /**
+     * @dev returns the network fee
```
yudilevi, line 57: add a clause that describes what the fee is - fee that's taken as a portion from all swaps in the network

solidity/contracts/NetworkSettings.sol
```diff
@@ -0,0 +1,84 @@
...
+    }
+
+    /**
+     * @dev sets the network fee wallet
+     * can be executed only by the owner
+     *
+     * @param newNetworkFeeWallet new network fee wallet
+     */
+    function setNetworkFeeWallet(address newNetworkFeeWallet) external ownerOnly validAddress(newNetworkFeeWallet) {
+        _networkFeeWallet = newNetworkFeeWallet;
```
lbeder, line 72: Perhaps we should add events here?

solidity/contracts/NetworkSettings.sol
barakman: Added at the top of the file (since the term 'network fee' appears in several places).

solidity/contracts/NetworkSettings.sol
yudilevi: oh yeah, good call

barakman: Done.

yudilevi (approved):

PR #550, opened 2021-03-03
Fix test-deployment script

barakman:


solidity/utils/README.md
```diff
@@ -25,23 +25,23 @@ Here is an example of the initial configuration file which should be provided to
 {
     "reserves": [
         {
-            "type": 0,
             "symbol": "XXX",
+            "decimals": 18,
```
yudilevi, line 29: we probably shouldn't have the decimals in the config if the user can't change it... we should just deploy every token with 18 decimals and query the blockchain in the script to tell the decimals of each token.

yudilevi (changes requested):

solidity/utils/README.md
barakman: he can

yudilevi: gotcha

yudilevi (approved):

PR #557, opened 2021-03-08
Add transfer-fee function

barakman:


barakman:
No tests with network fee > 0% as of yet.

barakman:
> No tests with network fee > 0% as of yet.

Tests for conversions with network fee > 0% added.

Tests for liquidity functions (add, remove, get return, get cost) with network fee > 0% not yet added.

solidity/contracts/converter/types/standard-pool/StandardPoolConverter.sol
```diff
@@ -369,6 +371,48 @@ contract StandardPoolConverter is
...
     }
 
+    /**
+     * @dev returns the balance fee of a given reserve token
+     *
+     * @param _reserveToken    reserve token contract address
+     *
+     * @return the balance fee of the given reserve token
+     */
+    function reserveBalanceFee(IERC20 _reserveToken) public view returns (uint256) {
```
yudilevi, line 381: I don't see why this should be public

solidity/contracts/converter/types/standard-pool/StandardPoolConverter.sol
```diff
@@ -369,6 +371,48 @@ contract StandardPoolConverter is
...
+        require(reserveId != 0, "ERR_INVALID_RESERVE");
+        return reserveBalanceFee(reserveId);
+    }
+
+    /**
+     * @dev returns the balance fees of both reserve tokens
+     *
+     * @return the balance fees of both reserve tokens
+     */
+    function reserveBalanceFees() public view returns (uint256, uint256) {
```
yudilevi, line 392: I don't see why this should be public

solidity/contracts/converter/types/standard-pool/StandardPoolConverter.sol
```diff
@@ -369,6 +371,48 @@ contract StandardPoolConverter is
...
+        return reserveBalanceFees(1, 2);
+    }
+
+    /**
+     * @dev transfers a portion of the accumulated fees
+     */
+    function transferFees() public {
```
yudilevi, line 399: maybe just `processNetworkFees`?

solidity/contracts/converter/types/standard-pool/StandardPoolConverter.sol
```diff
@@ -986,6 +1080,23 @@ contract StandardPoolConverter is
         return reserveAmounts;
     }
 
+    function reserveBalancesMinusFees(IERC20[] memory _reserveTokens) internal view returns (uint256[2] memory) {
```
yudilevi, line 1083: fees need to be calculated based on the fee calculation formula, not by subtracting the previous balance from the current balance.
consider a case in which you have previous fee balances of 100,100 and after a swap, the balances are 200,60 - what would the fee be in this case?

yudilevi (changes requested):

solidity/contracts/converter/types/standard-pool/StandardPoolConverter.sol
barakman: @yudilevi : I added this function as part of (i.e., in favor of) adding tests for the fee before/after calling `transferFee`.
It is not even used in the contract (i.e., it is currently used **only** for testing).
We can move it to the corresponding helper contract, but the question is - why not keep it?
Just like the set of functions that we have for the reserve balances, don't we want one for the reserve balance fees?

solidity/contracts/converter/types/standard-pool/StandardPoolConverter.sol
barakman: Same as above.

solidity/contracts/converter/types/standard-pool/StandardPoolConverter.sol
barakman: @lbeder : suggestions/alternatives?

solidity/contracts/converter/types/standard-pool/StandardPoolConverter.sol
lbeder: If this is only used in tests, you should move it to a `TestStandardPoolConverter.sol` helper instead

solidity/contracts/converter/types/standard-pool/StandardPoolConverter.sol
lbeder: `processNetworkFees` isn't nad

solidity/contracts/converter/types/standard-pool/StandardPoolConverter.sol
barakman: This is true for pretty much every external function, as well as for several public functions.
The question is as I stated it above.

PR #559, opened 2021-03-09
Remove/consolidate `notThis` modifiers

lbeder:
In this PR, I have consolidated `validAddress` and `notThis` into the existing `validExternalAddress` modifiers. 

Please note, that in some of the cases I've removed `notThis` and kept only `validAddess` (e.g., when the wrong address can be changed back, when removing a previously verified address, etc.)

### DIFF
--- a/solidity/contracts/liquidity-protection/LiquidityProtectionSettings.sol
+++ b/solidity/contracts/liquidity-protection/LiquidityProtectionSettings.sol
@@ -152,8 +152,7 @@ contract LiquidityProtectionSettings is ILiquidityProtectionSettings, AccessCont
     constructor(IERC20 token, IContractRegistry registry)
         public
         ContractRegistryClient(registry)
-        validAddress(address(token))
-        notThis(address(token))
+        validExternalAddress(address(token))
     {
         // set up administrative roles.
         _setRoleAdmin(ROLE_OWNER, ROLE_OWNER);
@@ -289,12 +288,7 @@ contract LiquidityProtectionSettings is ILiquidityProtectionSettings, AccessCont
      *
      * @param poolAnchor    pool anchor
      */
-    function addPoolToWhitelist(IConverterAnchor poolAnchor)
-        external
-        onlyOwner
-        validAddress(address(poolAnchor))
-        notThis(address(poolAnchor))
-    {
+    function addPoolToWhitelist(IConverterAnchor poolAnchor) external onlyOwner validAddress(address(poolAnchor)) {
         require(_poolWhitelist.add(address(poolAnchor)), "ERR_POOL_ALREADY_WHITELISTED");
 
         emit PoolWhitelistUpdated(poolAnchor, true);
@@ -306,12 +300,7 @@ contract LiquidityProtectionSettings is ILiquidityProtectionSettings, AccessCont
      *
      * @param poolAnchor    pool anchor
      */
-    function removePoolFromWhitelist(IConverterAnchor poolAnchor)
-        external
-        onlyOwner
-        validAddress(address(poolAnchor))
-        notThis(address(poolAnchor))
-    {
+    function removePoolFromWhitelist(IConverterAnchor poolAnchor) external onlyOwner {
         require(_poolWhitelist.remove(address(poolAnchor)), "ERR_POOL_NOT_WHITELISTED");
 
         emit PoolWhitelistUpdated(poolAnchor, false);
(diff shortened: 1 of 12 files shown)


solidity/contracts/liquidity-protection/LiquidityProtectionSettings.sol
```diff
@@ -351,12 +339,7 @@ contract LiquidityProtectionSettings is ILiquidityProtectionSettings, AccessCont
      *
      * @param subscriber    subscriber address
      */
-    function removeSubscriber(ILiquidityProtectionEventsSubscriber subscriber)
-        external
-        onlyOwner
-        validAddress(address(subscriber))
-        notThis(address(subscriber))
-    {
+    function removeSubscriber(ILiquidityProtectionEventsSubscriber subscriber) external onlyOwner {
```
barakman, line 342: No need to check, because the function will revert anyway if `subscriber` does not point to a valid ILiquidityProtectionEventsSubscriber instance?

solidity/contracts/utility/TokenHolder.sol
```diff
@@ -33,11 +33,10 @@ contract TokenHolder is ITokenHolder, Owned, Utils {
         IERC20 _token,
         address _to,
         uint256 _amount
-    ) public virtual override ownerOnly validAddress(address(_token)) validAddress(_to) notThis(_to) {
+    ) public virtual override ownerOnly validAddress(address(_token)) validAddress(_to) {
```
barakman, line 36: Why bother checking `_token` when the function will revert anyway if it doesn't point to a valid `IERC20` instance?

barakman, line 36: Shouldn't we check `validExternalAddress(_to)`? I mean, if anything, then `_to` is likely to be the address of an EOA.

barakman (changes requested):

solidity/contracts/liquidity-protection/LiquidityProtectionSettings.sol
lbeder: Kind of. `_subscribers.remove` will revert with any address that wasn't previously verified via `addSubscriber`, which does check this.

solidity/contracts/utility/TokenHolder.sol
lbeder: I decided not to, since in the worst case, you could use `withdrawTokens` again

lbeder: I agree. @yudilevi, what do you think?

solidity/contracts/liquidity-protection/LiquidityProtectionSettings.sol
```diff
@@ -276,12 +275,7 @@ contract LiquidityProtectionSettings is ILiquidityProtectionSettings, AccessCont
      *
      * @param poolAnchor    pool anchor
      */
-    function addPoolToWhitelist(IConverterAnchor poolAnchor)
-        external
-        onlyOwner
-        validAddress(address(poolAnchor))
-        notThis(address(poolAnchor))
-    {
+    function addPoolToWhitelist(IConverterAnchor poolAnchor) external onlyOwner validAddress(address(poolAnchor)) {
```
yudilevi, line 278: shouldn't be `validExternalAddress`?

solidity/contracts/utility/TokenHolder.sol
yudilevi: sure, we can remove it. but on the other hand, can probably use `validExternalAddress` instead of `validAddress`

solidity/contracts/liquidity-protection/LiquidityProtectionSettings.sol
barakman: I think that the idea is to get rid of the 'notThis' check almost everywhere

solidity/contracts/utility/TokenHolder.sol
lbeder: This is what I've already done in the next PR. I didn't want to keep the `notThis` check because it's recoverable

solidity/contracts/liquidity-protection/LiquidityProtectionSettings.sol
lbeder: Yeah, we don't really need to do `validExternalAddress` here, because an incorrect setting can be reverted via the `remove` method

solidity/contracts/utility/TokenHolder.sol
yudilevi: cool

solidity/contracts/liquidity-protection/LiquidityProtectionSettings.sol
yudilevi: isn't the idea to prevent the setting from getting added in the first place?

lbeder: :+1:

yudilevi (approved):

barakman (approved):

PR #560, opened 2021-03-11
Transfer fees

barakman:


solidity/contracts/converter/types/standard-pool/StandardPoolConverter.sol
```diff
@@ -1154,6 +1212,14 @@ contract StandardPoolConverter is
         emit TokenRateUpdate(poolToken, _targetToken, _targetBalance, poolTokenSupply);
     }
 
+    function safeTransfer(IERC20 token, address to, uint256 amount) private {
```
lbeder, line 1215: @barakman I am moving this as part of my PR, but please note that this function (and some functions below) should include natspec.

barakman: Done.

solidity/contracts/converter/types/standard-pool/StandardPoolConverter.sol
```diff
@@ -1285,6 +1360,47 @@ contract StandardPoolConverter is
...
+     * @return numerator of the portion that should be taken from the balances of this pool
+     * @return denominator of the portion that should be taken from the balances of this pool
+     */
+    function networkFeeRatio(
+        uint256 currPoint,
+        uint256 prevPoint,
+        uint256 networkFee
+    ) private pure returns (uint256, uint256) {
+        return (
+            (currPoint - prevPoint) * networkFee,
```
lbeder, line 1399: Can't this overflow as well?

solidity/contracts/converter/types/standard-pool/StandardPoolConverter.sol
```diff
@@ -1285,6 +1360,47 @@ contract StandardPoolConverter is
...
+     * @param networkFee    the relative portion (in PPM units) that
+     * should be taken from the conversion fees accumulated in this pool
+     *
+     * @return numerator of the portion that should be taken from the balances of this pool
+     * @return denominator of the portion that should be taken from the balances of this pool
+     */
+    function networkFeeRatio(
+        uint256 currPoint,
+        uint256 prevPoint,
+        uint256 networkFee
```
lbeder, line 1396: Why isn't `networkFee` a `uint32`, like we usually do with `PPM` stuff?

solidity/contracts/converter/types/standard-pool/StandardPoolConverter.sol
```diff
@@ -1285,6 +1360,47 @@ contract StandardPoolConverter is
...
+    ) private pure returns (uint256, uint256) {
+        return (
+            (currPoint - prevPoint) * networkFee,
+            (currPoint * (PPM_RESOLUTION - networkFee)).add(prevPoint * networkFee * 2)
```
lbeder, line 1400: Can you elaborate on this calculation? Why are we taking `1 - networkFee`? Multiplying by 2?

solidity/test/BancorNetwork.js
```diff
@@ -22,9 +22,13 @@ const ConverterV28OrHigherWithoutFallback = contract.fromArtifact('ConverterV28O
 const ConverterV28OrHigherWithFallback = contract.fromArtifact('ConverterV28OrHigherWithFallback');
 
 const LiquidityPoolV1Converter = contract.fromArtifact('LiquidityPoolV1Converter');
+const NetworkSettings = contract.fromArtifact('NetworkSettings');
 
 const DSToken = contract.fromArtifact('DSToken');
 
+const NETWORK_FEE_WALLET = '0x'.padEnd(42, '1');
```
lbeder, line 29: Please use a "real" address here or even better - an instance of `TokenHolder` (following my previous PR). In any case, the network fee isn't a constant - it's a "real" address, so using a constant makes the tests confusing to real

solidity/contracts/converter/types/standard-pool/StandardPoolConverter.sol
```diff
@@ -8,6 +8,7 @@ import "../../ConverterVersion.sol";
 import "../../interfaces/IConverter.sol";
 import "../../interfaces/IConverterAnchor.sol";
 import "../../interfaces/IConverterUpgrader.sol";
+import "../../../INetworkSettings.sol";
```
lbeder, line 11: IMHO, you should also update `ConverterVersion`, so that it'd be easier for us to keep track on opted-in/upgraded converters with this feature

solidity/contracts/converter/types/standard-pool/StandardPoolConverter.sol
```diff
@@ -37,6 +38,7 @@ contract StandardPoolConverter is
     uint256 private constant AVERAGE_RATE_PERIOD = 10 minutes;
 
     uint256 private __reserveBalances;
+    uint256 private __reserveBalancesProd;
```
lbeder, line 41: AFAIK, this should've been added to `ConverterBase` as well. @yudilevi?

solidity/test/StandardPoolConverter.js
```diff
@@ -1126,13 +1134,13 @@ describe('StandardPoolConverter', () => {
 
                 for (let i = 0; i < allowances.length; i++) {
                     const diff = Decimal(allowances[i].toString()).div(reserveAmounts[i].toString());
-                    expect(diff.eq('0')).to.be.true();
+                    expect(diff.toFixed()).to.be.equal('0');
```
lbeder, line 1137: In general, we should really avoid mixing "numbers" and strings. In most of the cases (e.g., when precision isn't an issue), it's unneeded and complicates the code.

solidity/test/StandardPoolConverter.js
```diff
@@ -1188,4 +1196,212 @@ describe('StandardPoolConverter', () => {
...
+
+    describe('verifies that the network fee is transferred correctly via', () => {
+        const ONE_TOKEN = new BN(10).pow(new BN(18));
+        const TOTAL_SUPPLY = ONE_TOKEN.muln(1000000);
+        const CONVERSION_AMOUNT = ONE_TOKEN.muln(100);
+
+        for (const initialBalance1 of [100000, 200000, 400000, 800000]) {
+            for (const initialBalance2 of [100000, 300000, 500000, 700000]) {
+                for (const conversionFeePercent of [0, 5, 10, 25, 75]) {
+                    for (const networkFeePercent of [0, 5, 10, 25, 75]) {
```
lbeder, line 1208: Can you also check 100%, just to be on the safe side? It's as unrealistic as 0%, but can uncover potentially dangerous corner case (which can be maliciously triggered by the owner of the converter)

solidity/test/StandardPoolConverter.js
```diff
@@ -1188,4 +1196,212 @@ describe('StandardPoolConverter', () => {
...
+
+                            expectAlmostEqual(actualFee1, expectedFee1, '0', '0.001367');
+                            expectAlmostEqual(actualFee2, expectedFee2, '0', '0.001367');
+                        });
+                    }
+                }
+            }
+        }
+
+        function description(prefix, initialBalance1, initialBalance2, conversionFeePercent, networkFeePercent) {
```
lbeder, line 1338: In general, these are really good tests. Can you also add the following cases and make sure that the protocol behaves correctly:
1. Simulate multiple additions/removal in the same block.
2. Partial removal of liquidity.
3. Adding liquidity after a conversion.
4. Full liquidity removal, but in incremental steps (e.g., 50% and then the remainder).
5. Adding/removing very small amounts (e.g., weis) and see that the imprecision doesn't screw the system. There were similar exploits in the space already

solidity/utils/test_deployment.js
```diff
@@ -169,6 +169,10 @@ const run = async () => {
     const converterRegistryData = await web3Func(deploy, 'converterRegistryData', 'ConverterRegistryData', [
         contractRegistry._address
     ]);
+    const networkSettings = await web3Func(deploy, 'networkSettings', 'NetworkSettings', [
+        getConfig().networkSettingsParams.networkFeeWallet,
+        percentageToPPM(getConfig().networkSettingsParams.networkFee)
+    ]);
```
lbeder, line 175: 💪

solidity/contracts/converter/types/standard-pool/StandardPoolConverter.sol
lbeder, line 11: General reminder: don't forget to format the code before the final submission

lbeder (changes requested):

solidity/contracts/converter/types/standard-pool/StandardPoolConverter.sol
yudilevi: I think it can wait - we're gonna restructure the inheritance model soon and we might decide to have the standard pool be the parent contract, we'll see.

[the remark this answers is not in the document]
barakman: Yeah, the square root of that, but I've changed my mind.

solidity/contracts/converter/types/standard-pool/StandardPoolConverter.sol
```diff
@@ -230,7 +232,7 @@ contract StandardPoolConverter is
         // verify the the converter has exactly two reserves
         require(reserveTokenCount() == 2, "ERR_INVALID_RESERVE_COUNT");
         anchor.acceptOwnership();
-        syncReserveBalances();
+        syncReserveBalances(0);
```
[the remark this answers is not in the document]
barakman: @lbeder : @yudilevi has opposed that in the past, so please convince him.

solidity/contracts/converter/types/standard-pool/StandardPoolConverter.sol
```diff
@@ -368,6 +370,86 @@ contract StandardPoolConverter is
         return reserveBalances(1, 2);
     }
 
+    /**
+     * @dev transfers a portion of the accumulated conversion fees
+     */
+    function transferFees() external protected {
```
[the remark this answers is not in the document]
barakman: Fixed.

solidity/contracts/converter/types/standard-pool/StandardPoolConverter.sol
```diff
@@ -368,6 +370,86 @@ contract StandardPoolConverter is
...
+        (uint256 reserveBalance0, uint256 reserveBalance1) = transferFeesAndSyncReserveBalances(0);
+        __reserveBalancesProd = reserveBalance0 * reserveBalance1;
+    }
+
+    /**
+     * @dev transfers a portion of the accumulated conversion fees and syncs the reserve balances
+     *
+     * @param _value    amount of ether to exclude from the ether reserve balance (if relevant)
```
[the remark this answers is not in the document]
barakman: Fixed.

solidity/contracts/converter/types/standard-pool/StandardPoolConverter.sol
```diff
@@ -368,6 +370,86 @@ contract StandardPoolConverter is
...
+        return (reserveBalance0, reserveBalance1);
+    }
+
+    /**
+     * @dev returns the reserve balances of the given reserve tokens minus their corresponding fees
+     *
+     * @param _reserveTokens    reserve tokens
+     * @return reserve balances minus their corresponding fees
+     */
+    function reserveBalancesMinusFees(IERC20[] memory _reserveTokens) internal view returns (uint256[2] memory) {
```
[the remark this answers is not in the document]
barakman: Fixed.

solidity/contracts/converter/types/standard-pool/StandardPoolConverter.sol
```diff
@@ -368,6 +370,86 @@ contract StandardPoolConverter is
...
+        }
+
+        return [reserveBalance0, reserveBalance1];
+    }
+
+    /**
+     * @dev returns the reserve balances of the given reserve token minus its corresponding fee
+     *
+     * @param _reserveToken reserve token
+     * @return reserve balance minus its corresponding fee
```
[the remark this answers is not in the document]
barakman: We actually don't have a consistent convention for that throughout our code, but fixed.

solidity/contracts/converter/types/standard-pool/StandardPoolConverter.sol
```diff
@@ -368,6 +370,86 @@ contract StandardPoolConverter is
...
+     * @dev transfers a portion of the accumulated conversion fees and syncs the reserve balances
+     *
+     * @param _value    amount of ether to exclude from the ether reserve balance (if relevant)
+     * @return new reserve balances
+     */
+    function transferFeesAndSyncReserveBalances(uint256 _value) internal returns (uint256, uint256) {
+        syncReserveBalances(_value);
+
+        (uint256 reserveBalance0, uint256 reserveBalance1) = reserveBalances(1, 2);
+        (uint256 currPoint, uint256 prevPoint) = networkFeePoints(reserveBalance0 * reserveBalance1, __reserveBalancesProd);
```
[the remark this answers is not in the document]
barakman: No, they are 128-bit each.

solidity/contracts/converter/types/standard-pool/StandardPoolConverter.sol
```diff
@@ -797,12 +862,11 @@ contract StandardPoolConverter is
...
 
-        // sync the balances to ensure no mismatch
-        syncReserveBalances(msg.value);
-
         uint256[2] memory oldReserveBalances;
         uint256[2] memory newReserveBalances;
-        (oldReserveBalances[0], oldReserveBalances[1]) = reserveBalances();
+
+        // transfer the fees and sync the balances to ensure no mismatch
+        (oldReserveBalances[0], oldReserveBalances[1]) = transferFeesAndSyncReserveBalances(msg.value);
```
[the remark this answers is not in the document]
barakman: See my response to that note.

solidity/contracts/converter/types/standard-pool/StandardPoolConverter.sol
```diff
@@ -1153,6 +1211,23 @@ contract StandardPoolConverter is
...
 
+    /**
+     * @dev transfers tokens or ether from this contract to a given recipient
+     *
+     * @param token     address of the token
+     * @param to        address of the recipient
+     * @param amount    amount of tokens or ether
+     */
+    function safeTransfer(IERC20 token, address to, uint256 amount) private {
+        if (amount > 0) {
```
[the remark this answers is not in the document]
barakman: OK.

solidity/contracts/converter/types/standard-pool/StandardPoolConverter.sol
```diff
@@ -1049,7 +1112,8 @@ contract StandardPoolConverter is
         returns (uint256[] memory)
     {
         uint256 totalSupply = IDSToken(address(anchor)).totalSupply();
-        return removeLiquidityReserveAmounts(_amount, _reserveTokens, totalSupply);
+        uint256[2] memory _reserveBalances = reserveBalancesMinusFees(_reserveTokens);
```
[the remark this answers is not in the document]
barakman: It's used by `removeLiquidityReturn`, which is tested in the new use-cases added to this contract's test file.

solidity/contracts/converter/types/standard-pool/StandardPoolConverter.sol
```diff
@@ -1285,6 +1360,47 @@ contract StandardPoolConverter is
...
+     *
+     * @return the square root of the current product of the balances
+     * @return the square root of the previous product of the balances
+     */
+    function networkFeePoints(
+        uint256 currProd,
+        uint256 prevProd
+    ) private pure returns (uint256, uint256) {
+        return (
+            currProd > 0 ? MathEx.floorSqrt(currProd) : 0,
```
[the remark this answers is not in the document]
barakman: It allows to save gas when this case is not needed (which is mostly the case).

solidity/contracts/converter/types/standard-pool/StandardPoolConverter.sol
barakman: No. Each of `currPoint` and `prevPoint` is a square root of some value, hence at most 128 bits.
Hence their difference is at most 128 bits.
And since `networkFee` is at most 32 bits, the product above is at most 160 bits.

solidity/contracts/converter/types/standard-pool/StandardPoolConverter.sol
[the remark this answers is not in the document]
barakman: Unfixed; This requires changing many other functions in the file. I will open a separate task for it.

solidity/test/StandardPoolConverter.js
barakman: Unfortunately, mocha doesn't have any builtin methods for `Decimal` (the type of variable `diff`).
My fix here is intended to output a more descriptive error-message when an error is thrown (since "expected false to be true" isn't a very descriptive error-message).

solidity/contracts/converter/types/standard-pool/StandardPoolConverter.sol
```diff
@@ -368,6 +370,86 @@ contract StandardPoolConverter is
...
+     * @param _value    amount of ether to exclude from the ether reserve balance (if relevant)
+     * @return new reserve balances
+     */
+    function transferFeesAndSyncReserveBalances(uint256 _value) internal returns (uint256, uint256) {
+        syncReserveBalances(_value);
+
+        (uint256 reserveBalance0, uint256 reserveBalance1) = reserveBalances(1, 2);
+        (uint256 currPoint, uint256 prevPoint) = networkFeePoints(reserveBalance0 * reserveBalance1, __reserveBalancesProd);
+
+        if (currPoint > prevPoint) {
```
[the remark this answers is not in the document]
barakman: And return the exact same thing? What's the point?

solidity/contracts/converter/types/standard-pool/StandardPoolConverter.sol
```diff
@@ -368,6 +370,86 @@ contract StandardPoolConverter is
...
+     * @return reserve balance minus its corresponding fee
+     */
+    function reserveBalanceMinusFee(IERC20 _reserveToken) internal view returns (uint256) {
+        uint256 reserveId = __reserveIds[_reserveToken];
+        (uint256 thisReserveBalance, uint256 otherReserveBalance) = reserveBalances(reserveId, 3 - reserveId);
+        (uint256 currPoint, uint256 prevPoint) = networkFeePoints(thisReserveBalance * otherReserveBalance, __reserveBalancesProd);
+
+        if (currPoint > prevPoint) {
+            uint32 networkFee = INetworkSettings(addressOf(NETWORK_SETTINGS)).networkFee();
+            (uint256 n, uint256 d) = networkFeeRatio(currPoint, prevPoint, networkFee);
```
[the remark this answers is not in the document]
barakman: It repeats 2 times, one of which will likely be removed after we fix the prototype of function `addLiquidityReturn`.
And even regardless of that, I don't think that these 2 lines of code repeating in 2 different places are "worth" putting into a separate function. I've done that for the code in private helper `networkFeeRatio`, which seemed "worth" it.

solidity/contracts/converter/types/standard-pool/StandardPoolConverter.sol
```diff
@@ -368,6 +370,86 @@ contract StandardPoolConverter is
...
+
+    /**
+     * @dev returns the reserve balances of the given reserve token minus its corresponding fee
+     *
+     * @param _reserveToken reserve token
+     * @return reserve balance minus its corresponding fee
+     */
+    function reserveBalanceMinusFee(IERC20 _reserveToken) internal view returns (uint256) {
+        uint256 reserveId = __reserveIds[_reserveToken];
+        (uint256 thisReserveBalance, uint256 otherReserveBalance) = reserveBalances(reserveId, 3 - reserveId);
```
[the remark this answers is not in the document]
barakman: I've used the same convention as in function `setReserveBalance`.

solidity/contracts/converter/types/standard-pool/StandardPoolConverter.sol
barakman: Fixed.

solidity/test/BancorNetwork.js
barakman: Fixed.

solidity/contracts/converter/types/standard-pool/StandardPoolConverter.sol
barakman: Done.

solidity/contracts/converter/types/standard-pool/StandardPoolConverter.sol
barakman: Uniswap's code:
```
uint numerator = totalSupply.mul(rootK.sub(rootKLast));
uint denominator = rootK.mul(5).add(rootKLast);
```

Translation Phase #1:
```
uint n = totalSupply * (currPoint - prevPoint);
uint d = currPoint * 5 + prevPoint;
```

Translation Phase #2:
```
uint n = totalSupply * (currPoint - prevPoint);
uint d = currPoint * (PPM_RESOLUTION / networkFee - 1) + prevPoint;
```

Translation Phase #3:
```
uint n = totalSupply * (currPoint - prevPoint);
uint d = currPoint * (PPM_RESOLUTION - networkFee) / networkFee + prevPoint;
```

Translation Phase #4:
```
uint n = totalSupply * (currPoint - prevPoint);
uint d = (currPoint * (PPM_RESOLUTION - networkFee) + prevPoint * networkFee) / networkFee;
```

Translation Phase #5:
```
uint n = totalSupply * (currPoint - prevPoint) * networkFee;
uint d = currPoint * (PPM_RESOLUTION - networkFee) + prevPoint * networkFee;
```

This gives the fee in units of the pool tokens.
We need to convert half of it to units of reserve token 1, and half of it to units of reserve token 2.

In order to represent half, we multiply the denominator by 2:
```
uint d = currPoint * (PPM_RESOLUTION - networkFee) + prevPoint * networkFee * 2;
```

In order to represent by units of reserve token X, we:
1. Multiply by the balance of the pool in reserve token X
2. Divide by the total supply of pool tokens

**Important Note:**
Note that this relies on spot rate.
In order to rely on the bonding-curve rate, we would need to convert this amount using `removeLiquidityReturn`.
We may want to dwell on this in order to determine which one of these two options we should be using.

In any case, multiplying by the balance of the pool in reserve token X is conducted outside of this function.
In this function, we only divide by the total supply of pool tokens, so this:
```
uint n = totalSupply * (currPoint - prevPoint) * networkFee;
```
Changes to this:
```
uint n = (currPoint - prevPoint) * networkFee;
```

solidity/contracts/converter/types/standard-pool/StandardPoolConverter.sol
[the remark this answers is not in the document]
yudilevi: yeah I don't like the implicit side effect of using msg.value in an internal function.
it can cause problems when the function is called from different flows.

solidity/contracts/converter/types/standard-pool/StandardPoolConverter.sol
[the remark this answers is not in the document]
yudilevi: will give it some thought in my review...

solidity/test/StandardPoolConverter.js
barakman: Done.

barakman: BTW, it increases the duration of this already-long test by 20% (checking 6 different network fees instead of 5).

solidity/contracts/converter/types/standard-pool/StandardPoolConverter.sol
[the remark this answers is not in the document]
lbeder: It reduces nested logic, making the code easier to read and review.

solidity/test/StandardPoolConverter.js
lbeder: My note is regarding `'0'`.

solidity/test/StandardPoolConverter.js
lbeder: That's a small price to pay

solidity/contracts/converter/types/standard-pool/StandardPoolConverter.sol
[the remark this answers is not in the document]
lbeder: The only reason that we don't a consistent convention is since we keep postponing standardizing legacy code, but new or modified/touch code should be written according to our new conventions

solidity/test/BancorNetwork.js
```diff
@@ -157,7 +154,7 @@ describe('BancorNetwork', () => {
         const converterFactory = await ConverterFactory.new();
         await contractRegistry.registerAddress(registry.CONVERTER_FACTORY, converterFactory.address);
 
-        const networkSettings = await NetworkSettings.new(NETWORK_FEE_WALLET, NETWORK_FEE);
+        const networkSettings = await NetworkSettings.new(defaultSender, 0);
```
lbeder, line 157: It's not what I had in mind, but that's fine for now. We will change it soon anyway.

solidity/test/StandardPoolConverter.js
```diff
@@ -135,7 +133,7 @@ describe('StandardPoolConverter', () => {
         const factory = await ConverterFactory.new();
         await contractRegistry.registerAddress(registry.CONVERTER_FACTORY, factory.address);
 
-        networkSettings = await NetworkSettings.new(NETWORK_FEE_WALLET, NETWORK_FEE);
+        networkSettings = await NetworkSettings.new(defaultSender, 0);
```
lbeder: Should be `networkFeeWallet`

solidity/contracts/converter/ConverterVersion.sol
```diff
@@ -2,5 +2,5 @@
 pragma solidity 0.6.12;
 
 contract ConverterVersion {
-    uint16 public constant version = 45;
+    uint16 public constant version = 46;
```
lbeder, line 5: Keep in mind that the other PR is also using `46`, but since neither has been deployed yet - it's fine.

solidity/test/StandardPoolConverter.js
```diff
@@ -1203,7 +1203,7 @@ describe('StandardPoolConverter', () => {
         for (const initialBalance1 of [100000, 200000, 400000, 800000]) {
             for (const initialBalance2 of [100000, 300000, 500000, 700000]) {
                 for (const conversionFeePercent of [0, 5, 10, 25, 75]) {
```
lbeder, line 1205: I'd consider adding a similar case for a 100% `conversionFeePercent` (realizing that it'll slow the test, but that's life)

barakman: Cannot; that yields division by zero.

solidity/test/StandardPoolConverter.js
barakman: No it shouldn't. There is no purpose to test the network-fee here.
This is just in order to make the rest of the network operable, as we need a viable `NetworkSettings` contract, and the constructor of that contract takes a valid wallet as input).
Where the network-fee tests are relevant, we use an actual network-fee wallet.

solidity/test/BancorNetwork.js
barakman: There is no purpose to test the network-fee here, just make the system operable.

solidity/test/StandardPoolConverter.js
barakman: Done.

solidity/test/StandardPoolConverter.js
barakman: So is mine.
Variable `diff` is of type `Decimal`.
We can either do the above, or we can do:
```
expect(diff.eq('0')).to.be.true();
```
Which is not any better, and in fact - worse, because it doesn't provide valuable information upon error.

solidity/contracts/converter/ConverterVersion.sol
barakman: OK.

solidity/test/StandardPoolConverter.js
lbeder: I'm not sure that I follow. Why wouldn't `expect(diff.toFixed()).to.be.equal(0);` work?

solidity/test/BancorNetwork.js
lbeder: Gotcha

solidity/test/StandardPoolConverter.js
lbeder: Wait, this can be actually a corner-case we should understand better. If the conversion fee is 100%, why wouldn't any network fee equivalent to 0%? This is something that an `if` should have solved

solidity/test/StandardPoolConverter.js
barakman: `toFixed` returns a string, why compare it with a number?

solidity/test/StandardPoolConverter.js
barakman: I don't understand the question, but I will find out exactly what reverts and where (not related to the network-fee mechanism as far as I remember), and will let you know.

barakman: If the conversion fee is 100%, then the returned (target) amount is obviously zero, which reverts at function `doConvert`:
```
// ensure that the trade gives something in return
require(amount != 0, "ERR_ZERO_TARGET_AMOUNT");
```

barakman: BTW, no similar problem with the network fee (conceptual or technical), since it represents a portion of the conversion fee, not a portion of the total amount.

solidity/contracts/converter/types/standard-pool/StandardPoolConverter.sol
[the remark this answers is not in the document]
barakman: Done (unnecessarily imo).

solidity/contracts/converter/types/standard-pool/StandardPoolConverter.sol
barakman: Any idea why scoping a block of code with `/* eslint-disable max-len */` and `/* eslint-enable max-len */` doesn't do what it's supposed to do?

solidity/contracts/converter/types/standard-pool/StandardPoolConverter.sol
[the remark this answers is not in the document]
barakman: @lbeder : from [the official docimentation](https://docs.soliditylang.org/en/v0.5.0/types.html?highlight=calldata#data-location):
> Calldata is only valid for parameters of external contract functions and is required for this type of parameter. Calldata is a non-modifiable, non-persistent area where function arguments are stored, and behaves mostly like memory.

The function above is internal, so I don't think that `calldata` is even permitted in this case.

solidity/test/StandardPoolConverter.js
lbeder: Oh, you're right. Let's hope that switching to Hardhat will make these irregularities disappear.

solidity/contracts/converter/types/standard-pool/StandardPoolConverter.sol
lbeder: It might require `//` here, but in any case - try to disable the rule for a single line first via `// eslint-disable-next-line max-len`, so that it won't "infect" unrelated lines

solidity/test/StandardPoolConverter.js
lbeder: This is what I wanted us to understand and you did help with that. Thanks!

solidity/contracts/converter/types/standard-pool/StandardPoolConverter.sol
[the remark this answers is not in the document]
lbeder: You can use `calldata` for `internal` or `private` function as well. It will work well for functions that receive a propagated "read-only" arguments from `external` or `public` function, so should be applicable in some of these cases. In any case, keep this in mind during refactoring of existing/new code.

solidity/contracts/converter/types/standard-pool/StandardPoolConverter.sol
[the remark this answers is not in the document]
lbeder: I meant that we aren't using the "virtual balances" in swaps like we use in add/remove liquidity, but rather the actual balances, which might (?) cause the following issue:

1. Do a huge swap, heavily impacting the balance.
2. Try to remove liquidity and it reverts due to some possible issue.

Can you test this case as well?

solidity/contracts/converter/types/standard-pool/StandardPoolConverter.sol
lbeder: 👍 👍 👍

solidity/test/StandardPoolConverter.js
```diff
@@ -1333,6 +1333,87 @@ describe('StandardPoolConverter', () => {
             }
         }
 
+        it('all functions', async () => {
```
lbeder: Perhaps `it('should properly handle complex conversions and liquidity provision scenarios')`?

barakman: This is weird; I have changed this description long before you posted this comment.

lbeder: Ah, I've probably commented on an outdated commit

lbeder (approved):

solidity/contracts/converter/types/standard-pool/StandardPoolConverter.sol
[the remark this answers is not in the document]
barakman: Done.

solidity/test/StandardPoolConverter.js
```diff
@@ -1188,4 +1194,374 @@ describe('StandardPoolConverter', () => {
...
+                            await addLiquidity(reserveToken1, reserveToken2, converter, [initialBalance1, initialBalance2].map(n => ONE_TOKEN.muln(n)));
+
+                            const conversion = await convert(reserveToken1, poolToken, reserveToken2, bancorNetwork, converter, CONVERSION_AMOUNT);
+                            const expectedFeeBase = conversion.fee.divn(2).muln(networkFeePercent).divn(100 + networkFeePercent);
+                            const reserveBalance1 = ONE_TOKEN.muln(initialBalance1).add(CONVERSION_AMOUNT);
+                            const reserveBalance2 = ONE_TOKEN.muln(initialBalance2).sub(conversion.amount);
+
+                            await converter.processNetworkFees();
+
+                            const balance1After = await reserveToken1.balanceOf(networkFeeWallet);
```
sshmaxime, line 1218: Seems like `balance1After` and `balance2After` aren't used, here and below.

barakman: Fixed.

solidity/contracts/converter/types/standard-pool/StandardPoolConverter.sol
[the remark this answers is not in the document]
lbeder: @barakman The new test doesn't test removeLiquidity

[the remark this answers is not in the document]
barakman: done

solidity/contracts/converter/types/standard-pool/StandardPoolConverter.sol
```diff
@@ -29,6 +30,7 @@ contract StandardPoolConverter is ConverterVersion, IConverter, ContractRegistry
     uint256 private constant AVERAGE_RATE_PERIOD = 10 minutes;
 
     uint256 private __reserveBalances;
+    uint256 private __reserveBalancesProd;
```
yudilevi, line 33: __reserveBalancesProd -> _reserveBalancesProd

solidity/contracts/converter/types/standard-pool/StandardPoolConverter.sol
```diff
@@ -342,6 +344,91 @@ contract StandardPoolConverter is ConverterVersion, IConverter, ContractRegistry
         return reserveBalances(1, 2);
     }
 
+    /**
+     * @dev transfers a portion of the accumulated conversion fees
+     */
+    function processNetworkFees() external protected {
```
yudilevi, line 350: why not receive a _syncReserves flag instead of 2 separate functions?

solidity/contracts/converter/types/standard-pool/StandardPoolConverter.sol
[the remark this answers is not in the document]
yudilevi: `baseReserveBalances`?

solidity/contracts/converter/types/standard-pool/StandardPoolConverter.sol
```diff
@@ -342,6 +344,91 @@ contract StandardPoolConverter is ConverterVersion, IConverter, ContractRegistry
...
+     * @dev returns the reserve balances of the given reserve tokens minus their corresponding fees
+     *
+     * @param _reserveTokens reserve tokens
+     *
+     * @return reserve balances minus their corresponding fees
+     */
+    function reserveBalancesMinusFees(IERC20[] memory _reserveTokens) internal view returns (uint256[2] memory) {
+        uint256 reserveId0 = __reserveIds[_reserveTokens[0]];
+        uint256 reserveId1 = __reserveIds[_reserveTokens[1]];
+        (uint256 reserveBalance0, uint256 reserveBalance1) = reserveBalances(reserveId0, reserveId1);
```
yudilevi, line 396: seems like there's a bunch of duplicated code between this function and `processNetworkFeesAndSyncReserveBalances`

solidity/contracts/converter/types/standard-pool/StandardPoolConverter.sol
```diff
@@ -825,6 +896,9 @@ contract StandardPoolConverter is ConverterVersion, IConverter, ContractRegistry
         // set the reserve balances
         setReserveBalances(1, 2, newReserveBalances[0], newReserveBalances[1]);
 
+        // set the reserve balances product
+        __reserveBalancesProd = newReserveBalances[0] * newReserveBalances[1];
```
yudilevi, line 900: saw this in multiple places, probably worth moving to a utility function

solidity/contracts/converter/types/standard-pool/StandardPoolConverter.sol
```diff
@@ -1259,6 +1324,47 @@ contract StandardPoolConverter is ConverterVersion, IConverter, ContractRegistry
...
+    /**
+     * @dev returns two samples required for calculating the network fee
+     *
+     * @param currProd  the current product of the balances
+     * @param prevProd  the previous product of the balances
+     *
+     * @return the square root of the current product of the balances
+     * @return the square root of the previous product of the balances
+     */
+    function networkFeePoints(
```
yudilevi, line 1336: I'd rather have a more self explanatory name here

solidity/contracts/converter/types/standard-pool/StandardPoolConverter.sol
```diff
@@ -1259,6 +1324,47 @@ contract StandardPoolConverter is ConverterVersion, IConverter, ContractRegistry
...
+     *
+     * @param currPoint     the current product of the balances
+     * @param prevPoint     the previous product of the balances
+     * @param networkFee    the relative portion (in PPM units) that
+     * should be taken from the conversion fees accumulated in this pool
+     *
+     * @return numerator of the portion that should be taken from the balances of this pool
+     * @return denominator of the portion that should be taken from the balances of this pool
+     */
+    function networkFeeRatio(
```
yudilevi, line 1357: it makes sense to have a utility function that returns the actual amounts.
I understand we have one case in which we only need one of the reserves, let's discuss that prototype.

yudilevi (changes requested):

solidity/contracts/converter/types/standard-pool/StandardPoolConverter.sol

solidity/contracts/converter/types/standard-pool/StandardPoolConverter.sol
barakman: The internal function (`processNetworkFeesAndSyncReserveBalances`) is called from a few other places, and the update of the state variable which occurs in this function, occurs at a different point in time in each one of these places (i.e., not immediately after the internal function is called).

solidity/contracts/converter/types/standard-pool/StandardPoolConverter.sol
barakman: It's a 1-liner (which appears in 2 places).

solidity/contracts/converter/types/standard-pool/StandardPoolConverter.sol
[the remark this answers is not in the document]

solidity/contracts/converter/types/standard-pool/StandardPoolConverter.sol
barakman: please suggest

solidity/contracts/converter/types/standard-pool/StandardPoolConverter.sol
barakman: I've done (before this review) some work of extracting common code into dedicated functions, like `networkFeePoints` and `networkFeeRatio`. But the rest ofd the code in these functions is logically different, and making a common function in this case would yield higher gas cost.

solidity/contracts/converter/types/standard-pool/StandardPoolConverter.sol
barakman: We have (in their new names, as suggested by you):
A. Function `baseReserveBalance`
B. Function `baseReserveBalances`

Function A is used only in function `addLiquidityReturn`, which takes a single reserve token as input.

So it's going to be difficult to use function B here, since function `addLiquidityReturn` will need to determine what the other reserve token is.

Moreover, we need to revise function `addLiquidityReturn` to take all reserve tokens (and the corresponding amounts) as input, following which - we can get rid of function A altogether.

BTW, I'm not sure how this is related to function `networkFeeRatio` though.
Did you mean to post this comment elsewhere in the code?

barakman: BTW 2 - combine all of the above with our plan to end support for allowing users to pass input amounts in any order, which means that we will be able to get rid of input reserve tokens in all related functions.

solidity/contracts/converter/types/standard-pool/StandardPoolConverter.sol
yudilevi: let's discuss this

solidity/contracts/converter/types/standard-pool/StandardPoolConverter.sol
yudilevi: I don't quite understand the math step here in order to propose a new name

solidity/contracts/converter/types/standard-pool/StandardPoolConverter.sol
yudilevi: ok let's discuss the final option here

solidity/contracts/converter/types/standard-pool/StandardPoolConverter.sol
barakman: This function returns the geometric mean of each point in the network fee mechanism (the current point, and the previous point recorded, where 'point' stands for the status of the balances (i.e., the balance of the pool in each reserve).

The "problem" here is that this function doesn't receive the actual points (current and previous), but the products, i.e., the product of the current point and the product of the previous point.

So all this function does, really, is to return the square-roots of 2 values.
Unfortunately, there isn't a real good name for it.
Essentially, all this function does is to aggregate a piece of common code which appears in several different places.

The 'networkFee' prefix isn't event relevant here TBH.

We may as well just call it 'squareRoots' or something like that.

solidity/contracts/converter/types/standard-pool/StandardPoolConverter.sol
barakman: Renamed `processNetworkFeesAndSyncReserveBalances` to `processNetworkFees` (overloaded function).

solidity/contracts/converter/types/standard-pool/StandardPoolConverter.sol
barakman: Fixed in https://github.com/bancorprotocol/contracts-solidity/pull/568.

solidity/contracts/converter/types/standard-pool/StandardPoolConverter.sol
barakman: Removed in https://github.com/bancorprotocol/contracts-solidity/pull/568.

solidity/contracts/converter/types/standard-pool/StandardPoolConverter.sol
barakman: Removed in https://github.com/bancorprotocol/contracts-solidity/pull/568.

solidity/contracts/converter/types/standard-pool/StandardPoolConverter.sol
barakman: Done

solidity/contracts/converter/types/standard-pool/StandardPoolConverter.sol
barakman: Done

solidity/contracts/converter/types/standard-pool/StandardPoolConverter.sol
barakman: Done

solidity/contracts/converter/types/standard-pool/StandardPoolConverter.sol
```diff
@@ -343,17 +345,45 @@ contract StandardPoolConverter is ConverterVersion, IConverter, ContractRegistry
     }
 
     /**
-     * @dev returns the reserve balances of the given reserve tokens
+     * @dev transfers a portion of the accumulated conversion fees
+     */
+    function processNetworkFees() external protected {
```
yudilevi, line 350: add more descriptive comment, such as "calculates and transfers a portion of the fees to the network fee wallet"

solidity/contracts/converter/types/standard-pool/StandardPoolConverter.sol
```diff
@@ -343,17 +345,45 @@ contract StandardPoolConverter is ConverterVersion, IConverter, ContractRegistry
...
+        (uint256 reserveBalance0, uint256 reserveBalance1) = processNetworkFees(0);
+        _reserveBalancesProduct = reserveBalance0 * reserveBalance1;
+    }
+
+    /**
+     * @dev transfers a portion of the accumulated conversion fees
```
yudilevi, line 356: explain how this permutation is different than the public one

solidity/contracts/converter/types/standard-pool/StandardPoolConverter.sol
```diff
@@ -343,17 +345,45 @@ contract StandardPoolConverter is ConverterVersion, IConverter, ContractRegistry
...
+     *
+     * @param _value amount of ether to exclude from the ether reserve balance (if relevant)
+     *
+     * @return new reserve balances
+     */
+    function processNetworkFees(uint256 _value) internal returns (uint256, uint256) {
+        syncReserveBalances(_value);
+        (uint256 reserveBalance0, uint256 reserveBalance1) = reserveBalances(1, 2);
+        (address wallet, uint256 fee0, uint256 fee1) = networkWalletAndFees(reserveBalance0, reserveBalance1);
+        reserveBalance0 -= fee0;
```
yudilevi, line 366: I'd add underflow protection here just in case

solidity/contracts/converter/types/standard-pool/StandardPoolConverter.sol
```diff
@@ -749,12 +779,11 @@ contract StandardPoolConverter is ConverterVersion, IConverter, ContractRegistry
...
         uint256 totalSupply = poolToken.totalSupply();
 
-        // sync the balances to ensure no mismatch
-        syncReserveBalances(msg.value);
-
         uint256[2] memory oldReserveBalances;
         uint256[2] memory newReserveBalances;
-        (oldReserveBalances[0], oldReserveBalances[1]) = reserveBalances();
+
+        // transfer the fees and sync the balances to ensure no mismatch
```
yudilevi, line 785: transfer -> process?

solidity/contracts/converter/types/standard-pool/StandardPoolConverter.sol
```diff
@@ -749,12 +779,11 @@ contract StandardPoolConverter is ConverterVersion, IConverter, ContractRegistry
...
+        (oldReserveBalances[0], oldReserveBalances[1]) = processNetworkFees(msg.value);
```
yudilevi, line 786: old -> prev?

solidity/contracts/converter/types/standard-pool/StandardPoolConverter.sol
```diff
@@ -863,14 +895,15 @@ contract StandardPoolConverter is ConverterVersion, IConverter, ContractRegistry
...
         uint256 newPoolTokenSupply = totalSupply.sub(_amount);
 
         uint256[2] memory oldReserveBalances;
         uint256[2] memory newReserveBalances;
-        (oldReserveBalances[0], oldReserveBalances[1]) = reserveBalances();
+
+        // transfer the fees and sync the balances to ensure no mismatch
+        (oldReserveBalances[0], oldReserveBalances[1]) = processNetworkFees(0);
+
+        uint256[] memory reserveAmounts = removeLiquidityReserveAmounts(_amount, totalSupply, oldReserveBalances);
```
yudilevi, line 906: duplicate line

yudilevi (changes requested):

solidity/contracts/converter/types/standard-pool/StandardPoolConverter.sol
barakman: Function `networkWalletAndFees` takes care of that.
If you find this insufficient, then we have the same problem in function `baseReserveBalances` (which also calls function `networkWalletAndFees`).

solidity/contracts/converter/types/standard-pool/StandardPoolConverter.sol
barakman: That's actually a "leftover" of the pull from 0.6.34, which yields a compilation error; fixed.

solidity/contracts/converter/types/standard-pool/StandardPoolConverter.sol
barakman: fixed

solidity/contracts/converter/types/standard-pool/StandardPoolConverter.sol
barakman: done

solidity/contracts/converter/types/standard-pool/StandardPoolConverter.sol
barakman: This will changed code which is so far untouched in this PR. Are you sure you want to proceed?

barakman:
> IMHO, we don't need to specify such a tight boundary here. 50% is a good enough definition, even if the actual value can be technically higher.

reverted.

barakman:
> If `networkFee` can be `PPM_RESOLUTION * 2 / 3` - it would underflow, right?

reverted.

solidity/contracts/converter/types/standard-pool/StandardPoolConverter.sol

solidity/test/StandardPoolConverter.js
```diff
@@ -1210,4 +1148,394 @@ describe('StandardPoolConverter', () => {
...
+            async function createPool(networkFeePercent, conversionFeePercent, ethIndex) {
+                const poolToken = await DSToken.new('poolToken', 'poolToken', 18);
+                const reserveToken1 = await TestStandardToken.new('reserveToken1', 'reserveToken1', 18, TOTAL_SUPPLY);
+                const reserveToken2 = await TestStandardToken.new('reserveToken2', 'reserveToken2', 18, TOTAL_SUPPLY);
+                const converter = await StandardPoolConverter.new(poolToken.address, contractRegistry.address, 1000000);
+
+                const reserveToken = [{}, reserveToken1, reserveToken2][ethIndex];
+                reserveToken.address = NATIVE_TOKEN_ADDRESS;
+                reserveToken.approve = async (spender, value) => {};
+                reserveToken.balanceOf = async (account) => new BN(await web3.eth.getBalance(account));
```
sshmaxime, line 1482: I don't understand the purpose of `reserveToken` here. It is never used.

lbeder: @sshmaxime please ignore this. I've already rewritten lots of this code in an upcoming PR.

barakman: It points to `reserveToken1` when ethIndex == 1`, to `reserveToken2` when `ethIndex == 2`, and to an unused object when `ethToken == 0`.

yudilevi (approved):

PR #561, opened 2021-03-13
TokenHolder extraction and upgrade

lbeder:
- Remove `TokenHolder` from `ConverterBase` and `StandardPoolConverter`.
- Increase `ConverterVersion` and update the `ConverterUpgrader` accordingly .
- Add `payable` `receive()` callback to `TokenHolder`.
- Add `withdrawMultipleTokens` to `TokenHolder`.
- Expose `TokenHolder`'s `safeTransfer` as **internal** helper function.
- Disallow external removal of non-reserve tokens from converters.
- Note: LP's `wallet` has to be `payable`.


solidity/contracts/utility/TokenHolder.sol
```diff
@@ -13,32 +13,66 @@ import "./interfaces/ITokenHolder.sol";
...
+
+    /**
+     * @dev withdraws multiple funds held by the contract and sends them to an account
+     * can only be called by the owner
+     *
+     * @param tokens ERC20 token contract addresses (with a special handling of NATIVE_TOKEN_ADDRESS)
+     * @param to account to receive the new amount
+     * @param amounts amounts to withdraw
+     */
+    function withdrawMultipleTokens(
```
yudilevi, line 47: this allows the owner to withdraw reserves, overriding the protections in the converter.
I think we should remove token holder as a base contract for the converters and just add a custom function there for upgrading that will also send both (all?) reserve balances to a new address in a single call.

solidity/contracts/utility/interfaces/ITokenHolder.sol
```diff
@@ -9,9 +9,17 @@ import "./IOwned.sol";
...
     function withdrawTokens(
-        IERC20 _token,
-        address _to,
-        uint256 _amount
+        IERC20 token,
+        address to,
+        uint256 amount
+    ) external;
+
+    function withdrawMultipleTokens(
```
yudilevi, line 20: maybe `withdrawTokensMultiple` is better in this case? not sure... :)

yudilevi (changes requested):

solidity/contracts/utility/interfaces/ITokenHolder.sol
lbeder: 👍

barakman (changes requested):

solidity/contracts/converter/ConverterBase.sol
```diff
@@ -231,26 +213,51 @@ abstract contract ConverterBase is ConverterVersion, IConverter, TokenHolder, Co
...
+        // verify that the converter is inactive or that the owner is the upgrader contract
+        require(owner == addressOf(CONVERTER_UPGRADER), "ERR_ACCESS_DENIED");
+
+        uint256 reserveCount = reserveTokens.length;
+        for (uint256 i = 0; i < reserveCount; ++i) {
+            IERC20 reserveToken = reserveTokens[i];
+
+            uint256 amount;
+            if (reserveToken == NATIVE_TOKEN_ADDRESS) {
+                amount = address(this).balance;
```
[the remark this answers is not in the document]
lbeder: I don't think that it's required here, since all tests are passing. In the previous issue, it was mostly a ganache issue and not an EVM issue.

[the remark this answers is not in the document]
barakman: It occurred when forking mainnet on ganache and then try to execute on a contract in the forked part, so you are probably right.

solidity/contracts/utility/TokenHolder.sol
lbeder: Done and done

solidity/contracts/utility/interfaces/ITokenHolder.sol
lbeder: Done

solidity/contracts/converter/ConverterBase.sol
```diff
@@ -231,26 +213,51 @@ abstract contract ConverterBase is ConverterVersion, IConverter, TokenHolder, Co
...
 
     /**
-     * @dev withdraws tokens held by the converter and sends them to an account
+     * @dev transfers reserves to a new converter during an upgrade
+     * can only be called by the converter upgraded which should be set at its owner
+     *
+     * @param _newConverter account to receive the new amount
+     */
+    function upgradeReserves(address _newConverter) external override protected ownerOnly {
+        // verify that the converter is inactive or that the owner is the upgrader contract
```
yudilevi, line 222: comment isn't relevant and we could potentially move the upgrader check to a modifier (not a must).

solidity/contracts/converter/ConverterBase.sol
```diff
@@ -231,26 +213,51 @@ abstract contract ConverterBase is ConverterVersion, IConverter, TokenHolder, Co
     }
 
     /**
-     * @dev withdraws tokens held by the converter and sends them to an account
+     * @dev transfers reserves to a new converter during an upgrade
+     * can only be called by the converter upgraded which should be set at its owner
+     *
+     * @param _newConverter account to receive the new amount
+     */
+    function upgradeReserves(address _newConverter) external override protected ownerOnly {
```
yudilevi, line 221: maybe rename to `transferReservesOnUpgrade` to make it more explicit? `upgradeReserves` is a bit off

solidity/contracts/converter/ConverterBase.sol
```diff
@@ -231,26 +213,51 @@ abstract contract ConverterBase is ConverterVersion, IConverter, TokenHolder, Co
...
      * unless the owner is the converter upgrader contract
      *
-     * @param _token   ERC20 token contract address
-     * @param _to      account to receive the new amount
+     * @param _token ERC20 token contract address
+     * @param _to account to receive the new amount
      * @param _amount  amount to withdraw
      */
-    function withdrawTokens(
+    function withdrawInactiveTokens(
```
yudilevi, line 252: I would completely remove the option to withdraw a non-reserve token, owners practically never use it and it just adds complexity to a sensitive code.

solidity/contracts/converter/ConverterBase.sol
```diff
@@ -419,7 +426,9 @@ abstract contract ConverterBase is ConverterVersion, IConverter, TokenHolder, Co
      */
     function syncReserveBalances() internal {
```
yudilevi, line 427: I'd like to make that function public (can be a separate PR) but we need to be 100% sure it has no side effects.
It will allow supporting rebase tokens and some are waiting for a while now.

solidity/contracts/converter/ConverterUpgrader.sol
```diff
@@ -5,6 +5,16 @@ import "./interfaces/IConverterUpgrader.sol";
 import "./interfaces/IConverterFactory.sol";
 import "../utility/ContractRegistryClient.sol";
 
+interface IConverterWithTokenHolder is IConverter {
```
yudilevi, line 8: can just rename to ILegacyConverter in case we need to add more interfaces there.

solidity/contracts/converter/types/standard-pool/StandardPoolConverter.sol
```diff
@@ -247,27 +220,53 @@ contract StandardPoolConverter is
     }
 
     /**
-     * @dev withdraws tokens held by the converter and sends them to an account
+     * @dev transfers reserves to a new converter during an upgrade
+     * can only be called by the converter upgraded which should be set at its owner
+     *
+     * @param _newConverter account to receive the new amount
+     */
+    function upgradeReserves(address _newConverter) external override protected ownerOnly {
```
yudilevi, line 228: same comment as the one in the ConverterBase

solidity/contracts/converter/types/standard-pool/StandardPoolConverter.sol
```diff
@@ -247,27 +220,53 @@ contract StandardPoolConverter is
...
      *
-     * @param _token   ERC20 token contract address
-     * @param _to      account to receive the new amount
-     * @param _amount  amount to withdraw
+     * @param _token ERC20 token contract address
+     * @param _to account to receive the new amount
+     * @param _amount amount to withdraw
      */
-    function withdrawTokens(
+    function withdrawInactiveTokens(
```
yudilevi, line 259: same comment as the one in the ConverterBase

solidity/contracts/utility/TokenHolder.sol
```diff
@@ -13,32 +13,66 @@ import "./interfaces/ITokenHolder.sol";
...
-        IERC20 _token,
-        address _to,
-        uint256 _amount
-    ) public virtual override ownerOnly validAddress(address(_token)) validAddress(_to) notThis(_to) {
-        if (_token == NATIVE_TOKEN_ADDRESS) {
-            payable(_to).transfer(_amount);
+        IERC20 token,
+        address to,
+        uint256 amount
+    ) public virtual override ownerOnly validAddress(to) {
```
yudilevi, line 35: can change to `validExternalAddress` once we merge the other PR

solidity/contracts/utility/TokenHolder.sol
```diff
@@ -13,32 +13,66 @@ import "./interfaces/ITokenHolder.sol";
...
+
+    /**
+     * @dev withdraws multiple funds held by the contract and sends them to an account
+     * can only be called by the owner
+     *
+     * @param tokens ERC20 token contract addresses (with a special handling of NATIVE_TOKEN_ADDRESS)
+     * @param to account to receive the new amount
+     * @param amounts amounts to withdraw
+     */
+    function withdrawTokensMultiple(
```
yudilevi, line 47: can change to `validExternalAddress` once we merge the other PR

solidity/test/ConverterUpgrader.js
```diff
@@ -40,7 +40,7 @@ describe('ConverterUpgrader', () => {
         const upgrader = await ConverterUpgrader.new(contractRegistry.address);
 
         await contractRegistry.registerAddress(registry.CONVERTER_UPGRADER, upgrader.address);
-        if (version) {
+        if (version && version < 45) {
```
yudilevi, line 43: why < 45 here? if any, probably <= 45, but I think that we can still use addConnector anyway as we still have backward compatibility for it

solidity/test/ConverterUpgrader.js
```diff
@@ -23,7 +23,7 @@ const RESERVE1_BALANCE = new BN(5000);
 const RESERVE2_BALANCE = new BN(8000);
 const TOKEN_TOTAL_SUPPLY = new BN(20000);
 
-const VERSIONS = [9, 10, 11, 23];
+const VERSIONS = [9, 10, 11, 23, 45];
```
yudilevi, line 26: I might be missing something but it's critical to have a test for 46 => 46 upgrade since that's the only version that tests the new `upgradeReserves` functionality (and I might not remember and we already do that by default...)

solidity/test/TokenHolder.js
```diff
@@ -1,56 +1,169 @@
...
+    describe('withdraw multiple assets', () => {
+        let tokenAddresses;
+        let amounts;
+
+        beforeEach(async () => {
+            tokenAddresses = [NATIVE_TOKEN_ADDRESS, token.address, token2.address];
+            amounts = {};
+
+            for (const tokenAddress of tokenAddresses) {
+                amounts[tokenAddress] = new BN(100);
```
yudilevi, line 113: we should also test transferring different amounts (amount * index?)

yudilevi (changes requested):

solidity/contracts/converter/ConverterBase.sol
lbeder: 👍

solidity/contracts/converter/ConverterBase.sol
lbeder: 👍

solidity/contracts/converter/ConverterBase.sol
lbeder: Roger 👍

solidity/contracts/converter/ConverterBase.sol
lbeder: I'd definitely think it over and only do it in a new PR. There were multiple exploits, in other protocols, which utilized a public `sync` function (including involving a front-running on an AMPL-like rebase event)

solidity/contracts/converter/ConverterUpgrader.sol
lbeder: 👍

solidity/contracts/utility/TokenHolder.sol
lbeder: Are you sure you want `validExternalAddress` instead of `validAddress`? In the previous PR, we have explicitly removed the need for the `notThis` check in this case.

solidity/test/ConverterUpgrader.js
lbeder: 1. We no longer have `addConnector` in neither `ConverterBase` or `StandardPoolConveter` :(
2. Since we don't have `addConnector` in version 45, I had to exclude it explicitly. The actual version it was removed from is <45, but since we only check 23 and 45 - excluding 45 works as well.

solidity/test/ConverterUpgrader.js
lbeder: Oh, we do check the `latest1 version in addition to these explicit versions. I'll rename this to `LEGACU_VERSIONS` to make it clearer

lbeder: See line 208: `describe(`${init.name}(version = ${version || 'latest'}, activate = ${activate}):`, () => {`

solidity/test/TokenHolder.js
lbeder: 👍

solidity/test/ConverterUpgrader.js
barakman: At least in the past, this test used to run on all versions explicitly specified, along with the latest version (which didn't need to be specified).

solidity/contracts/utility/TokenHolder.sol
lbeder: `notThis` was explicitly removed there as well, so in merging it, this should stay `validAddress` as well

solidity/contracts/converter/ConverterBase.sol
barakman: Questions that should have been asked regardless of this change:
1. Why should an owner who is not the converter upgrader have the capability of withdrawing?
2. Why do we need the input amount here? I assume that this function should always withdraw everything, whether it's upon converter upgrade or due to some other reason (for which the owner needs to withdraw).
3. Why do we need the input token here? I assume that this function should always withdraw everything, whether it's upon converter upgrade or due to some other reason (for which the owner needs to withdraw).

solidity/contracts/converter/ConverterBase.sol
```diff
@@ -231,31 +213,54 @@ abstract contract ConverterBase is ConverterVersion, IConverter, TokenHolder, Co
...
-        // otherwise verify that the converter is inactive or that the owner is the upgrader contract
-        require(!reserves[_token].isSet || !isActive() || owner == converterUpgrader, "ERR_ACCESS_DENIED");
-        super.withdrawTokens(_token, _to, _amount);
+        safeTransfer(_token, _to, _amount);
 
-        // if the token is a reserve token, sync the reserve balance
-        if (reserves[_token].isSet) {
-            syncReserveBalance(_token);
-        }
+        syncReserveBalance(_token);
```
barakman, line 263: If we get rid of the input amount, then we can replace `syncReserveBalance` with something cheaper.
If we also get rid of the input token, then we can replace `syncReserveBalance` with something even cheaper.

solidity/contracts/converter/ConverterBase.sol
```diff
@@ -447,6 +454,25 @@ abstract contract ConverterBase is ConverterVersion, IConverter, TokenHolder, Co
...
+        uint256 amount
+    ) private {
+        if (token == NATIVE_TOKEN_ADDRESS) {
+            payable(to).transfer(amount);
+        } else {
+            token.safeTransfer(to, amount);
+        }
+    }
+
     /**
```
barakman, line 476: I'd wrap everything with `if (amount > 0) { ... }`, as any one of the reserve tokens may require this condition inside its `transfer` function.
TBH, this risk also exists in many other places in our code, and specifically in this contract.
But this new function is one 'bottleneck' where we can get rid of that risk for several different flows.

solidity/contracts/converter/ConverterUpgrader.sol
```diff
@@ -181,13 +206,36 @@ contract ConverterUpgrader is IConverterUpgrader, ContractRegistryClient {
...
+     */
+    function transferReserveBalances(
+        IConverter _oldConverter,
+        IConverter _newConverter,
+        uint16 _version
+    ) private {
+        if (_version <= 45) {
+            transferReserveBalancesVersion45(ILegacyConverter(address(_oldConverter)), _newConverter);
+
+            return;
```
barakman, line 224: Semantics: I'd use if/else instead.

solidity/contracts/converter/interfaces/IConverter.sol
```diff
@@ -44,16 +44,16 @@ interface IConverter is IOwned {
 
     function setConversionFee(uint32 _conversionFee) external;
 
-    function withdrawTokens(
+    function addReserve(IERC20 _token, uint32 _ratio) external;
```
barakman, line 47: Minor: We may as well rename `_ratio` to `_weight`.

solidity/contracts/converter/types/standard-pool/StandardPoolConverter.sol
```diff
@@ -247,32 +220,56 @@ contract StandardPoolConverter is
...
         IERC20 _token,
         address _to,
         uint256 _amount
-    ) public override(IConverter, TokenHolder) protected ownerOnly {
-        address converterUpgrader = addressOf(CONVERTER_UPGRADER);
+    ) public override(IConverter) protected ownerOnly {
```
barakman, line 264: same comment as the one in the ConverterBase

solidity/contracts/converter/types/standard-pool/StandardPoolConverter.sol
```diff
@@ -247,32 +220,56 @@ contract StandardPoolConverter is
...
+        // verify that the token is a reserve token and that the converter is inactive
+        require(reserveId != 0 && !isActive(), "ERR_ACCESS_DENIED");
 
-        // if the token is a reserve token, sync the reserve balance
-        if (reserveId != 0) {
-            syncReserveBalance(_token);
-        }
+        safeTransfer(_token, _to, _amount);
+
+        syncReserveBalance(_token);
```
barakman, line 272: same comment as the one in the ConverterBase

solidity/contracts/converter/types/standard-pool/StandardPoolConverter.sol
```diff
@@ -1286,6 +1283,25 @@ contract StandardPoolConverter is
...
     }
 
+    /**
+     * @dev transfers funds held by the contract and sends them to an account
+     *
+     * @param token ERC20 token contract address
+     * @param to account to receive the new amount
+     * @param amount amount to withdraw
+     */
+    function safeTransfer(
```
barakman, line 1293: same comment as the one in the ConverterBase

solidity/contracts/liquidity-protection/LiquidityProtection.sol
```diff
@@ -105,7 +105,7 @@ contract LiquidityProtection is ILiquidityProtection, Utils, Owned, ReentrancyGu
         store = ILiquidityProtectionStore(_contractAddresses[1]);
         stats = ILiquidityProtectionStats(_contractAddresses[2]);
         systemStore = ILiquidityProtectionSystemStore(_contractAddresses[3]);
-        wallet = ITokenHolder(_contractAddresses[4]);
+        wallet = ITokenHolder(payable(_contractAddresses[4]));
```
barakman, line 108: What does `payable` achieve here? I cannot see how this would resolve any compilation error or warning (let alone, runtime behavior).

solidity/contracts/utility/TokenHolder.sol
```diff
@@ -13,32 +13,66 @@ import "./interfaces/ITokenHolder.sol";
...
+    }
+
+    /**
+     * @dev transfers funds held by the contract and sends them to an account
+     *
+     * @param token ERC20 token contract address
+     * @param to account to receive the new amount
+     * @param amount amount to withdraw
+     */
+    function safeTransfer(
```
barakman, line 67: same comment as the one in the ConverterBase

barakman (changes requested):

solidity/contracts/converter/ConverterUpgrader.sol
lbeder: We always prefer early returns

solidity/contracts/converter/ConverterBase.sol
lbeder: I'll add an early return, but note that we always try to avoid "wrapping everything" to improve code readibility

solidity/contracts/converter/interfaces/IConverter.sol
lbeder: 👍

solidity/contracts/converter/ConverterBase.sol
lbeder: Let's don't change it, since all these interfaces are heavily affected by the network fees PR

solidity/contracts/liquidity-protection/LiquidityProtection.sol
lbeder: It is, since `TokenHandler` is now `payable`

solidity/contracts/converter/types/standard-pool/StandardPoolConverter.sol
lbeder: 👍

solidity/contracts/converter/types/standard-pool/StandardPoolConverter.sol
lbeder: 👍

solidity/contracts/converter/types/standard-pool/StandardPoolConverter.sol
lbeder: 👍

solidity/contracts/converter/ConverterBase.sol
barakman: My question answered.

solidity/contracts/liquidity-protection/LiquidityProtection.sol
barakman:
```
CompileError: /Users/leonid/work/bancor/contracts/solidity/contracts/liquidity-protection/LiquidityProtection.sol:108:18: TypeError: Explicit type conversion not allowed from non-payable "address" to "contract ITokenHolder", which has a payable fallback function.
        wallet = ITokenHolder(_contractAddresses[4]);
                 ^---------------------------------^
```

sshmaxime (approved):

solidity/contracts/utility/TokenHolder.sol
yudilevi: not critical

solidity/test/ConverterUpgrader.js
yudilevi: sounds good

solidity/contracts/converter/ConverterBase.sol
yudilevi: early return is preferable over more indentations.

solidity/contracts/utility/TokenHolder.sol
yudilevi: ok

solidity/contracts/converter/ConverterBase.sol
```diff
@@ -231,31 +213,54 @@ abstract contract ConverterBase is ConverterVersion, IConverter, TokenHolder, Co
...
      *
-     * @param _token   ERC20 token contract address
-     * @param _to      account to receive the new amount
-     * @param _amount  amount to withdraw
+     * @param _token ERC20 token contract address
+     * @param _to account to receive the new amount
+     * @param _amount amount to withdraw
      */
-    function withdrawTokens(
+    function withdrawInactiveTokens(
```
yudilevi, line 253: the name is a bit misleading, maybe just `withdrawReserve` and use the `inactive`, `validReserve` modifiers we already have instead of the inline logic?

alternatively - not support it at all :) any pool that's created through the registry (factory) is already active.

solidity/contracts/converter/types/standard-pool/StandardPoolConverter.sol
```diff
@@ -247,32 +220,56 @@ contract StandardPoolConverter is
...
      *
-     * @param _token   ERC20 token contract address
-     * @param _to      account to receive the new amount
-     * @param _amount  amount to withdraw
+     * @param _token ERC20 token contract address
+     * @param _to account to receive the new amount
+     * @param _amount amount to withdraw
      */
-    function withdrawTokens(
+    function withdrawInactiveTokens(
```
yudilevi, line 260: same comment as in converter base

solidity/contracts/converter/ConverterBase.sol
lbeder: I like the alternative better :)

yudilevi (approved):

solidity/contracts/converter/ConverterBase.sol
```diff
@@ -231,30 +213,32 @@ abstract contract ConverterBase is ConverterVersion, IConverter, TokenHolder, Co
     }
 
     /**
-     * @dev withdraws tokens held by the converter and sends them to an account
-     * can only be called by the owner
-     * note that reserve tokens can only be withdrawn by the owner while the converter is inactive
-     * unless the owner is the converter upgrader contract
+     * @dev transfers reserves to a new converter during an upgrade
```
barakman, line 216: transfers reserve balances

solidity/contracts/converter/ConverterBase.sol
```diff
@@ -231,30 +213,32 @@ abstract contract ConverterBase is ConverterVersion, IConverter, TokenHolder, Co
...
+     * can only be called by the converter upgraded which should be set at its owner
```
barakman, line 217: 'can be called only' IMO, but @yudilevi and me disagree on that

solidity/contracts/converter/ConverterBase.sol
```diff
@@ -231,30 +213,32 @@ abstract contract ConverterBase is ConverterVersion, IConverter, TokenHolder, Co
...
      *
-     * @param _token   ERC20 token contract address
-     * @param _to      account to receive the new amount
-     * @param _amount  amount to withdraw
+     * @param _newConverter account to receive the new amount
```
barakman, line 219: If it's a converter, then we should say 'address of the converter...'.

If it's some generic account, then we should use variable name 'account' or 'recipient' or similar.

solidity/contracts/converter/ConverterBase.sol
```diff
@@ -446,6 +432,29 @@ abstract contract ConverterBase is ConverterVersion, IConverter, TokenHolder, Co
         emit Conversion(_sourceToken, _targetToken, _trader, _amount, _returnAmount, int256(_feeAmount));
     }
 
+    /**
+     * @dev transfers funds held by the contract and sends them to an account
+     *
+     * @param token ERC20 token contract address
+     * @param to account to receive the new amount
+     * @param amount amount to withdraw
```
barakman, line 440: You have a redundant 'amount' here ('amount amount')

solidity/contracts/converter/ConverterBase.sol
```diff
@@ -446,6 +432,29 @@ abstract contract ConverterBase is ConverterVersion, IConverter, TokenHolder, Co
         emit Conversion(_sourceToken, _targetToken, _trader, _amount, _returnAmount, int256(_feeAmount));
     }
 
+    /**
+     * @dev transfers funds held by the contract and sends them to an account
+     *
+     * @param token ERC20 token contract address
+     * @param to account to receive the new amount
```
barakman, line 439: IMO, we should "redeploy" our previous convention of aligning the descriptions.
It is otherwise unreadable when viewing the contract, as in the example above:
```
@param to account to receive the new amount
```
Keep in mind that some users (including ourselves) read our documentation directly in the contracts and not on our GitBook.

solidity/contracts/converter/ConverterUpgrader.sol
```diff
@@ -64,8 +74,24 @@ contract ConverterUpgrader is IConverterUpgrader, ContractRegistryClient {
...
      */
-    function upgrade(uint16 _version) public override {
-        upgradeOld(IConverter(msg.sender), bytes32(uint256(_version)));
+    function upgrade(uint16 _version) external override {
+        upgrade(IConverter(msg.sender), _version);
+    }
+
+    /**
+     * @dev upgrades an old converter to the latest version
+     * will throw if ownership wasn't transferred to the upgrader before calling this function.
```
barakman, line 83: IMO: 'will throw' ==> 'reverts' (consistent with our present simple tense throughout the contracts).

solidity/contracts/converter/ConverterVersion.sol
```diff
@@ -2,5 +2,5 @@
 pragma solidity 0.6.12;
 
 contract ConverterVersion {
-    uint16 public constant version = 45;
+    uint16 public constant version = 46;
```
barakman, line 5: This will "collide" with the same change in my PR, or even worse - it will not "yell" when I merge my (identical) change.
Assuming that we want an increment of at most 1 on every release **to master**, perhaps we should both revert this change, and do it only before releasing the current version (0.6.34).

solidity/contracts/converter/types/standard-pool/StandardPoolConverter.sol
```diff
@@ -247,31 +220,32 @@ contract StandardPoolConverter is
     }
 
     /**
-     * @dev withdraws tokens held by the converter and sends them to an account
-     * can only be called by the owner
-     * note that reserve tokens can only be withdrawn by the owner while the converter is inactive
-     * unless the owner is the converter upgrader contract
+     * @dev transfers reserves to a new converter during an upgrade
```
barakman, line 223: transfers reserve balances

solidity/contracts/converter/types/standard-pool/StandardPoolConverter.sol
```diff
@@ -247,31 +220,32 @@ contract StandardPoolConverter is
...
+     * can only be called by the converter upgraded which should be set at its owner
      *
-     * @param _token   ERC20 token contract address
-     * @param _to      account to receive the new amount
-     * @param _amount  amount to withdraw
+     * @param _newConverter account to receive the new amount
```
barakman, line 226: If it's a converter, then we should say 'address of the converter...'.

If it's some generic account, then we should use variable name 'account' or 'recipient' or similar.

solidity/contracts/converter/types/standard-pool/StandardPoolConverter.sol
```diff
@@ -1285,6 +1259,29 @@ contract StandardPoolConverter is
         return _amount.mul(_reserveBalance) / _supply;
     }
 
+    /**
+     * @dev transfers funds held by the contract and sends them to an account
+     *
+     * @param token ERC20 token contract address
+     * @param to account to receive the new amount
+     * @param amount amount to withdraw
```
barakman, line 1267: 'amount amount'

solidity/contracts/utility/TokenHolder.sol
```diff
@@ -13,31 +13,70 @@ import "./interfaces/ITokenHolder.sol";
...
+     * @param to account to receive the new amount
+     * @param amounts amounts to withdraw
+     */
+    function withdrawTokensMultiple(
+        IERC20[] calldata tokens,
+        address to,
+        uint256[] calldata amounts
+    ) public virtual override ownerOnly validAddress(to) {
+        uint256 length = tokens.length;
+        require(length == amounts.length, "ERR_INVALID_LENGTH");
```
barakman, line 53: `ERR_INVALID_INPUT_LENGTH` IMO

solidity/contracts/utility/TokenHolder.sol
```diff
@@ -13,31 +13,70 @@ import "./interfaces/ITokenHolder.sol";
...
+
+        for (uint256 i = 0; i < length; ++i) {
```
barakman, line 55: We usually use postix incrementing (`i++`), though I admit that prefix incrementing is probably slightly cheaper.

solidity/test/TokenHolder.js
```diff
@@ -1,52 +1,179 @@
...
 
+const { NATIVE_TOKEN_ADDRESS } = require('./helpers/Constants');
+
 const TokenHolder = contract.fromArtifact('TokenHolder');
 const TestStandardToken = contract.fromArtifact('TestStandardToken');
 
 describe('TokenHolder', () => {
+    const getBalance = async (tokenAddress, account) => {
+        if (tokenAddress === NATIVE_TOKEN_ADDRESS) {
+            return balance.current(account);
```
barakman, line 15: I have no idea what `balance` is, but I would imagine that we require an `await` in order to resolve the promise object that it returns.
I understand that omitting an `await` on `return` statement is technically fine, but just in case we change this in the future to an assignment into some temporary variable (for example), and forget that we need an `await` here, it is better that we stick to placing `await` before every asynchronous function call.

solidity/test/TokenHolder.js
```diff
@@ -1,52 +1,179 @@
...
+        }
+
+        return (await TestStandardToken.at(tokenAddress)).balanceOf.call(account);
```
barakman, line 18: Watch out, we need two separate awaits here - one for the `at` (starting from web3.js v1.x), and one for the `balanceOf.call`.

solidity/test/TokenHolder.js
```diff
@@ -1,52 +1,179 @@
...
+        for (const tokenAddress of tokenAddresses) {
+            balances[tokenAddress] = await getBalance(tokenAddress, account);
+        }
+
+        return balances;
+    };
+
     let holder;
-    let erc20Token;
+    let token;
```
barakman, line 31: An optional suggestion: I'm highly in favor of starting to use token1/token2 (rather than token/token2).

barakman (changes requested):

solidity/contracts/converter/ConverterBase.sol
lbeder: 👍

solidity/contracts/converter/ConverterBase.sol
lbeder: Not important right now...

solidity/contracts/converter/ConverterBase.sol
lbeder: No, `amount` is the name of the parameter

solidity/contracts/converter/ConverterBase.sol
lbeder: Sure

lbeder: 👍

solidity/contracts/utility/TokenHolder.sol
lbeder: This is the error message that is used in other of our repos, so I'll keep it for now

solidity/contracts/utility/TokenHolder.sol
lbeder: In solidity, `++i` is prefered

solidity/test/TokenHolder.js
lbeder: Actually no `await`ing on returns from an `async` function is the preferred way and it's also more performarnt

solidity/test/TokenHolder.js
lbeder: Same as above

solidity/test/TokenHolder.js
lbeder: Not for now, though

barakman (approved):

PR #563, opened 2021-03-22
StandardPoolConverter general improvements

barakman:
In the SP converter and all dependent contracts, replace all dynamic arrays with static arrays of length 2.


solidity/contracts/converter/types/standard-pool/StandardPoolConverter.sol
```diff
@@ -31,7 +31,8 @@ contract StandardPoolConverter is ConverterVersion, IConverter, ContractRegistry
 
     uint256 private __reserveBalances;
     uint256 private __reserveBalancesProd;
-    IERC20[] private __reserveTokens;
+    IERC20[2] private __reserveTokens;
+    uint16 __reserveTokensLength;
```
lbeder, line 35: Can you try to use an `immutable` or hardcoded 2 instead of a storage variable?

lbeder (changes requested):
Please see my comments below. This PR can be a good opportunity to some of the additional optimizations we've discussed

solidity/contracts/converter/types/standard-pool/StandardPoolConverter.sol
barakman: I don't understand.
Have you followed variable `__reserveTokensLength` in the code? It can be anything between 0 and 2.

lbeder: Ah, I remembered that we are always creating it with two reserves from the get-go.

yudilevi: @barakman any reason why it's not `_reserveTokensLength` according to our convention?

barakman: I don't understand the question.

yudilevi: __reserveTokensLength should be _reserveTokensLength

lbeder:
@barakman I have comments, other than:
1. Considering adding `calldata` optimization (whenever applicable).
2. Resolving merging conflict with `StandardPoolConverter.sol`

barakman:
> @barakman I have comments, other than:
> 
> 1. Considering adding `calldata` optimization (whenever applicable).
> 2. Resolving merging conflict with `StandardPoolConverter.sol`

1 - at https://trello.com/c/vUmFSAJ0
2 - done

lbeder (approved):

solidity/contracts/converter/types/standard-pool/StandardPoolConverter.sol
```diff
@@ -31,7 +31,8 @@ contract StandardPoolConverter is ConverterVersion, IConverter, ContractRegistry
 
     uint256 private __reserveBalances;
     uint256 private _reserveBalancesProduct;
-    IERC20[] private __reserveTokens;
+    IERC20[2] private __reserveTokens;
+    uint16 _reserveTokensLength;
```
yudilevi, line 35: _reserveTokensLength -> _reserveTokensCount?

solidity/contracts/converter/types/standard-pool/StandardPoolConverter.sol
```diff
@@ -274,15 +275,15 @@ contract StandardPoolConverter is ConverterVersion, IConverter, ContractRegistry
...
+        return _reserveTokensLength;
     }
 
     /**
      * @dev returns the array of reserve tokens
      *
      * @return array of reserve tokens
      */
-    function reserveTokens() public view returns (IERC20[] memory) {
+    function reserveTokens() public view returns (IERC20[2] memory) {
```
yudilevi, line 286: isn't that an interface breaking change that also means contracts can't use the function universally on all converter versions?

solidity/contracts/converter/types/standard-pool/StandardPoolConverter.sol
```diff
@@ -974,9 +975,9 @@ contract StandardPoolConverter is ConverterVersion, IConverter, ContractRegistry
      */
     function removeLiquidity(
```
yudilevi, line 976: isn't this another interface breaking change?

yudilevi (changes requested):

solidity/contracts/converter/types/standard-pool/StandardPoolConverter.sol
barakman: Yes, it is indeed breaking onchain API (but not offchain API).
However, once all SP converters are upgraded, the problem will be only per type, not per version (i.e., you'll need to fetch the return value differently for V1 and for SP).

solidity/contracts/converter/types/standard-pool/StandardPoolConverter.sol
barakman: Yes.
It is solvable here, since we can overload the function and have the older version as part of the contract (unlike `function reserveTokens`, where the change is only in the return-value type).
However, it would be very unfortunate if we ended up keeping the older version. Those dynamic arrays yield a chaos in the contract.

solidity/contracts/converter/types/standard-pool/StandardPoolConverter.sol
barakman: @yudilevi:
Yes, but I don't think that it is used anywhere onchain outside this contract.

solidity/contracts/converter/types/standard-pool/StandardPoolConverter.sol
barakman: @yudilevi:
Yes, and this is indeed a problem in the LP contract which relies on the old versions of both `addLiquidity` and `removedLiquidity` (a problem which was "hidden from the compiler" by the `IConverter` interface that we have declared locally inside that contract).
In any case, I have overloaded these two functions with versions that take and return dynamic arrays.
They are placed at the bottom of the file (alongside all the other deprecated functions).
This will allow us to upgrade the LP contract independently of the upgrade of each one of the converters.
We should make sure that there are no other API-breaking changes, as this one has eluded the compiler.

solidity/contracts/converter/types/standard-pool/StandardPoolConverter.sol

solidity/contracts/converter/types/standard-pool/StandardPoolConverter.sol
yudilevi: it's used by others

solidity/contracts/converter/types/standard-pool/StandardPoolConverter.sol
```diff
@@ -997,7 +998,7 @@ contract StandardPoolConverter is ConverterVersion, IConverter, ContractRegistry
         // transfer the fees and sync the balances to ensure no mismatch
         (oldReserveBalances[0], oldReserveBalances[1]) = processNetworkFees(0);
 
-        uint256[] memory reserveAmounts = removeLiquidityReserveAmounts(_amount, totalSupply, oldReserveBalances);
+        uint256[2] memory reserveAmounts = removeLiquidityReserveAmounts(_amount, totalSupply, oldReserveBalances);
 
         for (uint256 i = 0; i < 2; i++) {
```
yudilevi, line 1003: no need for all these for loops if there are only 2.

solidity/contracts/converter/types/standard-pool/StandardPoolConverter.sol
yudilevi: wouldn't that add yet another layer of friction for integrators? another public overloaded functions?

solidity/contracts/converter/types/standard-pool/StandardPoolConverter.sol
barakman: Yes, but there are 7 lines of code inside the loop. Do you actually want to replicate them instead of using a `for` loop?

solidity/contracts/converter/types/standard-pool/StandardPoolConverter.sol
barakman: No.
Onchain, the compiler knows which overloaded version to call, so there's no issue.
Offchain, the only friction is having to use the explicit (rather "ugly") syntax of web3. But this is an issue once you have 2 overloaded versions of the function, which is already the case before we add this (3rd) overloaded version.

solidity/contracts/converter/types/standard-pool/StandardPoolConverter.sol
barakman: We can apply a fix very easily, by reverting this function to its original version (which returns a dynamic array), and simply copy the array of tokens into a dynamic array before returning it. This is less efficient, but if this function is called only from constants functions, then there's no issue.
The issue here is that we will never be able to get rid of that inefficiency, because we cannot overload this function (as functions cannot be overloaded only by distinguishing the return-value type).
So preferably, if we know of anyone who has embedded it **in their contracts**, then we might want to try and resolve it with them.
And again - if it's not used in any contract, then we may as well change this function's prototype (rather then keeping it with an inefficient internal implementation).

yudilevi: We'll figure it out in the next pool type and then we can remove the backward compatible functions.
However, the bigger question here is why are we using arrays to begin with.
This converter has exactly 2 reserves, why not just use 2 values (everywhere)? Keep is stupid simple for devs.

solidity/contracts/converter/types/standard-pool/StandardPoolConverter.sol
```diff
@@ -627,11 +659,7 @@ contract StandardPoolConverter is ConverterVersion, IConverter, ContractRegistry
         setReserveBalances(sourceId, targetId, actualSourceBalance, targetBalance - amount);
 
         // transfer funds to the beneficiary in the to reserve token
-        if (_targetToken == NATIVE_TOKEN_ADDRESS) {
-            _beneficiary.transfer(amount);
-        } else {
-            _targetToken.safeTransfer(_beneficiary, amount);
-        }
+        safeTransfer(_targetToken, _beneficiary, amount);
```
yudilevi, line 662: might be worth adding a private function that checks for `NATIVE_TOKEN_ADDRESS` and handles transferring it interrnally

yudilevi (changes requested):

solidity/contracts/converter/types/standard-pool/StandardPoolConverter.sol
```diff
@@ -272,15 +275,15 @@ contract StandardPoolConverter is ConverterVersion, IConverter, ContractRegistry
      * @return number of reserve tokens
      */
     function reserveTokenCount() public view returns (uint16) {
-        return uint16(__reserveTokens.length);
+        return _reserveTokensCount;
```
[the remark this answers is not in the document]
barakman: I don't understand.

solidity/contracts/converter/types/standard-pool/StandardPoolConverter.sol
barakman: That's exactly what function `safeTransfer` is and does, so I don't understand the comment.

PR #567, opened 2021-03-24
Fix the prototype of function `addLiquidityReturn`

barakman:
1. Fix function 'addLiquidityReturn' to take all reserve tokens and their corresponding amounts instead of a single reserve token and its corresponding amount (in the V1-Pool-Converter contract and in the Standard-Pool-Converter contract).

2. Fix all web3 calls to functions 'addLiquidity' and 'removeLiquidity' by stating the overloaded version of each function explicitly (in the test of the Standard-Pool-Converter contract and in the test of the Fixed-Rate-Pool-Converter contract).


solidity/contracts/converter/types/standard-pool/StandardPoolConverter.sol
```diff
@@ -1064,20 +1043,23 @@ contract StandardPoolConverter is ConverterVersion, IConverter, ContractRegistry
...
      */
-    function addLiquidityReturn(IERC20 _reserveToken, uint256 _reserveAmount) public view returns (uint256) {
+    function addLiquidityReturn(
+        IERC20[2] memory _reserveTokens,
+        uint256[2] memory _reserveAmounts
+    ) public view returns (uint256) {
         uint256 totalSupply = IDSToken(address(anchor)).totalSupply();
-        uint256 _reserveBalance = baseReserveBalance(_reserveToken);
-        return fundSupplyAmount(totalSupply, _reserveBalance, _reserveAmount);
+        uint256[2] memory _reserveBalances = baseReserveBalances(_reserveTokens);
```
lbeder, line 1060: Should be `reserveBalances`

solidity/test/ConverterLiquidity.js
```diff
@@ -318,17 +316,14 @@ describe('ConverterLiquidity', () => {
...
 
-    const getLiquidityReturns = async (firstTime, converter, reserveTokens, reserveAmounts) => {
+    const getLiquidityReturn = async (firstTime, converter, reserveTokens, reserveAmounts) => {
         if (firstTime) {
             const length = Math.round(
                 reserveAmounts.map((reserveAmount) => reserveAmount.toString()).join('').length / reserveAmounts.length
             );
-            const retVal = new BN('1'.padEnd(length, '0'));
-            return reserveAmounts.map((reserveAmount, i) => retVal);
+            return new BN('1'.padEnd(length, '0'));
```
lbeder, line 324: Can you use `pow` instead of string padding?

solidity/test/FixedRatePoolConverter.js
```diff
@@ -164,7 +164,7 @@ describe('FixedRatePoolConverter', () => {
         await reserveToken.approve(converter.address, value, { from: sender });
         await reserveToken2.approve(converter.address, value, { from: sender });
 
-        const res = await converter.addLiquidity(
+        const res = await converter.methods['addLiquidity(address[2],uint256[2],uint256)'](
```
lbeder, line 167: The usual approach is to implement a differently named "override" in the test helper contract, instead of explicitly working with selectors. Could you refactor this away?

solidity/test/FixedRatePoolConverter.js
```diff
@@ -192,7 +192,7 @@ describe('FixedRatePoolConverter', () => {
     it('verifies the TokenRateUpdate event after removing liquidity', async () => {
         const converter = await initConverter(true, false);
 
-        const res = await converter.removeLiquidity(
+        const res = await converter.methods['removeLiquidity(uint256,address[2],uint256[2])'](
```
lbeder, line 195: See my comment below (etc.)

solidity/utils/test_deployment.js
```diff
@@ -308,9 +308,18 @@ const run = async () => {
...
-            const deployedConverterType = {1: 'LiquidityPoolV1Converter', 3: 'StandardPoolConverter'}[type];
-            const deployedConverter = deployed(web3, deployedConverterType, converterBase._address);
-            await execute(deployedConverter.methods['addLiquidity(address[],uint256[],uint256)'](tokens, amounts, 1), value);
+            const contractName = {
+                1: 'LiquidityPoolV1Converter',
+                3: 'StandardPoolConverter'
+            };
+
+            const funcSignature = {
+                1: 'addLiquidity(address[],uint256[],uint256)',
```
lbeder, line 317: My previous comment couldn't be applied here, but that's understandable. Hopefully, we'd remove this in the future as well.

lbeder (changes requested):
Please see my comments

solidity/test/ConverterLiquidity.js
barakman: done

solidity/test/FixedRatePoolConverter.js
barakman: There's another trick BTW.

In the actual test, the Truffle Contract object (`converter`) has the following fields:
1. `converter.methods['addLiquidity(address[2],uint256[2],uint256)']`
2. `converter.methods['addLiquidity(uint256,uint256,uint256)']`
3. `converter.addLiquidity`

When the Truffle Contract object is created, the 3rd field is set to "point" to either one of the other fields (depending on how the ABI was built).

So immediately after creating it, we can override this, and set it to "point" to the one that we intend to call (assuming that we intend to call only one of them, as is the case at present).

Subsequently, we can use  `converter.addLiquidity` in the usual ("natural") manner.

I didn't use this option, because we create the Truffle Contract object in several different places, so the benefit of keeping our code clean is diminished.

solidity/utils/test_deployment.js
barakman: Yup, and we can use a similar trick as the one that I've mentioned above, but it wouldn't yield any significant improvement to the current implementation in this script.

solidity/contracts/converter/types/standard-pool/StandardPoolConverter.sol
lbeder: You can use a different name, as long as it doesn't introduce new/conflicting naming conventions.

solidity/test/FixedRatePoolConverter.js
lbeder: I totally understand the problem, it's just that the solution is too "low level"

solidity/utils/test_deployment.js
lbeder: Yeah, but if we can limit this only here - it'd be better

solidity/test/FixedRatePoolConverter.js
barakman: @lbeder : This request is revoked, correct?

solidity/test/FixedRatePoolConverter.js
barakman: @lbeder : This request is revoked, correct?

lbeder: Yap

solidity/test/FixedRatePoolConverter.js
lbeder: Correct

solidity/contracts/converter/types/standard-pool/StandardPoolConverter.sol
barakman: Fixed (here and in a few other places).

lbeder (approved):

solidity/contracts/converter/types/standard-pool/StandardPoolConverter.sol
```diff
@@ -1050,34 +1029,37 @@ contract StandardPoolConverter is ConverterVersion, IConverter, ContractRegistry
         uint256 _reserveTokenIndex,
         uint256 _reserveAmount
     ) public view returns (uint256[2] memory) {
-        uint256[2] memory _reserveAmounts;
+        uint256[2] memory reserveAmounts;
 
         uint256 totalSupply = IDSToken(address(anchor)).totalSupply();
-        uint256[2] memory _reserveBalances = baseReserveBalances(_reserveTokens);
-        uint256 amount = fundSupplyAmount(totalSupply, _reserveBalances[_reserveTokenIndex], _reserveAmount);
+        uint256[2] memory balancesMinusFees = baseReserveBalances(_reserveTokens);
```
yudilevi, line 1035: balancesMinusFees -> baseBalances (keep it consistent)

solidity/contracts/converter/types/standard-pool/StandardPoolConverter.sol
```diff
@@ -1050,34 +1029,37 @@ contract StandardPoolConverter is ConverterVersion, IConverter, ContractRegistry
...
+        uint256 amount = fundSupplyAmount(totalSupply, balancesMinusFees[_reserveTokenIndex], _reserveAmount);
 
         for (uint256 i = 0; i < 2; i++) {
```
yudilevi, line 1038: no need for `for` loops if it's always 2 reserves

yudilevi (changes requested):

solidity/contracts/converter/types/standard-pool/StandardPoolConverter.sol

solidity/contracts/converter/types/standard-pool/StandardPoolConverter.sol

PR #569, opened 2021-03-25
Vortex Burner

lbeder:


solidity/contracts/vortex/VortexBurner.sol
```diff
@@ -0,0 +1,558 @@
...
+        _govTokenGovernance = govTokenGovernance;
+        _govToken = govTokenGovernance.token();
+        _stats = stats;
+
+        _burnIncentiveFee = burnIncentiveFee;
+        _maxBurnIncentiveFeeAmount = maxBurnIncentiveFeeAmount;
+    }
+
+    // ensures that the fee is valid
+    modifier validFee(uint32 fee) {
```
barakman, line 116: Already implemented in parent contract `Utils`.

barakman (changes requested):

solidity/contracts/liquidity-protection/LiquidityProtection.sol
```diff
@@ -1082,7 +1082,7 @@ contract LiquidityProtection is ILiquidityProtection, Utils, Owned, ReentrancyGu
         amounts[1] = _reserveAmount2;
         _converter.addLiquidity{ value: _value }(reserveTokens, amounts, 1);
 
-        // ensure that the contract can receive ETH
+        // ensure that the contract can no longer receive ETH
```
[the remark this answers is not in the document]
lbeder: 👍

solidity/contracts/utility/ContractRegistryClient.sol
```diff
@@ -20,6 +20,7 @@ contract ContractRegistryClient is Owned, Utils {
     bytes32 internal constant BANCOR_X = "BancorX";
     bytes32 internal constant BANCOR_X_UPGRADER = "BancorXUpgrader";
     bytes32 internal constant LIQUIDITY_PROTECTION = "LiquidityProtection";
+    bytes32 internal constant NETWORK_SETTINGS = "NetworkSettings";
```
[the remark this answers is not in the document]
lbeder: This (and similar) notes will be addressed by rebasing this from the #560

solidity/contracts/vortex/VortexBurner.sol
lbeder: It seems that it's only in `NetworkSettings`. I'll move it to `Utils`

solidity/contracts/vortex/VortexBurner.sol
```diff
@@ -0,0 +1,558 @@
...
+/**
+ * @dev This contract provides any user to perform a vortex
+ */
+contract VortexBurner is Owned, Utils, ReentrancyGuard, ContractRegistryClient {
+    using SafeMath for uint256;
+    using Math for uint256;
+    using SafeERC20 for IERC20;
+
+    // the vortex is only designed to work with 50/50 standard pool converters
+    uint32 private constant STANDARD_POOL_RESERVE_WEIGHT = 500000;
```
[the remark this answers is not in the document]
lbeder: 👍

solidity/contracts/vortex/VortexBurner.sol
```diff
@@ -0,0 +1,558 @@
...
+        uint32 burnIncentiveFee,
+        uint256 maxBurnIncentiveFeeAmount,
+        IContractRegistry registry
+    )
+        public
+        validAddress(address(networkToken))
+        validAddress(address(govTokenGovernance))
+        validAddress(address(stats))
+        validFee(burnIncentiveFee)
+        ContractRegistryClient(registry)
```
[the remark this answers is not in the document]
lbeder: 👍

solidity/contracts/vortex/VortexBurner.sol
```diff
@@ -0,0 +1,558 @@
...
+
+    // ensures that the fee is valid
+    modifier validFee(uint32 fee) {
+        _validFee(fee);
+        _;
+    }
+
+    // error message binary size optimization
+    function _validFee(uint32 fee) internal pure {
+        require(fee <= PPM_RESOLUTION, "ERR_INVALID_FEE");
```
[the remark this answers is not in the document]
lbeder: See above

solidity/contracts/vortex/VortexBurner.sol
```diff
@@ -0,0 +1,558 @@
...
+    }
+
+    /**
+     * @dev ETH receive callback
+     */
+    receive() external payable {}
+
```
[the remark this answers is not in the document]
lbeder: We actually discussed this specifically and decided against it. I'll explain later

solidity/contracts/vortex/VortexBurner.sol
```diff
@@ -0,0 +1,558 @@
...
+        // update the stats of the vortex
+        _stats.setLastVortexTime(block.timestamp);
+        _stats.incTotalBurnedAmount(totalGovTokenAmountToBurn);
+
+        // burn all the converter governance tokens
+        _govTokenGovernance.burn(totalGovTokenAmountToBurn);
+
+        // transfer the incentive fee to the caller
+        _networkToken.transfer(msg.sender, incentiveFeeAmount);
```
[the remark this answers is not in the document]
lbeder: The commission is from the total converted amount (e.g., a 50% of 300 is the same as 50% from 100 x3)

solidity/contracts/vortex/VortexBurner.sol
```diff
@@ -0,0 +1,558 @@
...
+     *
+     * @param tokens the tokens to convert
+     *
+     * @return the amounts of the converted tokens
+     * @return the network token amounts the tokens were converted to
+     * @return the total burned amount in this vortex
+     * @return the incentive fee resulting from this vortex
+     */
+    function availableVortex(IERC20[] calldata tokens)
+        external
```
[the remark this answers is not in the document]
lbeder: It involves extra gas costs we wanted to avoid (an extra `rateByPath`, for example)

solidity/contracts/vortex/VortexBurner.sol
```diff
@@ -0,0 +1,558 @@
...
+        for (uint256 i = 0; i < tokens.length; ++i) {
+            IERC20 token = tokens[i];
+
+            address[] memory path = new address[](3);
+            path[0] = address(token);
+
+            // don't look up for a converter for either the network or the governance token, since they are going to be
+            // handled in a special way during the burn itself
+            if (token == _networkToken || token == _govToken) {
+                path[1] = address(0);
```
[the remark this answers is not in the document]
lbeder: I'll make it less explicit

solidity/contracts/vortex/VortexBurner.sol
barakman: Watch out! It's implemented in NetworkSettings separately because the network fee is allowed to be zero, while we generally don't allow zero portions elsewhere. So the modifier in the NetworkSettings contract is different from the one in Utils.

solidity/contracts/vortex/VortexBurner.sol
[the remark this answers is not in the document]
barakman: So I guess I was wrong, we cannot use the one in Utils, because we wanna allow zero fee.

solidity/contracts/vortex/VortexBurner.sol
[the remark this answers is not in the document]
barakman: I do not understand.

solidity/utils/test_deployment.js
```diff
@@ -169,10 +169,13 @@ const run = async () => {
     const converterRegistryData = await web3Func(deploy, 'converterRegistryData', 'ConverterRegistryData', [
         contractRegistry._address
     ]);
+
+    const networkFeeWallet = await web3Func(deploy, 'networkFeeWallet', 'TokenHolder', []);
     const networkSettings = await web3Func(deploy, 'networkSettings', 'NetworkSettings', [
-        getConfig().networkSettingsParams.networkFeeWallet,
-        percentageToPPM(getConfig().networkSettingsParams.networkFee)
+        networkFeeWallet._address,
```
barakman, line 175: Why not support the configuration of `networkFee`?

barakman, line 175: If you choose to remove these configuration parameters, then you also need to update the README.md file (same folder).

solidity/utils/test_deployment.js
```diff
@@ -290,11 +293,14 @@ const run = async () => {
                 weights
             )
         );
+        console.log('BBB');
```
barakman, line 296: You forgot to remove your debug leftovers.

solidity/utils/test_deployment.js
```diff
@@ -290,11 +293,14 @@ const run = async () => {
...
         const converterAnchor = deployed(
             web3,
             'IConverterAnchor',
             await converterRegistry.methods.getAnchor(index).call()
         );
+        console.log('CCC');
+
```
barakman, line 303: You forgot to remove your debug leftovers.

solidity/contracts/vortex/VortexBurner.sol
```diff
@@ -0,0 +1,522 @@
...
+
+    /**
+     * @dev triggered after a completed vortex
+     *
+     * @param tokens the converted tokens
+     * @param amounts the amounts of the converted tokens
+     * @param conversionAmounts the network token amounts the tokens were converted to
+     * @param totalBurnedAmount the total burned amount in this vortex
+     */
+    event Burned(IERC20[] tokens, uint256[] amounts, uint256[] conversionAmounts, uint256 totalBurnedAmount);
```
barakman, line 77: Missing some `memory` qualifiers?

solidity/contracts/vortex/VortexBurner.sol
```diff
@@ -0,0 +1,522 @@
...
+
+        uint256[] memory networkTokenConversionAmounts = new uint256[](tokens.length);
+        uint256 grossNetworkTokenConversionAmount = 0;
+        uint256 totalGovTokenAmountToBurn = 0;
+
+        for (uint256 i = 0; i < paths.length; ++i) {
+            // avoid empty conversions
+            uint256 amount = amounts[i];
+            if (amount == 0) {
+                continue;
```
barakman, line 180: Why not filter these out in `vortexStrategy`?

solidity/contracts/vortex/VortexBurner.sol
lbeder: That's actually fine, because the burning incentive fee can zero as well. I think that, in any case, this should be the use case and the difference between `validPortion` and `validFee`, like in both of these cases.

solidity/contracts/vortex/VortexBurner.sol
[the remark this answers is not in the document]
lbeder: Yeah, but I've moved `validFee` to `Utils` as well. Please check my newer commits

solidity/contracts/vortex/VortexBurner.sol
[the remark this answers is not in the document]
lbeder: I meant that I'll remove this explicit setting. Please check my fix

solidity/contracts/vortex/VortexBurner.sol
```diff
@@ -0,0 +1,522 @@
...
+                    // if the source token is a regular token, approve the converter to withdraw the token amount
+                    ensureAllowance(token, network, amount);
+                } else {
+                    // if the source token is actually an ETH reserve, make sure to pass its value to the network
+                    value = amount;
+                }
+
+                // perform the actual conversion and optionally send ETH to the network
+                uint256 networkTokenConversionAmount =
+                    network.convertByPath2{ value: value }(path, amount, 1, address(this));
```
barakman, line 210: So min-return does not protect us here (and we obviously don't want to ask the user to provide it per token).
How can we be sure that we're protected from front-running here?

solidity/contracts/vortex/VortexBurner.sol
```diff
@@ -0,0 +1,522 @@
...
+        (uint256 netNetworkTokenConversionAmount, uint256 incentiveFeeAmount) =
+            applyIncentiveFee(grossNetworkTokenConversionAmount);
+
+        // approve the governance token converter to withdraw the network token amount
+        ensureAllowance(_networkToken, network, netNetworkTokenConversionAmount);
+
+        // convert all network token amounts to the governance token
+        totalGovTokenAmountToBurn = totalGovTokenAmountToBurn.add(
+            network.convertByPath2(govPath, netNetworkTokenConversionAmount, 1, address(this))
+        );
```
barakman, line 228: Same question about min-return, as above. Perhaps we should first call `rateByPath` in order to determine the min-return (e.g., 90% of the returned value)?

solidity/contracts/vortex/VortexBurner.sol
```diff
@@ -0,0 +1,522 @@
...
+
+        uint256[] memory networkTokenConversionAmounts = new uint256[](tokens.length);
+        uint256 grossNetworkTokenConversionAmount = 0;
+        uint256 totalGovTokenAmountToBurn = 0;
+
+        for (uint256 i = 0; i < paths.length; ++i) {
+            // avoid empty conversions
+            uint256 amount = amounts[i];
+            if (amount == 0) {
+                continue;
```
barakman, line 280: Same comment as previous (why not filter out these amounts in `vortexStrategy`?).

solidity/contracts/vortex/VortexBurner.sol
```diff
@@ -0,0 +1,522 @@
...
+            } else {
+                // calculate the expected target amount
+                uint256 networkTokenConversionAmount = network.rateByPath(path, amount);
+                require(networkTokenConversionAmount > 0, "ERR_ZERO_TARGET_AMOUNT");
+
+                networkTokenConversionAmounts[i] = networkTokenConversionAmount;
+                grossNetworkTokenConversionAmount = grossNetworkTokenConversionAmount.add(networkTokenConversionAmount);
+            }
+        }
+
```
barakman, line 306: I have a dejavu here.
Seems like a lot of duplicated code.
Perhaps we can move some pieces of code into common utility functions?

solidity/contracts/vortex/VortexBurner.sol
```diff
@@ -0,0 +1,522 @@
...
+            paths[i] = path;
+
+            // make sure to retrieve the balance of either an ERC20 or an ETH reserve
+            if (token == NATIVE_TOKEN_ADDRESS) {
+                amounts[i] = address(feeWallet).balance;
+            } else {
+                amounts[i] = token.balanceOf(address(feeWallet));
+            }
+        }
```
barakman, line 381: This piece of code appears in several other contracts; consider an internal function in Utils.

solidity/contracts/vortex/VortexBurner.sol
```diff
@@ -0,0 +1,522 @@
...
+     * @return the net amount
+     * @return the fee amount
+     */
+    function applyIncentiveFee(uint256 amount) private view returns (uint256, uint256) {
+        uint256 fee = _burnIncentiveFee;
+        if (fee == 0) {
+            return (amount, 0);
+        }
+
+        uint256 incentiveAmount = Math.min(amount.mul(fee).div(PPM_RESOLUTION), _maxBurnIncentiveFeeAmount);
```
barakman, line 406: Minor: can use `/` instead of `div`.

solidity/contracts/vortex/VortexBurner.sol
```diff
@@ -0,0 +1,522 @@
...
+
+        return (amount.sub(incentiveAmount), incentiveAmount);
```
barakman, line 408: Can use `amount - incentiveAmount` here.

solidity/contracts/vortex/VortexBurner.sol
```diff
@@ -0,0 +1,522 @@
...
+     *
+     * @param amount the network tokens amount
+     *
+     * @return the net amount
+     * @return the fee amount
+     */
+    function applyIncentiveFee(uint256 amount) private view returns (uint256, uint256) {
+        uint256 fee = _burnIncentiveFee;
+        if (fee == 0) {
+            return (amount, 0);
```
barakman, line 403: This piece of code (special case) is redundant.
You can reduce this function to be only the last two lines of it.

solidity/contracts/vortex/VortexBurner.sol
```diff
@@ -0,0 +1,522 @@
...
+        // initialize both the source and the target tokens
+        IERC20[] memory reserveTokens = new IERC20[](2);
+        reserveTokens[0] = _networkToken;
+        reserveTokens[1] = token;
+
+        // make sure to only look up for 50/50 converters
+        uint32[] memory standardReserveWeights = new uint32[](2);
+        standardReserveWeights[0] = STANDARD_POOL_RESERVE_WEIGHT;
+        standardReserveWeights[1] = STANDARD_POOL_RESERVE_WEIGHT;
+
```
barakman, line 433: Minor: I'd use `PPM_RESOLUTION / 2` here instead of declaring another constant.

solidity/contracts/vortex/VortexBurner.sol
```diff
@@ -0,0 +1,522 @@
...
+    ) private {
+        address networkAddress = address(network);
+        uint256 allowance = token.allowance(address(this), networkAddress);
+        if (allowance < amount) {
+            if (allowance > 0) {
+                token.safeApprove(networkAddress, 0);
+            }
+            token.safeApprove(networkAddress, amount);
+        }
+    }
```
barakman, line 466: I think it's time for us to consider adding this function in Utils.

solidity/contracts/vortex/VortexBurner.sol
```diff
@@ -0,0 +1,522 @@
...
+
+    /**
+     * @dev returns whether the provided list of tokens include duplicates
+     *
+     * @param tokens the list of tokens to check
+     *
+     * @return whether the provided list of tokens include duplicates
+     */
+    function hasDuplicates(IERC20[] calldata tokens) private pure returns (bool) {
```
barakman, line 511: Perhaps rename this function to `invalidTokenList` and also check for zero addresses?

solidity/contracts/vortex/VortexBurner.sol
```diff
@@ -0,0 +1,522 @@
...
+            for (uint256 j = i + 1; j < tokens.length; ++j) {
+                if (tokens[i] == tokens[j]) {
+                    return true;
+                }
+            }
+        }
+
+        return false;
+    }
+}
```
barakman, line 522: Consider flipping between the `true` and `false`, which will allow to get rid of the `!` every place the function is called (less code and less gas).

barakman (changes requested):

solidity/utils/test_deployment.js
lbeder: With your permission, I'll remove it from `README.md` as well. It's reconfigure, so it's unnecessary to configure it with a value during a test deployment and the less configuration options we'd have - the better

solidity/utils/test_deployment.js
lbeder: Oops

solidity/utils/test_deployment.js
lbeder: Oops

solidity/contracts/vortex/VortexBurner.sol
lbeder: You can't add storage qualifiers to events :(

solidity/contracts/vortex/VortexBurner.sol
lbeder: I wanted to avoid the extra gas costs in this case

solidity/contracts/vortex/VortexBurner.sol
lbeder: That would be a code duplication we want to generally avoid

solidity/contracts/vortex/VortexBurner.sol
lbeder: Probably, but we currently don't have a place for that. I'll open a task for it

solidity/contracts/vortex/VortexBurner.sol
lbeder: This method is only used in `availableVortex`, so the name would be misleading. I'd prefer to keep it. It's more descriptive.

solidity/contracts/vortex/VortexBurner.sol
lbeder: But then its name would need to be `hasNoDuplicated`.... let's keep this way

solidity/utils/test_deployment.js
lbeder: See my comment below

solidity/contracts/vortex/VortexBurner.sol
lbeder: Yes, good idea

solidity/contracts/vortex/VortexBurner.sol
lbeder: Yes, this is by design, like we've discussed during the call. Allowing the caller to arb the conversion is part of the incentive mechanism

solidity/contracts/vortex/VortexBurner.sol
lbeder: See my comment above

solidity/contracts/vortex/VortexBurner.sol
lbeder: Fixed

solidity/contracts/vortex/VortexBurner.sol
lbeder: We discussed this as well. It's hard to deduplicate this code because of the subtle difference embedded with (`convertByPath2` vs `rateByPath`, `ensureAllowance` and ETH handling, etc.). I always try to deduplicate code, but sometimes it's hard or adds unnecessary switching logic

solidity/contracts/vortex/VortexBurner.sol
lbeder: `Utils` shouldn't handle token-specific stuff. I've already opened a task for this.

solidity/contracts/vortex/VortexBurner.sol
lbeder: 👍

solidity/contracts/vortex/VortexBurner.sol
lbeder: 👍

solidity/contracts/vortex/VortexBurner.sol
barakman: NP, remember that next time you ask me to unify the network-fee functions ;)

solidity/contracts/vortex/VortexBurner.sol
barakman: It's an extremely minor gas-cost (a couple of arithmetic operations on non-storage variables), and in any case, the branch itself also yields some gas-cost.
IMO, when it comes to minor gas issues like this one vs an additional piece code which is effectively redundant, I'd opt for cleaner code.

solidity/test/VortexBurner.js
```diff
@@ -0,0 +1,691 @@
...
+const StandardPoolConverterFactory = contract.fromArtifact('TestStandardPoolConverterFactory');
+const StandardPoolConverter = contract.fromArtifact('TestStandardPoolConverter');
+const TokenHolder = contract.fromArtifact('TokenHolder');
+const TokenGovernance = contract.fromArtifact('TestTokenGovernance');
+const NetworkSettings = contract.fromArtifact('NetworkSettings');
+const VortexBurner = contract.fromArtifact('VortexBurner');
+
+const PPM_RESOLUTION = new BN(1_000_000);
+
+const TKN = new BN(10).pow(new BN(18));
```
barakman, line 34: `ONE_TOKEN` is more descriptive IMO (especially since we use "TKN" as part of our terminology for some other purpose).
BTW, did you take this from my network-fee tests?

solidity/contracts/vortex/VortexBurner.sol
lbeder: It's a different story, but I will own your a favour in any case 😉

solidity/test/VortexBurner.js
```diff
@@ -0,0 +1,691 @@
...
+        govTokenGovernance = await TokenGovernance.new(govToken.address);
+        await govToken.transferOwnership(govTokenGovernance.address);
+        await govTokenGovernance.acceptTokenOwnership();
+
+        await networkToken.issue(owner, TOTAL_SUPPLY);
+
+        vortex = await VortexBurner.new(
+            networkToken.address,
+            govTokenGovernance.address,
+            new BN(0),
```
barakman, line 98: We can use native JS numbers when the values that we want to pass are sufficiently small.
I'm not saying that we should use it for every `x <= Number.MAX_SAFE_INTEGER`, but I believe that it is pretty standard when passing `0`.

solidity/contracts/vortex/VortexBurner.sol
lbeder: I don't mind either way. I'll change it

barakman (changes requested):

solidity/contracts/vortex/VortexBurner.sol
```diff
@@ -0,0 +1,523 @@
...
+            }
+
+            // avoid empty conversions
+            if (amount == 0) {
+                continue;
+            }
+
+            // increment the count of the elements in the strategy
+            count++;
+
```
barakman, line 373: got it

barakman (changes requested):

solidity/contracts/vortex/VortexBurner.sol
```diff
@@ -0,0 +1,523 @@
...
+            amounts[i] = amount;
```
barakman, line 374: Shouldn't this be `amounts[count]` though (and before you increment `count`)?

solidity/contracts/vortex/VortexBurner.sol
```diff
@@ -0,0 +1,523 @@
...
+            path[0] = address(token);
+
+            // don't look up for a converter for either the network or the governance token, since they are going to be
+            // handled in a special way during the burn itself
+            if (token != _networkToken && token != _govToken) {
+                path[1] = address(networkTokenConverterAnchor(token, registry));
+                path[2] = address(_networkToken);
+            }
+
+            paths[i] = path;
```
barakman, line 386: Same comment as above (`i` vs `count`).

barakman (changes requested):

solidity/test/VortexBurner.js
lbeder: `ONE_TOKEN` makes the code below less readable, IMHO.

solidity/test/VortexBurner.js
lbeder: We should reduce the amount of "energy" we spend when writing/reading code, so I'd prefer to keep it `BN` everywhere instead of thinking where should we revert to plain numbers

solidity/contracts/vortex/VortexBurner.sol
lbeder: Yes, you're right

solidity/contracts/vortex/VortexBurner.sol
lbeder: Yes, you're right

solidity/contracts/vortex/VortexBurner.sol
```diff
@@ -0,0 +1,517 @@
...
+    uint16 private constant STANDARD_POOL_CONVERTER_TYPE = 3;
+
+    // the address of the network token
+    IERC20 private immutable _networkToken;
+
+    // the address of the governance token
+    IERC20 private immutable _govToken;
+
+    // the address of the governance token security module
+    ITokenGovernance public immutable _govTokenGovernance;
```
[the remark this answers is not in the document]
lbeder: Maybe something like a `Network.sol` or `System.sol`. Can you open a task for it?

solidity/contracts/liquidity-protection/LiquidityProtection.sol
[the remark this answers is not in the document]
yudilevi: if the goal here is to have a more specific comment, it should probably say "cannot currently receive ETH" and not "no longer", as the latter implies that it's a permanent change that prevents the contract from receiving after a certain condition is met

solidity/contracts/utility/Utils.sol
```diff
@@ -53,4 +53,15 @@ contract Utils {
     function _validExternalAddress(address _address) internal view {
         require(_address != address(0) && _address != address(this), "ERR_INVALID_EXTERNAL_ADDRESS");
     }
+
+    // ensures that the fee is valid
+    modifier validFee(uint32 fee) {
```
yudilevi, line 58: Utils is a generic contract with no bancor specific logic. This is a deviation from that path.
Can this function be defined such as that it's generic?

solidity/contracts/vortex/VortexBurner.sol
```diff
@@ -0,0 +1,530 @@
...
+
+    // the percentage of the converted network tokens to be sent to the caller of the vortex (in units of PPM)
```
yudilevi, line 56: generally speaking, `vortex` is the name of the mechanism, `burner` is the name of this contract, `burning event` is the actual event/action triggered by the user (might be a better name for this one). it makes the comments/terminology in a bunch of places in the contract a bit confusing ("completed vortex")

solidity/contracts/vortex/VortexBurner.sol
```diff
@@ -0,0 +1,530 @@
...
+     *
+     * @param tokens the converted tokens
+     * @param amounts the amounts of the converted tokens
+     * @param conversionAmounts the network token amounts the tokens were converted to
+     * @param totalBurnedAmount the total burned amount in this vortex
+     */
+    event Burned(IERC20[] tokens, uint256[] amounts, uint256[] conversionAmounts, uint256 totalBurnedAmount);
+
+    /**
+     * @dev initializes a new VortexContract contract
```
yudilevi, line 91: wrong contract name

solidity/contracts/vortex/VortexBurner.sol
```diff
@@ -0,0 +1,530 @@
...
+     *
+     * @param networkToken the address of the network token
+     * @param govTokenGovernance the address of the governance token security module
+     * @param burnIncentiveFee the percentage of the converted network tokens to be sent to the caller of the vortex (in units of PPM)
+     * @param maxBurnIncentiveFeeAmount the maximum incentive fee to be sent to the caller of the vortex
+     * @param registry the address of the contract registry
+     */
+    constructor(
```
yudilevi, line 99: I'd separate the settings from the construction

solidity/contracts/vortex/VortexBurner.sol
```diff
@@ -0,0 +1,530 @@
...
+            address[] memory path = strategy.paths[i];
+            IERC20 token = IERC20(path[0]);
+            uint256 value = 0;
+
+            if (token == _govToken) {
+                // if the source token is the governance token, don't try to convert it either, but rather include it in
+                // the amount to burn
+                //
+                // Please note that networkTokenConversionAmounts[i] will remain 0 in this case
+                totalGovTokenAmountToBurn = totalGovTokenAmountToBurn.add(amount);
```
yudilevi, line 202: why do we need to track this? can't we just use the balanceOf(burner) before burning?

solidity/contracts/vortex/VortexBurner.sol
```diff
@@ -0,0 +1,530 @@
...
+            } else if (token == _networkToken) {
+                // if the source token is the network token, don't try to convert it, but rather add its amount to the
+                // total amount to convert to the governance token
+                grossNetworkTokenConversionAmount = grossNetworkTokenConversionAmount.add(amount);
```
yudilevi, line 206: why do we need to track this? can't we just use the balanceOf(burner) before swapping to the gov token?
that will also have the benefit of including tokens that were sent here by mistake.

solidity/contracts/vortex/VortexBurner.sol
```diff
@@ -0,0 +1,530 @@
...
+
+                networkTokenConversionAmounts[i] = amount;
+            } else {
+                if (token != NATIVE_TOKEN_ADDRESS) {
+                    // if the source token is a regular token, approve the converter to withdraw the token amount
+                    ensureAllowance(token, network, amount);
```
yudilevi, line 212: technically we could do without the amount here either and just use balanceOf(burner) here as well. the overhead here will be a bit more gas, the benefit here is that we'll also swap tokens that were sent to the burner by mistake. thoughts?

solidity/contracts/vortex/VortexBurner.sol
```diff
@@ -0,0 +1,530 @@
...
+                } else {
+                    // if the source token is actually an ETH reserve, make sure to pass its value to the network
+                    value = amount;
+                }
+
+                // perform the actual conversion and optionally send ETH to the network
+                uint256 networkTokenConversionAmount =
+                    network.convertByPath2{ value: value }(path, amount, 1, address(this));
```
yudilevi, line 220: pardon my ignorance but what is `convertByPath2`? :) I don't see it in the network

solidity/contracts/vortex/VortexBurner.sol
```diff
@@ -0,0 +1,530 @@
...
+
+    /**
+     * @dev triggered after a completed vortex
+     *
+     * @param tokens the converted tokens
+     * @param amounts the amounts of the converted tokens
+     * @param conversionAmounts the network token amounts the tokens were converted to
+     * @param totalBurnedAmount the total burned amount in this vortex
+     */
+    event Burned(IERC20[] tokens, uint256[] amounts, uint256[] conversionAmounts, uint256 totalBurnedAmount);
```
yudilevi, line 88: did you consider a separate event per burning? it will allow indexing, filtering by a specific token etc. and will simplify the event

solidity/contracts/vortex/VortexBurner.sol
```diff
@@ -0,0 +1,530 @@
...
+     * @dev returns the result of the conversion the provided tokens to the governance token and its burn
+     *
+     * @param tokens the tokens to convert
+     *
+     * @return the amounts of the converted tokens
+     * @return the network token amounts the tokens were converted to
+     * @return the total burned amount in this vortex
+     * @return the incentive fee resulting from this vortex
+     */
+    function availableVortex(IERC20[] calldata tokens)
```
yudilevi, line 262: I don't quite understand the purpose of this function.
it's easy to see which tokens are available on etherscan as well as a good estimation of the the total $ value.
does it serve a specific need of the app? like showing how many gov tokens are available for burning?
not even sure we need to ever include the burning function in the app, it seems to me like a function shows all the arb opportunities... these actions are performed on chain and there's an incentive for it.

yudilevi (changes requested):

solidity/contracts/NetworkSettings.sol
```diff
@@ -12,20 +12,9 @@ import "./utility/Utils.sol";
  * (in PPM units) taken from all conversion fees accumulated in the network.
  */
 contract NetworkSettings is INetworkSettings, Owned, Utils {
-    address private _networkFeeWallet;
+    address payable private _networkFeeWallet;
```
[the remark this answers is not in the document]
lbeder: That's a good idea. I'll update it.

solidity/contracts/utility/Utils.sol
lbeder: It is somewhat generic (fee is not a Bancor-specific term after all), but a better direction can be to introduce something like `TokenUtils` and move this and some other shared methods (such as a ``NATIVE_TOKEN_ADDRESS` aware transfer of balance retrieval methods) there.

solidity/contracts/vortex/VortexBurner.sol
lbeder: I'll rename

solidity/contracts/vortex/VortexBurner.sol
lbeder: 👍

solidity/contracts/vortex/VortexBurner.sol
lbeder: Do you think it'd be cleaner to implicitly track it via balanceOf? I'm neutral.

solidity/contracts/vortex/VortexBurner.sol
lbeder: Do you think it'd be cleaner to implicitly track it via balanceOf? I'm neutral.

solidity/contracts/vortex/VortexBurner.sol
lbeder: See my comment above (i.e., let's either fix all of them or neither?)

solidity/contracts/vortex/VortexBurner.sol
lbeder: This is what the product team requested, so I'm not sure if a per-burning event would've been better here or not

solidity/contracts/vortex/VortexBurner.sol
lbeder: This function was also explicitly requested by the product team as well, so if it's not needed - I'd love to remove it...

solidity/contracts/vortex/VortexBurner.sol
lbeder: We'll:
1. Use `convertByPath` instead
2. Add a missing `IBancorNetwork` interface so that the compiler enforces it

solidity/contracts/utility/Utils.sol
yudilevi: We can also just treat it as bancor utils, practically that's what it is I guess, better than having 2 separate contracts

solidity/contracts/vortex/VortexBurner.sol
yudilevi: I think the `vortex` function also needs a different name. maybe `burn`?

solidity/contracts/vortex/VortexBurner.sol
yudilevi: less stuff to pass around, less room for errors

solidity/contracts/vortex/VortexBurner.sol
yudilevi: I think it's more flexible and better suited for ingestion by scripts

solidity/contracts/vortex/VortexBurner.sol
yudilevi: ok, let's discuss this one offline with product

solidity/contracts/utility/Utils.sol
lbeder: 👍

solidity/contracts/vortex/VortexBurner.sol
lbeder: 👍

solidity/contracts/vortex/VortexBurner.sol
lbeder: In this case, I'd rather leave it as is, since the `amount` is already a result of a different `balanceOf` and this will generate another external call for every token

solidity/contracts/vortex/VortexBurner.sol
lbeder: 👍

solidity/contracts/liquidity-protection/LiquidityProtection.sol
```diff
@@ -114,7 +114,7 @@ contract LiquidityProtection is ILiquidityProtection, Utils, Owned, ReentrancyGu
         govToken = IERC20(address(ITokenGovernance(_contractAddresses[6]).token()));
     }
 
-    // ensures that the contract is currently removing liquidity from a converter
+    // ensures that the contract is currently removing liquidity from a converter and can accept ETH
```
barakman, line 117: Currently **adding or** removing liquidity, no?

solidity/contracts/vortex/VortexBurner.sol
```diff
@@ -0,0 +1,529 @@
...
+    struct NetNetworkTokenConversionAmounts {
+        uint256 amount;
+        uint256 incentiveFeeAmount;
+    }
+
+    // the mechanism is only designed to work with 50/50 standard pool converters
+    uint32 private constant STANDARD_POOL_RESERVE_WEIGHT = PPM_RESOLUTION / 2;
+
+    // the type of the standard pool converter
+    uint16 private constant STANDARD_POOL_CONVERTER_TYPE = 3;
```
barakman, line 45: Maybe declare it `uint16 private immutable standardPoolConverterType`...
And then initialize it in the constructor using `StandardPoolConverterFactory.converterType()`.
?

solidity/contracts/vortex/VortexBurner.sol
```diff
@@ -0,0 +1,529 @@
...
+                // Please note that networkTokenConversionAmounts[i] will remain 0 in this case
+                totalGovTokenAmountToBurn = totalGovTokenAmountToBurn.add(amount);
+            } else if (token == _networkToken) {
+                // if the source token is the network token, don't try to convert it, but rather add its amount to the
+                // total amount to convert to the governance token
+                grossNetworkTokenConversionAmount = grossNetworkTokenConversionAmount.add(amount);
+
+                networkTokenConversionAmounts[i] = amount;
+            } else {
+                if (token != NATIVE_TOKEN_ADDRESS) {
```
barakman, line 202: You can flatten that last if/else here (i.e., `else if` and then `else`).

solidity/contracts/vortex/VortexBurner.sol
```diff
@@ -0,0 +1,529 @@
...
+                    // if the source token is a regular token, approve the converter to withdraw the token amount
+                    ensureAllowance(token, network, amount);
+                } else {
+                    // if the source token is actually an ETH reserve, make sure to pass its value to the network
+                    value = amount;
+                }
+
+                // perform the actual conversion and optionally send ETH to the network
+                uint256 networkTokenConversionAmount =
+                    network.convertByPath{ value: value }(path, amount, 1, address(this), address(0), 0);
```
barakman, line 212: You can use `convertByPath2` here (`convertByPath` is supposed to be deprecated at this point).

barakman (changes requested):

solidity/contracts/vortex/VortexBurner.sol
yudilevi: so we decided to remove this one

solidity/contracts/vortex/VortexBurner.sol
yudilevi: convertByPath2 isn't deployed yet (and not scheduled to be deployed anytime soon).

solidity/contracts/liquidity-protection/LiquidityProtection.sol
lbeder: 👍

solidity/contracts/vortex/VortexBurner.sol
lbeder: It's a cool idea, but it's currently not possible to call "static functions" by the contract name, but only via an instance

solidity/contracts/vortex/VortexBurner.sol
lbeder: This was already changed, so please take another look over the new code

lbeder: Also, please note that flattening it would also break the logic since the following lines should execute regardless of the result of this check.

solidity/contracts/vortex/VortexBurner.sol
barakman: Of course, but I didn't even mean to do it this way.
We can get the factory instance from the registry.

lbeder: Doesn't worth it

solidity/contracts/vortex/VortexBurner.sol
```diff
@@ -0,0 +1,421 @@
...
+    );
+
+    /**
+     * @dev triggered during conversion of a single token during the burning event
+     *
+     * @param token the converted token
+     * @param amount the amount of the converted token
+     * @param conversionAmount the network token amount the token were converted to
+     */
+    event Converted(IERC20 token, uint256 amount, uint256 conversionAmount);
```
yudilevi, line 87: sourceAmount/targetAmount?

solidity/contracts/vortex/VortexBurner.sol
```diff
@@ -0,0 +1,421 @@
...
+
+    /**
+     * @dev triggered after a completed burning event
+     *
+     * @param tokens the converted tokens
+     * @param conversionAmount the total network token amounts the tokens were converted to
+     * @param burntAmount the total burned amount in this burning event
+     */
+    event Burned(IERC20[] tokens, uint256 conversionAmount, uint256 burntAmount);
```
yudilevi, line 96: conversionAmount -> sourceAmount?

solidity/contracts/vortex/VortexBurner.sol
```diff
@@ -0,0 +1,421 @@
...
+            if (token != NATIVE_TOKEN_ADDRESS) {
+                // if the source token is a regular token, approve the converter to withdraw the token amount
+                ensureAllowance(token, network, amount);
+            } else {
+                // if the source token is actually an ETH reserve, make sure to pass its value to the network
+                value = amount;
+            }
+
+            // perform the actual conversion and optionally send ETH to the network
+            uint256 networkTokenConversionAmount =
```
yudilevi, line 208: networkTokenConversionAmount -> targetAmount?

solidity/contracts/vortex/VortexBurner.sol
```diff
@@ -0,0 +1,421 @@
...
+    using Math for uint256;
+    using SafeERC20 for IERC20;
+
+    struct Strategy {
+        address[][] paths;
+        uint256[] amounts;
+        address[] govPath;
+    }
+
+    struct NetNetworkTokenConversionAmounts {
```
yudilevi, line 36: seems a bit excessive to have this struct now that the `availableForBurning` function is removed, no?

solidity/contracts/vortex/VortexBurner.sol
```diff
@@ -0,0 +1,421 @@
...
+            IERC20 token = IERC20(path[0]);
+            uint256 value = 0;
+
+            if (token == _networkToken || token == _govToken) {
+                // if the source token is the network or the governance token, we won't try to convert it, but rather
+                // include its amount in either the total amount of tokens to convert or burn.
+                continue;
+            }
+
+            if (token != NATIVE_TOKEN_ADDRESS) {
```
yudilevi, line 199: maybe flip the condition/blocks here so that we'll use equality and not inequality?

solidity/contracts/vortex/VortexBurner.sol
```diff
@@ -0,0 +1,421 @@
...
+            uint256 networkTokenConversionAmount =
+                network.convertByPath{ value: value }(path, amount, 1, address(this), address(0), 0);
+
+            emit Converted(token, amount, networkTokenConversionAmount);
+        }
+
+        // calculate the burn incentive fee and reduce it from the total amount to convert
+        NetNetworkTokenConversionAmounts memory netNetworkTokenConversionAmounts = netNetworkConversionAmounts();
+
+        // approve the governance token converter to withdraw the network token amount
```
yudilevi, line 217: wrong comment (converter instead of network)

solidity/contracts/vortex/VortexBurner.sol
```diff
@@ -0,0 +1,421 @@
...
+        network.convertByPath(
+            strategy.govPath,
+            netNetworkTokenConversionAmounts.amount,
+            1,
+            address(this),
+            address(0),
+            0
+        );
+
+        uint256 totalGovTokenAmountToBurn = _govToken.balanceOf(address(this));
```
yudilevi, line 230: totalGovTokenAmountToBurn -> govTokenBalance?

solidity/contracts/vortex/VortexBurner.sol
lbeder: 👍

solidity/contracts/vortex/VortexBurner.sol
lbeder: 👍

solidity/contracts/vortex/VortexBurner.sol
lbeder: 👍

solidity/contracts/vortex/VortexBurner.sol
lbeder: I agree. I'll refactor

solidity/contracts/vortex/VortexBurner.sol
lbeder: 👍

solidity/contracts/vortex/VortexBurner.sol
lbeder: 👍

solidity/contracts/vortex/VortexBurner.sol
lbeder: 👍

solidity/contracts/vortex/VortexBurner.sol
yudilevi: still wrong :) "governance token network"

lbeder: 3a116884945f870924f1ffd3f36fc015

solidity/contracts/liquidity-protection/LiquidityProtection.sol
```diff
@@ -114,7 +114,7 @@ contract LiquidityProtection is ILiquidityProtection, Utils, Owned, ReentrancyGu
         govToken = IERC20(address(ITokenGovernance(_contractAddresses[6]).token()));
     }
 
-    // ensures that the contract is currently removing liquidity from a converter
+    // ensures that the contract is currently adding or removing liquidity from a converter and can accept ETH
```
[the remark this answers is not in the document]
lbeder: Done and done

solidity/utils/README.md
```diff
@@ -114,8 +114,7 @@ Here is an example of the initial configuration file which should be provided to
         }
     ],
     "networkSettingsParams": {
```
barakman, line 116: Should remove the entire `networkSettingsParams` field (not just the `networkFee` field).

barakman (changes requested):

solidity/utils/README.md
lbeder: Done

yudilevi (approved):

PR #572, opened 2021-03-30
Fix function `addLiquidityReturn`

barakman:


solidity/test/LiquidityProtection.js
```diff
@@ -98,7 +98,7 @@ describe('LiquidityProtection', () => {
                     await baseToken.approve(converter.address, RESERVE1_AMOUNT);
                 }
 
-                await converter.addLiquidity(
+                await converter.methods['addLiquidity(address[],uint256[],uint256)'](
```
yudilevi, line 101: we can delete all the "new" non array function overloads and clean this up

yudilevi (changes requested):

solidity/test/LiquidityProtection.js
lbeder: @barakman in any case, assuming that you'd still have to do it - DRY it. There is no reason to have all these selectors repeated all the time.

lbeder (changes requested):

PR #574, opened 2021-04-01
Fix sp add liquidity

barakman:


solidity/contracts/converter/types/standard-pool/StandardPoolConverter.sol
```diff
@@ -757,13 +757,17 @@ contract StandardPoolConverter is ConverterVersion, IConverter, ContractRegistry
...
 
         uint256 amount;
+        uint256[2] memory reserveAmounts;
 
         // calculate the amount of pool tokens to mint for the caller
         // and the amount of reserve tokens to transfer from the caller
         if (totalSupply == 0) {
             amount = MathEx.geometricMean(_reserveAmounts);
+            for (uint256 i = 0; i < 2; i++) {
+                reserveAmounts[i] = _reserveAmounts[i];
```
lbeder, line 767: We don't need a for loop here

solidity/contracts/converter/types/standard-pool/StandardPoolConverter.sol
```diff
@@ -774,8 +778,9 @@ contract StandardPoolConverter is ConverterVersion, IConverter, ContractRegistry
         uint256 newPoolTokenSupply = totalSupply.add(amount);
         for (uint256 i = 0; i < 2; i++) {
             IERC20 reserveToken = _reserveTokens[i];
-            uint256 reserveAmount = _reserveAmounts[i];
+            uint256 reserveAmount = reserveAmounts[i];
             require(reserveAmount > 0, "ERR_ZERO_TARGET_AMOUNT");
+            assert(reserveAmount <= _reserveAmounts[i]);
```
lbeder, line 783: Why do we need an `assert` here? If looks more like a `require`, but in any case this case can be handled via SafeMath below

lbeder (changes requested):
Please see my comments

solidity/contracts/converter/types/standard-pool/StandardPoolConverter.sol
barakman: I reverted to how it was before, but NP.

solidity/contracts/converter/types/standard-pool/StandardPoolConverter.sol
barakman: Not a `require`. The function that creates the new `reserveAmounts` array must return amounts which are at most the input amounts. If it doesn't then we have a bug in the code (i.e., this is not a user-input dependent condition, hence does need require a `require`).

solidity/contracts/converter/types/standard-pool/StandardPoolConverter.sol

solidity/contracts/converter/types/standard-pool/StandardPoolConverter.sol
lbeder: I'd still use SafeMath below and avoid this assert altogether. Even if this can happen (and I'm not sure) - we would still want a descriptive error message

barakman: 'reverted with subtraction overflow' is much more common than 'invalid opcode', so this would actually yield a less descriptive (and harder to analyze) error-message; are you sure?

lbeder: Yes. `invalid opcode` can be so many other things.

yudilevi (approved):

lbeder (approved):

PR #577, opened 2021-04-04
Update the product of reserve balances at the end of converter upgrade

barakman:


solidity/contracts/converter/ConverterBase.sol
```diff
@@ -258,6 +258,18 @@ abstract contract ConverterBase is ConverterVersion, IConverter, ContractRegistr
         acceptOwnership();
     }
 
+    /**
+     * @dev does nothing
+     */
+    function postUpgrade()
```
yudilevi, line 264: -> `onUpgradeComplete`

solidity/contracts/converter/ConverterBase.sol
```diff
@@ -258,6 +258,18 @@ abstract contract ConverterBase is ConverterVersion, IConverter, ContractRegistr
         acceptOwnership();
     }
 
+    /**
+     * @dev does nothing
```
yudilevi, line 262: "executed by the upgrader at the end of the upgrade process to handle and custom pool logic"

yudilevi (changes requested):

solidity/contracts/converter/ConverterBase.sol
barakman: And in the StandardPoolConverter?

solidity/contracts/converter/ConverterBase.sol
barakman: Done

solidity/contracts/converter/ConverterBase.sol
yudilevi: oops, had a typo there ("and")

yudilevi: and can use it anywhere

barakman: Done

yudilevi (approved):

PR #582, opened 2021-04-07
Add missing curly braces in conditional statements

barakman:


lbeder (approved):
Can you double-check that the code is formatted?

barakman:
> Can you double-check that the code is formatted?

done

lbeder (approved):
👍

PR #588, opened 2021-04-11
Verify rate deviation at an earlier point

barakman:
Verify average rate / spot rate deviation at the beginning of add/remove protected liquidity.


solidity/contracts/liquidity-protection/LiquidityProtection.sol
```diff
@@ -897,6 +902,15 @@ contract LiquidityProtection is ILiquidityProtection, Utils, Owned, ReentrancyGu
         uint256 poolAmount,
         uint256 reserveAmount
     ) internal returns (uint256) {
+        (
+            uint256 spotRateN,
+            uint256 spotRateD,
+            uint256 averageRateN,
+            uint256 averageRateD
+        ) = reserveTokenRates(poolToken, reserveToken);
```
lbeder, line 910: Perhaps `reserveTokenRates` can return a struct which `verifyRate` can then verify, which can be also refactored to `PackedRates`? It can be also a good opportunity to switch to `Fraction` (and `Fraction128` for `PackedRates`) so that we won't keep carrying around all these variables.

lbeder: We can also have an overload for `verifyRate` that received `PackedRates` as well

lbeder (changes requested):

solidity/contracts/liquidity-protection/LiquidityProtection.sol
barakman: done

lbeder (approved):

solidity/contracts/liquidity-protection/LiquidityProtection.sol
```diff
@@ -615,6 +614,22 @@ contract LiquidityProtection is ILiquidityProtection, Utils, Owned, ReentrancyGu
...
+        // get the various rates between the reserves upon adding liquidity and now
+        PackedRates memory packedRates =
+            packRates(
+                liquidity.poolToken,
+                liquidity.reserveToken,
+                liquidity.reserveRateN,
+                liquidity.reserveRateD
+            );
+
+        verifyRate(
```
yudilevi, line 626: cleaner implementation :) I'd just add a comment both here and in addLiquidity that explains why we verify the deviation so early

yudilevi (changes requested):

solidity/contracts/liquidity-protection/LiquidityProtection.sol
barakman: done

lbeder (approved):

yudilevi (approved):

PR #594, opened 2021-04-17
Allow transferring position with an optional external notification callback

lbeder:
This PR adds the `transferPositionAndCall` function, allowing the caller to specify a custom transfer position notification callback.

### DIFF
--- a/solidity/contracts/liquidity-protection/LiquidityProtection.sol
+++ b/solidity/contracts/liquidity-protection/LiquidityProtection.sol
@@ -783,11 +783,11 @@ contract LiquidityProtection is ILiquidityProtection, Utils, Owned, ReentrancyGu
         address newProvider,
         address target,
         bytes memory data
-    ) external protected validAddress(newProvider) returns (uint256) {
+    ) external protected validAddress(newProvider) validAddress(target) returns (uint256) {
         uint256 newId = transferPosition(msg.sender, id, newProvider);
 
         // make sure that we're not trying to call into the zero address or a fallback function
-        require(target != address(0) && data.length >= FUNC_SELECTOR_LENGTH, "ERR_INVALID_CALL_DATA");
+        require(data.length >= FUNC_SELECTOR_LENGTH, "ERR_INVALID_CALL_DATA");
 
         Address.functionCall(target, data, "ERR_CALL_FAILED");
 
--- a/solidity/contracts/liquidity-protection/LiquidityProtection.sol
+++ b/solidity/contracts/liquidity-protection/LiquidityProtection.sol
@@ -2,6 +2,7 @@
 pragma solidity 0.6.12;
 
 import "@openzeppelin/contracts/math/SafeMath.sol";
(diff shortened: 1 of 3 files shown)


solidity/contracts/liquidity-protection/LiquidityProtection.sol
```diff
@@ -763,8 +764,53 @@ contract LiquidityProtection is ILiquidityProtection, Utils, Owned, ReentrancyGu
...
+    function transferPositionAndCall(
+        uint256 id,
+        address newProvider,
+        address target,
+        bytes memory data
+    ) external protected validAddress(newProvider) returns (uint256) {
+        uint256 newId = transferPosition(msg.sender, id, newProvider);
+
+        // make sure that we're not trying to call into the zero address or a fallback function
+        require(target != address(0) && data.length > 0, "ERR_INVALID_CALL_DATA");
```
barakman, line 789: Alternative 1:
```
        validAddress(target);
        ...
        require(data.length > 0, "ERR_INVALID_CALL_DATA");
```

Alternative 2:
```
        require(target != address(0), "ERR_INVALID_CALL_TARGET");
        require(data.length > 0, "ERR_INVALID_CALL_DATA");
```

barakman: Also, it's possibly better to check `data.length >= 4` (length of function selector), but I'm not sure about this.

barakman (changes requested):

solidity/contracts/liquidity-protection/LiquidityProtection.sol
lbeder: Regarding the alternatives, I'm not sure there's any benefit in separating this test into two tests. I will extend the `data.length`, but keep in mind that we might need to remove it entirely if we'll decide to support fallback functions.

barakman (approved):

barakman (approved):

solidity/contracts/liquidity-protection/LiquidityProtection.sol
```diff
@@ -763,8 +765,53 @@ contract LiquidityProtection is ILiquidityProtection, Utils, Owned, ReentrancyGu
...
+    function transferPositionAndCall(
+        uint256 id,
+        address newProvider,
+        address target,
+        bytes memory data
+    ) external protected validAddress(newProvider) returns (uint256) {
+        uint256 newId = transferPosition(msg.sender, id, newProvider);
+
+        // make sure that we're not trying to call into the zero address or a fallback function
+        require(target != address(0) && data.length >= FUNC_SELECTOR_LENGTH, "ERR_INVALID_CALL_DATA");
```
yudilevi, line 790: any reason not to call `validAddress` on the target instead?

lbeder: @barakman mentioned it as well, so I'll just fix it 😎

yudilevi (approved):

PR #599, opened 2021-04-23
Fix BancorNetwork specific conversion path tests

lbeder:


solidity/test/BancorNetwork.js
```diff
@@ -40,79 +40,49 @@ Token network structure:
...
-            },
-            ERC3: {
-                ETH: [erc20Token3, anchor4, bntToken, anchor1, ''],
-                BNT: [erc20Token3, anchor4, bntToken],
-                ERC1: [erc20Token3, anchor4, bntToken, anchor2, erc20Token1],
-                ERC2: [erc20Token3, anchor4, bntToken, anchor3, erc20Token2]
-            }
+    const initTokens = async () => {
+        tokens = {
+            ETH: { address: NATIVE_TOKEN_ADDRESS },
```
barakman, line 45: Since we have this, I believe that we no longer need stuff like `reserveToken.address || reserveToken`, `typeof reserveToken === 'string'`, etc.

barakman (changes requested):

solidity/test/BancorNetwork.js
lbeder: I preferred to keep the "generic" `getBalance` function here as well so that it'd be easier to refactor it everywhere later on (which is one of our future tasks).

yudilevi (approved):
I didn't go over it in detail but we're phasing out the paths anyway so I assume most of this isn't gonna be deprecated.

lbeder:
@yudilevi that's right. Most of these changes will have to be adapted for v3.0.0.

PR #603, opened 2021-04-28
Add the ABIs to the NPM package

barakman:


lbeder (approved):
@sshmaxime please note that this would need to be adapted differently for our upcoming hardhat merge.

@barakman please make sure to update `3.0.0` with this as well.

yudilevi (commented):
would it be possible to generate these files during `npm publish`, so that we don't need to remember to do that step before publishing?

sshmaxime:
> @sshmaxime please note that this would need to be adapted differently for our upcoming hardhat merge.

On it

PR #604, opened 2021-04-29
Update terminology and style BancorNetwork

lbeder:


solidity/contracts/converter/interfaces/IConverter.sol
```diff
@@ -93,16 +93,16 @@ interface IConverter is IOwned {
...
      * @param conversionFee conversion fee in units of the target token
      */
     event Conversion(
-        IReserveToken indexed fromToken,
-        IReserveToken indexed toToken,
+        IReserveToken indexed sourceToken,
+        IReserveToken indexed targetToken,
         address indexed trader,
         uint256 amount,
         uint256 returnAmount,
```
barakman, line 108: Should also rename `amount` to `sourceAmount` and `returnAmount` to `targetAmount` IMO.
Another option is `inputAmount` and `outputAmount`.
It should then also be applied in function `dispatchConversionEvent` in the standard-pool converter.

solidity/contracts/BancorNetwork.sol
```diff
@@ -64,20 +64,20 @@ contract BancorNetwork is IBancorNetwork, TokenHolder, ContractRegistryClient, R
...
-        IReserveToken indexed _toToken,
-        uint256 _fromAmount,
-        uint256 _toAmount,
-        address _trader
+        IConverterAnchor indexed anchor,
+        IReserveToken indexed sourceToken,
+        IReserveToken indexed targetToken,
+        uint256 sourceAmount,
+        uint256 returnAmount,
+        address trader
```
barakman, line 80: Should also rename `returnAmount` to `targetAmount` IMO.

solidity/contracts/converter/interfaces/IConverter.sol
barakman: Also, in function `doConversion` in the standard-pool converter, we are still using local variables named `fromAmount` and `toAmount `, which we probably want to rename to `sourceAmount` and `targetAmount` as part of this task.

barakman (changes requested):

solidity/test/StandardPoolConverter.js
```diff
@@ -387,11 +387,11 @@ describe('StandardPoolConverter', () => {
...
-                        _fromToken: reserveToken1.address,
-                        _toToken: reserveToken2.address,
-                        _fromAmount: amount,
-                        _toAmount: purchaseAmount
+                        anchor: poolToken.address,
+                        sourceToken: reserveToken1.address,
+                        targetToken: reserveToken2.address,
+                        sourceAmount: amount,
+                        returnAmount: purchaseAmount
                     });
```
barakman, line 395: If you take my suggestion and rename `returnAmount` in the contract, then don't forget to rename it here too.

barakman (changes requested):

solidity/contracts/converter/interfaces/IConverter.sol
lbeder: 👍

solidity/contracts/BancorNetwork.sol
yudilevi: +1

yudilevi (commented):
no further comments other than barak's.

yudilevi (approved):

barakman (approved):

PR #605, opened 2021-05-02
Fix the deployment script

barakman:


yudilevi (approved):

lbeder (changes requested):
@barakman `LiquidityPoolV1Converter` was already removed. You should start from `3.0.0` instead

lbeder (approved):
Ah, I see it's `0.6.37` specific. Carry on :)

PR #606, opened 2021-05-02
Fix and extend the deployment script

barakman:
1. Fix the deployment script.
2. Extend the deployment script to deploy the staking-rewards module.

### DIFF
--- a/solidity/utils/test_deployment.js
+++ b/solidity/utils/test_deployment.js
@@ -13,6 +13,9 @@ const MIN_GAS_LIMIT = 100000;
 const ROLE_OWNER = Web3.utils.keccak256('ROLE_OWNER');
 const ROLE_GOVERNOR = Web3.utils.keccak256('ROLE_GOVERNOR');
 const ROLE_MINTER = Web3.utils.keccak256('ROLE_MINTER');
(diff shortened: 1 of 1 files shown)


yudilevi (approved):

solidity/utils/test_deployment.js
```diff
@@ -354,6 +366,13 @@ const run = async () => {
 
     await execute(checkpointStore.methods.grantRole(ROLE_OWNER, liquidityProtection._address));
 
+    await execute(stakingRewardsStore.methods.grantRole(ROLE_OWNER, stakingRewards._address));
+    await execute(stakingRewardsStore.methods.grantRole(ROLE_MANAGER, account.address));
+    await execute(stakingRewards.methods.grantRole(ROLE_UPDATER, account.address));
```
lbeder, line 371: There is no need to set the `ROLE_MANAGER` or the `ROLE_UPDATER` roles during deployment

lbeder (changes requested):

solidity/utils/test_deployment.js
barakman: fixed

lbeder (approved):

PR #607, opened 2021-05-02
Remove liquid-token leftovers

barakman:


lbeder (changes requested):
Please base this PR on `3.0.0` and update it accordingly

PR #608, opened 2021-05-02
Move the main contracts from solidity/contracts to contracts and remove unneeded scripts

lbeder:


yudilevi (approved):

sshmaxime (approved):

hardhat.config.js
```diff
@@ -62,21 +62,21 @@ module.exports = {
...
+        tests: './test',
         cache: './cache',
-        artifacts: './solidity/build/contracts'
+        artifacts: './build/contracts'
     },
 
     // Test Config
     mocha: {
-        spec: argv.spec || 'solidity/test',
+        spec: argv.spec || 'test',
```
barakman, line 79: IMO, this is a good opportunity to rename folder `test` (Truffle's default) to `tests`.

hardhat.config.js
```diff
@@ -62,21 +62,21 @@ module.exports = {
...
-        path: './solidity/build/abi',
+        path: './build/abi',
         clear: true
     },
 
     // System Config
     paths: {
-        sources: './solidity/contracts',
-        tests: './solidity/test',
+        sources: './contracts',
```
barakman, line 71: IMO, this is a good opportunity to rename folder `test` (Truffle's default) to `tests`.

barakman (changes requested):

hardhat.config.js
yudilevi: we should just use the standard terminology in hardhat.

lbeder: 👍. I'm moving it to the defaults (`test` remains, and `build` is changed to `artifacts`)

lbeder:
Please note that I've reverted to use Hardhat's defaults (while moving the abi-exporter output to `data/abi`, as it can't be at the same default folder).

yudilevi (approved):

barakman (approved):

sshmaxime (approved):

PR #609, opened 2021-05-03
Remove liquid-token leftovers

barakman:


solidity/contracts/converter/ConverterUpgrader.sol
```diff
@@ -163,8 +163,8 @@ contract ConverterUpgrader is IConverterUpgrader, ContractRegistryClient {
         // new converter - get the type from the converter itself
         if (isV28OrHigherConverter(_oldConverter)) {
             newType = _oldConverter.converterType();
-        } else if (reserveTokenCount > 1) {
-            // old converter - if it has 1 reserve token, the type is a liquid token, otherwise the type liquidity pool
+        } else {
+            assert(reserveTokenCount > 1);
```
barakman, line 167: Note: since we no longer support type 1 anyway, we can technically remove this entire section, including the part that follows it (below) and handles this type.
But because this PR is about removing type 0 (liquid-token) support, we will handle this in a separate PR.

yudilevi: not only we don't support type 1 anymore, we only support 2 reserve tokens, so we might even be able to remove more code, but we can wait with that until the 2 reserve token enforcement change.

yudilevi (approved):

PR #610, opened 2021-05-04
Move to oz reentrancy guard

barakman:


barakman:
A minor downside of this is slightly larger bytecode, which means slightly higher deployment gas cost (btw, there's also a minor upside in the form of slightly lower transaction gas cost).
Of course, gas cost is not an issue here, but the possibility of bytecode size exceeding the maximum limit (24kb).
And at least currently, it not the case in either one of the contracts involved in this PR.

lbeder (approved):
👍 👍 👍 👍 👍

yudilevi (approved):
oh, that's a fun cleanup :)

PR #611, opened 2021-05-09
Restructure the tests according to the structure of the contracts

lbeder:


test/helpers/BancorFormula.js
```diff
@@ -7,8 +7,8 @@ const { ContractFactory } = require('ethers');
 module.exports.new = async () => {
     const accounts = await ethers.getSigners();
 
-    const abi = fs.readFileSync(path.resolve(__dirname, '../bin/BancorFormula.abi'));
-    const bin = fs.readFileSync(path.resolve(__dirname, '../bin/BancorFormula.bin'));
+    const abi = fs.readFileSync(path.resolve(__dirname, './bin/BancorFormula.abi'));
```
barakman, line 10: What is this file still doing here?

barakman (approved):

test/helpers/BancorFormula.js
yudilevi: probably need to get rid of that

yudilevi (approved):

test/helpers/BancorFormula.js
lbeder: We're using this to test the upgrade from a legacy converter. Why should we remove it?

PR #622, opened 2021-05-18
Remove the whitelist contract

lbeder:


barakman (changes requested):
You forgot to remove it in test_deployment.js.

lbeder:
@barakman I don't see it there. Where is it?

barakman:
> @barakman I don't see it there. Where is it?

https://github.com/bancorprotocol/contracts-solidity/blob/3.0.0/utils/test_deployment.js#L416

lbeder:
@barakman that's a completely different whitelist... this PR removes the no longer used `Whitelist` contract

barakman:
> @barakman that's a completely different whitelist... this PR removes the no longer used `Whitelist` contract

Don't we want to remove al whitelist-related aspects in the system?

lbeder:
> > @barakman that's a completely different whitelist... this PR removes the no longer used `Whitelist` contract
> 
> Don't we want to remove all whitelist-related aspects in the system?

Not that I'm aware of, but even if it was the case - how is it related to this PR?

barakman:
> > > @barakman that's a completely different whitelist... this PR removes the no longer used `Whitelist` contract
> > 
> > 
> > Don't we want to remove all whitelist-related aspects in the system?
> 
> Not that I'm aware of, but even if it was the case - how is it related to this PR?

In the sense that we may want to consider the primary goal of this PR - whether it is the technical goal of removing a specific contract, or the conceptual goal of removing an entire aspect of the system.

lbeder:
There is only a single purpose to this PR and it's to remove a no longer used contract (a separate PR was opened to add it to the legacy repository). That's it. Current LP whitelisting functionality is implemented using the `EnumerableSet.AddressSet` primitive.

yudilevi (approved):

PR #623, opened 2021-05-19
Deployment Utility

barakman:
Extract a deployment utility from the test_deployment script, and use it separately from that script and from a hard-hat test file.


barakman:
Phase 1:
- The hard-hat test executes the Deployment utility using ethers.js functionality.
- The deployment script executes the Deployment utility using web3.js functionality.

In phase 2, we may want to consider porting the deployment script from web3.js to ethers.js.

sshmaxime:
Nice ! Might as well do the phase 2 in this PR no ? @barakman

yudilevi (approved):

test/Deployment.js
```diff
@@ -0,0 +1,101 @@
+const Contracts = require('./helpers/Contracts');
+const runDeployment = require('./helpers/runDeployment');
+
+const config =
+{
+    "reserves": [
```
lbeder, line 6: These shouldn't be quoted, since it's not a JSON file

test/Deployment.js
```diff
@@ -0,0 +1,101 @@
+const Contracts = require('./helpers/Contracts');
+const runDeployment = require('./helpers/runDeployment');
+
+const config =
+{
```
lbeder, line 5: The bracket should be in the same line. Did you format these files?

utils/test_deployment.js
```diff
@@ -2,19 +2,16 @@ const fs = require('fs');
 const path = require('path');
 const Web3 = require('web3');
 
+const runDeployment = require('../test/helpers/runDeployment');
```
lbeder, line 5: 👍 👍 👍

lbeder (changes requested):
Nice job! Please see some minor comment above

test/Deployment.js
barakman: ah yeah, copy-paste (from the README file) mistake

test/Deployment.js
barakman: ah yeah, copy-paste (from the README file) mistake

(yes, I actually did format these files afaik)

barakman:
> Nice ! Might as well do the phase 2 in this PR no ? @barakman

I am not yet familiar with signing transactions using ethers.js.
I am guessing that it might take a while to implement everything in a relatively robust manner (as robust as it is today, or more).

test/Deployment.js
barakman: fixed

test/Deployment.js
barakman: fixed

lbeder (approved):

PR #626, opened 2021-05-20
Refactoring and standardization 

lbeder:
- [x] Updated dependencies (especially hardhat to 2.3.0)
- [x] Updated the style and coding conventions of remaining contracts 
- [x]  Merged `Converter.js` tests with `StandardPoolConverter.js` and delete it entirely
- [x]  Refactored some methods to `external`
- [x]  Improved the usage of OZ's `Math` library
- [x]  Improved the usage of OZ's `ERC20` library


contracts/bancorx/BancorX.sol
```diff
@@ -34,395 +34,577 @@ contract BancorX is IBancorX, TokenHolder, ContractRegistryClient {
         bool completed;
     }
 
-    uint16 public constant version = 4;
+    uint16 public constant VERSION = 4;
```
barakman, line 37: API breaking change

contracts/bancorx/BancorX.sol
```diff
@@ -34,395 +34,577 @@ contract BancorX is IBancorX, TokenHolder, ContractRegistryClient {
...
-    mapping(address => bool) public reporters;
+    mapping(address => bool) private _reporters;
 
     /**
      * @dev triggered when tokens are locked in smart contract
      *
-     * @param _from    wallet address that the tokens are locked from
-     * @param _amount  amount locked
+     * @param from wallet address that the tokens are locked from
+     * @param amount amount locked
```
barakman, line 70: maybe 'source' instead of 'from'?

contracts/bancorx/BancorX.sol
```diff
@@ -34,395 +34,577 @@ contract BancorX is IBancorX, TokenHolder, ContractRegistryClient {
...
      */
-    event TokensLock(address indexed _from, uint256 _amount);
+    event TokensLock(address indexed from, uint256 amount);
 
     /**
      * @dev triggered when tokens are released by the smart contract
      *
-     * @param _to      wallet address that the tokens are released to
-     * @param _amount  amount released
+     * @param to wallet address that the tokens are released to
```
barakman, line 77: maybe 'target' instead of 'to'?

contracts/bancorx/BancorX.sol
```diff
@@ -34,395 +34,577 @@ contract BancorX is IBancorX, TokenHolder, ContractRegistryClient {
...
+        bytes32 toBlockchain,
+        bytes32 to,
+        uint256 amount
+    ) external xTransfersAllowed {
         // get the current lock limit
         uint256 currentLockLimit = getCurrentLockLimit();
 
         // verify lock limit
-        require(_amount >= minLimit && _amount <= currentLockLimit, "ERR_AMOUNT_TOO_HIGH");
+        require(amount >= _minLimit && amount <= currentLockLimit, "ERR_AMOUNT_TOO_HIGH");
```
barakman, line 490: suggestion: `_minLimit <= amount && amount <= currentLockLimit`

barakman: BTW, the error-message doesn't seem entire accurate for this condition

contracts/bancorx/BancorX.sol
barakman: Plus, we seem to be using `version` everywhere else

barakman (changes requested):

contracts/bancorx/BancorX.sol
lbeder: It's `BancorX`, so it's not going to be problem (nor we're going to deploy it any time soon).

lbeder: Yeah, we shouldn't have used `version` in `ConverterUpgrader` either. It's a violation of the standard naming conventions, but I've kept it there for backward compatibility, added a linter exception, and a comment.

contracts/bancorx/BancorX.sol
lbeder: Sure

contracts/bancorx/BancorX.sol
lbeder: Sure

barakman: TBH, maybe not worth the effort, so consider it accordingly...

contracts/bancorx/BancorX.sol
lbeder: I'll switch the condition, but let's keep the error message as-is for now. It's less critical, since it's BancorX

barakman (approved):

contracts/bancorx/XTransferRerouter.sol
```diff
@@ -1,54 +1,65 @@
 // SPDX-License-Identifier: SEE LICENSE IN LICENSE
 pragma solidity 0.6.12;
```
yudilevi, line 2: we can remove this contract altogether, it was never deployed

yudilevi (approved):
Now that's a big one :)

PR #635, opened 2021-05-29
Make sure that the generated bytecode doesn't include machine-specific metadata  + some cleanup

lbeder:


sshmaxime (approved):

hardhat.config.js
```diff
@@ -14,22 +14,18 @@ require('hardhat-abi-exporter');
...
     return config.apiKeys ? (config.apiKeys[apiKeyName] ? config.apiKeys[apiKeyName] : '') : '';
 };
 
-// Config
 const configNetworks = config.networks || {};
 const APIKeyEtherscan = loadAPIKey('etherscan');
 
 module.exports = {
-    // Network Config
     networks: {
```
barakman, line 28: I believe that we can get rid of the `networks` property altogether.

I've recently found that to be the case for Truffle, and I've now verified it for hardhat as well ([here](https://hardhat.org/config/#networks-configuration)).

barakman (changes requested):

hardhat.config.js
sshmaxime: fyi Some tests require this amount of eth: `accountsBalance: '10000000000000000000000000000'`. If you remove this line some test will fail - it's an easy fix

lbeder: We configure additional custom properties for the default `hardhat` network, so we can't really remove it.

barakman: I see only `gas` and `gasPrice`. The `gas` parameter is set by default to the block gas limit if I remember correctly, which should be fine as much as the tests are concerned. The `gasPrice` parameter doesn't really matter for the sake of the tests (i.e., its default configuration is as good as any other).
Am I missing anything?

barakman (changes requested):

hardhat.config.js
lbeder: Yes, you're missing the non-default ETH balance, as @sshmaxime mentioned. Also, please keep in mind that the best practice is to see a gas block limit lower than the actual mainnet limit since it can also decrease (theoretically) and you want to have a safety buffer in your tests.

barakman (approved):

yudilevi (approved):

PR #637, opened 2021-05-31
Add utility functions for computing `x * y / z` safely and accurately

barakman:
1. Add utility functions `mulDivF` and `mulDivC`, which allow to safely compute `x * y / z` (revert only if the actual result is larger than 256 bits).
2. In internal function `LiquidityProtection.protectedAmountPlusFee`, use function `mulDivF`, instead of function `reducedRatio` (which yields an approximated result).
3. Add utility function `productRatio`, which allows to compute `(x * y) / (z * z)` with high precision; we can later consider using it in order to improve the accuracy in internal function `LiquidityProtection.compensationAmount`


contracts/utility/MathEx.sol
```diff
@@ -158,4 +159,104 @@ library MathEx {
     function roundDivUnsafe(uint256 n, uint256 d) internal pure returns (uint256) {
         return (n + d / 2) / d;
     }
+
+    /**
+      * @dev Compute the largest integer smaller than or equal to `x * y / z`
+    */
+    function mulDivF(uint256 x, uint256 y, uint256 z) internal pure returns (uint256) {
```
lbeder, line 166: Could you add natspec for the input parameters and the return value? It's also relevant for some of the methods below.

contracts/utility/MathEx.sol
```diff
@@ -158,4 +159,104 @@ library MathEx {
...
+    */
+    function inv256(uint256 x) private pure returns (uint256) {
+        uint256 inv = 1;
+        for (uint256 i = 0; i < 8; i++) {
+            inv = unsafeMul(inv, unsafeSub(2, unsafeMul(inv, x)));
+        }
+        return inv;
+    }
+
+    // does not revert on overflow
```
lbeder, line 238: Could you also add a short `@dev` description as well?

contracts/utility/MathEx.sol
```diff
@@ -158,4 +159,104 @@ library MathEx {
...
+    function unsafeAdd(uint256 x, uint256 y) private pure returns (uint256) {
```
lbeder, line 239: Why do we need these `unsafeXXX` methods explicitly? Are you preparing to use `unchecked` in 0.8? 😉

test/helpers/MathUtils.js
```diff
@@ -37,6 +37,16 @@ function roundDiv(a, b) {
     return a.div(b).toFixed(0, Decimal.ROUND_HALF_UP);
 }
 
+function mulDivF(a, b, c) {
```
lbeder, line 40: Please use the `const mulDiv = (..) =>` notation, like we do in the rest of the code (other than some legacy convention here... and there).

lbeder (changes requested):

contracts/utility/MathEx.sol
barakman: Yes, but not only.

Mainly for two reasons:
1. As you've mentioned - when porting to solc 0.8, we will need to scope every piece of code which allows arithmetic overflow with `unchecked`.
2. It will make it easier for us to easily emulate this entire module in "natural" languages (for testing purposes), where overflows do not occur. In other words, we will have all the "overflowing" stuff concentrated in a single set of functions rather than being scattered all over.

For example, in Python you'd do:
```
MAX_VAL = (1 << 256) - 1

def unsafeAdd(x, y): return (x + y) & MAX_VAL
def unsafeSub(x, y): return (x - y) & MAX_VAL
def unsafeMul(x, y): return (x * y) & MAX_VAL
```

test/helpers/MathUtils.js
barakman: Done (the rest of this file in fact doesn't use `const`).

contracts/utility/MathEx.sol
barakman: Done.

contracts/utility/MathEx.sol
barakman: 1. It's an internal function, so it won't appear anywhere in our official documentation.
2. What exactly do you want me to say about the input parameters?
3. What exactly do you want me to say about the return value which isn't already mentioned in the 'dev' description?

test/helpers/MathUtils.js
barakman: It seems that the `arguments` keyword can be used inside a function, but when referred to from a function pointer (i.e., a "const" declaration of a variable which points to a function), it takes an additional item, which is the path of the file in which it is declared.
Subsequently, the function throws an exception (attempting to convert the path string to a `Decimal` object).
Will revert this change and handle it in a separate task.

contracts/utility/MathEx.sol
lbeder: I generally agree, it's just that we currently document everything and then it's harder to refactor (e.g., if an undocumented function becomes public or if we have a new public function with the same parameter, etc.)

contracts/utility/MathEx.sol
lbeder: 👍

contracts/utility/MathEx.sol
lbeder: 👍

contracts/utility/MathEx.sol
lbeder: Let's put a pin in it, while this discussion is being resolved.

test/helpers/MathUtils.js
lbeder: See my comment re: `const protectedAmountPlusFee = (...args) => {`

contracts/utility/MathEx.sol
barakman: 'put a pin' as in...? add dummy natspec for now?

test/helpers/MathUtils.js
barakman: Do you mean, use:
```
const decimalize = (func, ...args) => func(...args.map(x => Decimal(x))));
```
?

lbeder: Yeah

contracts/utility/MathEx.sol
lbeder: I meant that you can keep at as-is for now and we will revisit it later

lbeder (approved):

yudilevi (approved):

PR #638, opened 2021-05-31
Support test deployment using Ledger HW and refactor to yargs commands

lbeder:


test/Deployment.js
```diff
@@ -6,77 +8,77 @@ const config = {
...
+            supply: 0
         }
     ],
     converters: [
         {
             type: 3,
             symbol: 'ETHBNT',
             decimals: 18,
-            fee: '0.1%',
+            fee: 10_000,
```
yudilevi, line 34: makes sense for supply, but why here? why not user readable values?

utils/README.md
```diff
@@ -27,22 +34,22 @@ Here is an example of the initial configuration file which should be provided to
         {
             "symbol": "XXX",
             "decimals": 18,
-            "supply": "1829101"
+            "supply": 1829101
```
yudilevi, line 37: different format than Deployment.js - intentional?

test/Deployment.js
lbeder: I thought that it's easier to handle than a textual `0.1%`, but I can bring it back. Perhaps `0.1` (instead of `0.1%`)?

utils/README.md
lbeder: Nope :(

lbeder: It wasn't intentional, but I've tried to fix it now and realized that JSON doesn't support this :(

test/Deployment.js
yudilevi: sure, 0.1 is fine

lbeder: I've restored "%" as well so that it'd clearer that it's 0.1% and not 0.1 (10%).

barakman: What's wrong with the way that I've handled `%` in the original version of the deployment script?
```
const percentageToPPM = (value) => {
    return decimalToInteger(value.replace('%', ''), 4);
};
```

utils/README.md
barakman: The reason why I've used quotes for all amounts here is that technically they may exceed `Number.MAX_SAFE_INTEGER`.

Although with the introduction of the `decimals` field, the chances of that are slim, I figured that it would still be generally more robust to keep it that way.

Also, if I remember correctly, the function or functions which handle these amounts, expect a string and not a number, so if we're still using any of these functions then you may want to verify that our code doesn't break during runtime.

test/helpers/runDeployment.js
```diff
@@ -43,63 +45,71 @@ module.exports = async (account, deploy, deployed, execute, getConfig, keccak256
...
-            const name = reserve.symbol + ' DS Token';
-            const symbol = reserve.symbol;
-            const decimals = reserve.decimals;
-            const supply = decimalToInteger(reserve.supply, decimals);
-            const nonce = await getTransactionCount(account.address);
+            const { symbol, decimals } = reserve;
+            const name = symbol + ' DS Token';
+            const supply = toWei(reserve.supply, decimals);
             const token = await deploy('dsToken-' + symbol, 'DSToken', name, symbol, decimals);
-            if (nonce !== (await getTransactionCount(account.address))) {
```
barakman, line 74: This is of course meaningless when executed via the HH test.

But when executed via the deployment script, this check determines whether or not the previous transaction (deployment of a DS token) had executed "for real".

In other words, it is possible that the script was executed a first time, completed that transaction but failed at a later point, then executed again, in which case that DS token is not redeployed, the nonce doesn't increment, and the script knows that it should not re-issue tokens to the caller (i.e., execute the stuff inside the `if` clause).

barakman (changes requested):

test/Deployment.js
lbeder: Parsing numbers as strings as a bit buggy. For example, `percentageToPPM("0.123456%")` would have returned `123,456`

utils/README.md
lbeder: re: `Number.MAX_SAFE_INTEGER` - `BigNumber` will take care of this and in the worst case - we'd adapt. Working with numbers as strings by default is buggy (see the previous comment + it also opens a room of error when mixing hex or even non-ascii). I've refactored the script to expect numbers where numbers are due in any case.

test/helpers/runDeployment.js
```diff
@@ -43,63 +45,71 @@ module.exports = async (account, deploy, deployed, execute, getConfig, keccak256
...
+            formatBytes32String('BancorConverterRegistryData'),
+            converterRegistryData.address
+        )
     );
 
-    // initialize converter factory
     await execute(converterFactory.registerTypedConverterFactory(standardPoolConverterFactory.address));
 
-    for (const reserve of getConfig().reserves) {
+    for (const reserve of config.reserves) {
```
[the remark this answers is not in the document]
lbeder: Yes, of course. Please note that I also update the actual `config` object whenever it's used (instead of the previous approach of always rewriting and rereading it).

test/helpers/runDeployment.js
lbeder: This shouldn't have been handled via nonce heuristics, but rather via execution steps. In any case, we're going to adapt the whole deployment process to a migratory/incremental process, so this will change as well, so I'd suggest keeping the simplified version.

utils/README.md
barakman: In contrast with `BigNumber('111111111...')`, which is guaranteed to work correctly, I doubt that `BigNumber(11111111...)` is safe. The constructor function will have already received the inaccurate numeric value when called.

lbeder: It'll throw an exception and we would be able to adapt. Currently, our config files don't have these values though (in other words, let's handle the exception explicitly rather than the other way around)

test/Deployment.js
```diff
@@ -6,22 +8,22 @@ const config = {
         {
             symbol: 'XXX',
             decimals: 18,
-            supply: '1829101'
+            supply: 1_829_101
```
yudilevi, line 11: come to think of it, what type is this in json?

lbeder: It needs to be a number (1829101) or a string ("1829101"), but it can be a `_` separated js literals :(

yudilevi: weird, had no idea

yudilevi (approved):

utils/README.md
barakman: AFAIK, using very large numbers in native format doesn't cause any exception, but rather an inaccurate interpretation of that number during runtime.

test/Deployment.js
lbeder: It hasn't...

utils/README.md
lbeder: You're thinking about the older library we were using.

barakman: It's general to JS AFAIK (not related to a specific package). A number larger than `Number.MAX_SAFE_INTEGER` is not safe in terms of accuracy.

lbeder: AFAIK, it is handled by "Big Number" libraries exactly just for that

PR #639, opened 2021-06-02
Move infra to ts

sshmaxime:


hardhat.config.ts
```diff
@@ -0,0 +1,81 @@
...
+                count: 10,
+                accountsBalance: '10000000000000000000000000000'
+            }
+        },
+
+        ...configNetworks
+    },
+
+    solidity: {
+        version: '0.6.12',
```
barakman, line 45: Perhaps we should read this during runtime from package.json, i.e., something like:
```
JSON.parse(fs.readFileSync('package.json')).devDependencies.solc
```

barakman (changes requested):

hardhat.config.ts
lbeder: I think it's overkill. We'd still have many other locations (solhint, pragmas, etc.) with an explicit version specified

barakman: Of course, first and foremost - the contracts themselves.
But at least we'll have this dependency slightly reduced.
I think it's worth the cost, unless it yields a performance hit upon starting the tests.

lbeder: It's not standard. BTW, keep in mind that the only reason we're installing `solc` in `package.json` is for solhint. Hardhat (and truffle, previously) download their own version of `solc`. So your suggestion might break things :(

lbeder (approved):

yudilevi (approved):

barakman (approved):

PR #641, opened 2021-06-07
Refactor natspec

lbeder:
- [x] Remove @param and @return values
- [x] Add explicit "Requirements" section for security and other non-trivial assumptions

### DIFF
--- a/contracts/IConversionPathFinder.sol
+++ b/contracts/IConversionPathFinder.sol
@@ -3,9 +3,9 @@ pragma solidity 0.6.12;
 
 import "./token/interfaces/IReserveToken.sol";
 
-/*
-    Conversion Path Finder interface
-*/
+/**
+ * @dev Conversion Path Finder interface
+ */
 interface IConversionPathFinder {
     function findPath(IReserveToken sourceToken, IReserveToken targetToken) external view returns (address[] memory);
 }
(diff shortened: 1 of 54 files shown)


contracts/IConversionPathFinder.sol
```diff
@@ -3,9 +3,9 @@ pragma solidity 0.6.12;
 
 import "./token/interfaces/IReserveToken.sol";
 
-/*
-    Conversion Path Finder interface
-*/
+/**
+ * @dev Conversion Path Finder interface
```
yudilevi, line 7: the reason we didn't add @dev before was that it wasn't supported by gitbooks - if I'm not mistaken.
so we just need to make sure that one works.

contracts/NetworkSettings.sol
```diff
@@ -91,12 +76,14 @@ contract NetworkSettings is INetworkSettings, Owned, Utils {
 
     /**
      * @dev sets the network fee
-     * can be executed only by the owner
      *
-     * @param newNetworkFee new network fee in ppm units
+     * Requirements:
```
yudilevi, line 80: we need to remember to point out that these are in PPM (and later on rename it to feePPM)

yudilevi (approved):

contracts/IConversionPathFinder.sol
lbeder: Ah, super weird. It's pretty common. In any case, it'd be easy to revert later on

contracts/NetworkSettings.sol
lbeder: I'll add a comment

yudilevi: cool

PR #669, opened 2022-01-30
Update the style to the latest version of prettier

lbeder:


yudilevi (approved):

barakman (approved):
Are you sure that we want to open this Pandora Box?
What added value is there in fixing V2.1 semantically (style, etc), unless we are going to redeploy it and continue developing it?
Seems like a bottomless bit.

PR #671, opened 2022-02-01
Update type generation and ensure that custom types are published as well

lbeder:


yudilevi (approved):

test/staking-rewards/StakingRewards.js
```diff
@@ -661,7 +664,7 @@ describe('StakingRewards', () => {
 
             const reward = await staking.pendingRewards(providerAddress);
 
-            const effectiveTime = BigNumber.min(now, programEndTime);
+            const effectiveTime = min(now, programEndTime);
```
barakman, line 667: Minor: I'm pretty sure that you can use `Math.min` in all of those which take time values as input.

PR #676, opened 2022-02-14
Optimize `LiquidityProtection.migratePositions`

barakman:
1. Optimize function `migratePositions` to handle only an array with all positions being of the same type.
2. Extend the liquidity-migration tests to verify the migration of several positions on several pools by several providers.
3. Unrelated of this PR - fix format in contract `StakingRewards`.

### DIFF
--- a/contracts/liquidity-protection/LiquidityProtection.sol
+++ b/contracts/liquidity-protection/LiquidityProtection.sol
@@ -1,5 +1,6 @@
 // SPDX-License-Identifier: SEE LICENSE IN LICENSE
 pragma solidity 0.6.12;
+pragma experimental ABIEncoderV2;
 
 import "@openzeppelin/contracts/math/Math.sol";
 import "@openzeppelin/contracts/math/SafeMath.sol";
@@ -81,6 +82,12 @@ contract LiquidityProtection is ILiquidityProtection, Utils, Owned, ReentrancyGu
         uint128 removeSpotRateD; // spot rate of 1 A in units of B when liquidity is removed (denominator)
         uint128 removeAverageRateN; // average rate of 1 A in units of B when liquidity is removed (numerator)
         uint128 removeAverageRateD; // average rate of 1 A in units of B when liquidity is removed (denominator)
+    }
+
+    struct MigrationUnit {
+        IDSToken poolToken; // pool token address
+        IReserveToken reserveToken; // reserve token address
+        uint256[] positionIds; // position ids
     }
 
     uint256 internal constant MAX_UINT128 = 2**128 - 1;
@@ -546,7 +553,7 @@ contract LiquidityProtection is ILiquidityProtection, Utils, Owned, ReentrancyGu
      * removes network tokens
      */
     function removeLiquidity(uint256 id, uint32 portion) external override nonReentrant validPortion(portion) {
-        _removeLiquidity(msg.sender, id, portion, false);
+        _removeLiquidity(msg.sender, id, portion);
     }
 
     /**
@@ -555,18 +562,17 @@ contract LiquidityProtection is ILiquidityProtection, Utils, Owned, ReentrancyGu
     function _removeLiquidity(
         address payable provider,
         uint256 id,
-        uint32 portion,
-        bool isMigrating
+        uint32 portion
     ) internal {
         // remove the position from the store and update the stats and the last removal checkpoint
-        Position memory removedPos = _removePosition(provider, id, portion, isMigrating);
+        Position memory removedPos = _removePosition(provider, id, portion);
 
         // add the pool tokens to the system
         _systemStore.incSystemBalance(removedPos.poolToken, removedPos.poolAmount);
 
         // if removing network token liquidity, burn the governance tokens from the caller. we need to transfer the
         // tokens to the contract itself, since only token holders can burn their tokens
-        if (!isMigrating && _isNetworkToken(removedPos.reserveToken)) {
+        if (_isNetworkToken(removedPos.reserveToken)) {
             _govToken.safeTransferFrom(provider, address(this), removedPos.reserveAmount);
             _govTokenGovernance.burn(removedPos.reserveAmount);
         }
@@ -595,27 +601,14 @@ contract LiquidityProtection is ILiquidityProtection, Utils, Owned, ReentrancyGu
             removedPos.reserveAmount,
             packedRates,
             removedPos.timestamp,
-            isMigrating ? MAX_UINT256 : _time()
+            _time()
         );
 
         // remove network token liquidity
         if (_isNetworkToken(removedPos.reserveToken)) {
-            if (isMigrating) {
-                // mint network tokens for this contract and migrate them
-                _mintNetworkTokens(address(this), removedPos.poolToken, removedPos.reserveAmount);
-                _networkToken.approve(address(_networkV3), removedPos.reserveAmount);
-                _networkV3.migrateLiquidity(
-                    IReserveToken(address(_networkToken)),
-                    provider,
-                    targetAmount,
-                    targetAmount,
-                    removedPos.reserveAmount
-                );
-            } else {
-                // mint network tokens for the caller and lock them
-                _mintNetworkTokens(address(_wallet), removedPos.poolToken, targetAmount);
-                _lockTokens(provider, targetAmount);
-            }
+            // mint network tokens for the caller and lock them
+            _mintNetworkTokens(address(_wallet), removedPos.poolToken, targetAmount);
+            _lockTokens(provider, targetAmount);
             return;
         }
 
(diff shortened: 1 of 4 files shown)


contracts/liquidity-protection/LiquidityProtection.sol
```diff
@@ -639,42 +632,145 @@ contract LiquidityProtection is ILiquidityProtection, Utils, Owned, ReentrancyGu
...
+            PackedRates memory packedRates = _packRates(
+                removedPos.poolToken,
                 removedPos.reserveToken,
-                provider,
+                removedPos.reserveRateN,
+                removedPos.reserveRateD
+            );
+
+            // verify rate deviation as early as possible in order to reduce gas-cost for failing transactions
+            _verifyRateDeviation(
```
lbeder, line 708: We can move this check outside of the loop (i.e., it's enough to check that the pool is unstable once before proceeding with any migrations)

test/liquidity-protection/LiquidityProtection.js
```diff
@@ -3802,6 +3426,536 @@ describe('LiquidityProtection', () => {
...
                 });
             });
+
+            /**
+             * The liquidity-migration tests consist of 3 standard-converter pools:
+             * - XXX/BNT
+             * - YYY/BNT
+             * - ETH/BNT
+             * The deployer of the pools adds an initial liquidity to each one of them.
+             * Each one of 3 providers then adds 9 positions of protected-liquidity to each reserve token on each pool.
```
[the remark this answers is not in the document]
lbeder: Style: We also don't end comments with "."

test/liquidity-protection/LiquidityProtection.js
```diff
@@ -3802,6 +3426,536 @@ describe('LiquidityProtection', () => {
...
+             * - Total amount of ETH in the ETH/BNT pool
+             * - Total amount of BNT in the ETH/BNT pool
+             * - Per provider amount of XXX in the XXX/BNT pool
+             * - Per provider amount of BNT in the XXX/BNT pool
+             * - Per provider amount of YYY in the YYY/BNT pool
+             * - Per provider amount of BNT in the YYY/BNT pool
+             * - Per provider amount of ETH in the ETH/BNT pool
+             * - Per provider amount of BNT in the ETH/BNT pool
+             */
+            describe('liquidity-migration tests', () => {
```
lbeder, line 3480: `migration`

test/liquidity-protection/LiquidityProtection.js
```diff
@@ -3802,6 +3426,536 @@ describe('LiquidityProtection', () => {
...
+                                expect(currState.providerNetworkBalance).to.equal(prevState.providerNetworkBalance);
+                                expect(currState.providerPoolBalance).to.equal(prevState.providerPoolBalance);
+                                expect(currState.vaultGovBalance).to.equal(prevState.vaultGovBalance);
+                                expect(currState.vaultPoolBalance).to.equal(prevState.vaultPoolBalance);
+                                expect(currState.walletGovBalance).to.equal(prevState.walletGovBalance);
+                                expect(currState.walletBaseBalance).to.equal(prevState.walletBaseBalance);
+                                expect(currState.walletNetworkBalance).to.equal(prevState.walletNetworkBalance);
+
+                                expect(currState.vaultNetworkBalance).to.equal(networkReserveAmount.mul(3)); // TODO: investigate
+
```
lbeder, line 3708: TODO: investigate?

test/liquidity-protection/LiquidityProtection.js
```diff
@@ -3802,6 +3426,536 @@ describe('LiquidityProtection', () => {
...
+                        protectedLiquidityIds,
+                        positions,
+                        xxxToken.poolToken,
+                        xxxToken
+                    );
+
+                    await expect(
+                        liquidityProtection.connect(providers[1]).migratePositions([migrationUnit])
+                    ).to.be.revertedWith('ERR_ACCESS_DENIED');
+                });
```
lbeder, line 3928: Can you add a check of trying to migrate the same position ID (i.e., duplicated) in the same migration unit?

test/liquidity-protection/LiquidityProtection.js
lbeder, line 3480: There don't seem to be tests migrating multiple "migration units" in the same call

lbeder (changes requested):
I left a few comments, but I still need to go over the tests and the logic again. It's excessive and requires more attention

contracts/liquidity-protection/LiquidityProtection.sol
```diff
@@ -639,42 +632,145 @@ contract LiquidityProtection is ILiquidityProtection, Utils, Owned, ReentrancyGu
...
+    }
+
+    /**
+     * @dev migrates a list of positions to v3
+     *
+     * Requirements:
+     *
+     * - the caller must be the owner of all of the positions
+     */
+    function migratePositions(MigrationUnit calldata migrationUnit) internal {
```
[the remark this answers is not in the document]
barakman: OK, suggestions on using `calldata` vs `memory` here (regardless of our convention to use `calldata` for external functions and `memory` for internal functions)?

test/liquidity-protection/LiquidityProtection.js
```diff
@@ -3802,6 +3426,536 @@ describe('LiquidityProtection', () => {
...
+                        expect(currState.providerPoolBalance).to.equal(prevState.providerPoolBalance);
+                        expect(currState.vaultGovBalance).to.equal(prevState.vaultGovBalance);
+                        expect(currState.vaultPoolBalance).to.equal(prevState.vaultPoolBalance);
+                        expect(currState.vaultNetworkBalance).to.equal(prevState.vaultNetworkBalance);
+                        expect(currState.walletGovBalance).to.equal(prevState.walletGovBalance);
+                        expect(currState.walletBaseBalance).to.equal(prevState.walletBaseBalance);
+                        expect(currState.walletNetworkBalance).to.equal(prevState.walletNetworkBalance);
+
+                        expect(currState.vaultBaseBalance).to.equal(prevState.vaultBaseBalance.add(deltaPoolAmount)); // TODO: investigate
+                        expect(currState.walletPoolBalance).to.equal(prevState.walletPoolBalance.sub(deltaPoolAmount));
```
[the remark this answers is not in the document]
barakman: Same as above.

test/liquidity-protection/LiquidityProtection.js
barakman: The primary test (the first `it`) does.
Search for `migratePositions(migrationUnits)`.

test/liquidity-protection/LiquidityProtection.js
```diff
@@ -3802,6 +3426,536 @@ describe('LiquidityProtection', () => {
                     }
                 });
             });
+
+            /**
+             * The liquidity-migration tests consist of 3 standard-converter pools:
```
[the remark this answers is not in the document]
barakman: Fixed

test/liquidity-protection/LiquidityProtection.js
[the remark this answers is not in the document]
barakman: Fixed

contracts/liquidity-protection/LiquidityProtection.sol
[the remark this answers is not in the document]
barakman: But propagating it down the call stack seems possible... so keep it this way or conduct a gas test to see which one is better (or follow our convention and use `memory`)?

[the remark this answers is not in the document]
barakman: OK, so fixed the initial comment

test/liquidity-protection/LiquidityProtection.js
lbeder: Continuing our call, you don't really need to check v3 balances in these tests

barakman: Contract `TestBancorNetwork` mocks function `migrateLiquidity`, which merely transfers the tokens to some preconfigured "vault" address.
These extra checks are for ensuring that the function was called with correct amounts.
In a sense, they are partially redundant (because the mock function could do anything that we'd put into it).
However, they still have some added value in verifying that the real function is being called (internally from function `migratePositions`) with the correct input.
For this added value, I think that we should keep these extra checks.

contracts/liquidity-protection/LiquidityProtection.sol
barakman: The wasteful part is when `_reserveTokenRates` is called from `_packRates` (for each position).
That one could possibly be optimized to execute once for all positions, but we'll probably need to implement separate versions for both `_packRates` and `_reserveTokenRates`.
Function `_verifyRateDeviation` itself will still need to be called for each position, but that's pretty negligible in terms of gas-consumption.

barakman: A quick check on that:
Seems to work.
Contract size increases to 24595 (19 bytes above the maximum), so we will need to find a workaround.

barakman: Fixed

lbeder: `_verifyRateDeviation` doesn't need to be called once for each position. I think that you misunderstand its purpose.

re: contract size - you can deduplicate back `removeLiquidity` for both of the cases

barakman: You're right about `_verifyRateDeviation `.
It's pretty insignificant in terms of gas though, but now that I moved the more significant piece (a part of `_packRates`) outside of the loop, I can also move this one pretty easily.

test/liquidity-protection/LiquidityProtection.js
barakman: Nevertheless, I have split that test (of migrating a position more than once in the same transaction), to:
1. A test of migrating a position more than once in the same migration unit
2. A test of migrating a position more than once in separate migration units

contracts/liquidity-protection/LiquidityProtection.sol
yudilevi: good, I think it's better to only call it once in general

contracts/liquidity-protection/LiquidityProtection.sol
```diff
@@ -639,42 +632,147 @@ contract LiquidityProtection is ILiquidityProtection, Utils, Owned, ReentrancyGu
...
+    }
+
+    /**
+     * @dev migrates a set of position lists to v3
+     *
+     * Requirements:
+     *
+     * - the caller must be the owner of all of the positions
+     */
+    function migratePositions(MigrationUnit[] calldata migrationUnits) external nonReentrant {
```
yudilevi, line 665: I don't quite understand the flow - this is an external function while the migration flow needs only a list of position ids as input - doesn't make sense to me to expect the caller to input this new struct (and a good frontend is still not a solution).
the proper flow should be:
- receive the list of position ids
- build the input struct for the internal flow
- execute

I understand it will increase the contract size but we need to find a way to do that.

contracts/liquidity-protection/LiquidityProtection.sol
```diff
@@ -83,6 +84,12 @@ contract LiquidityProtection is ILiquidityProtection, Utils, Owned, ReentrancyGu
         uint128 removeAverageRateD; // average rate of 1 A in units of B when liquidity is removed (denominator)
     }
 
+    struct MigrationUnit {
```
yudilevi, line 87: `AggregatedPositions`?

contracts/liquidity-protection/LiquidityProtection.sol
```diff
@@ -639,42 +632,147 @@ contract LiquidityProtection is ILiquidityProtection, Utils, Owned, ReentrancyGu
...
+            // get the various rates between the reserves upon adding liquidity and now
+            PackedRates memory packedRates = _packRates(
+                removedPos.reserveRateN,
+                removedPos.reserveRateD,
+                removeSpotRate,
+                removeAverageRate
+            );
+
+            // get the target token amount
+            targetAmount = targetAmount.add(
```
yudilevi, line 718: is there any value in trying to find out the aggregated IL and then calculating the target amount only once, as if it was all a big single position? any significant gas savings?

test/liquidity-protection/LiquidityProtection.js
```diff
@@ -3802,6 +3426,565 @@ describe('LiquidityProtection', () => {
...
+             * +---------------+--------------------+-----------------+-----------------------------+
+             * | YYY Token     |          I         |        D        |              I              |
+             * +---------------+--------------------+-----------------+-----------------------------+
+             * | ETH "Token"   |          I         |        D        |              I              |
+             * +---------------+--------------------+-----------------+-----------------------------+
+             * | Network Token |          I         |        D        |              I              |
+             * +---------------+--------------------+-----------------+-----------------------------+
+             * | Pool Token    |          I         |        I        |              D              |
+             * +---------------+--------------------+-----------------+-----------------------------+
+             * the ones marked I are expected to be identical before and after each migration request
```
yudilevi, line 3457: maybe just remove the I, replace the D with X and note that the ones that are marked with X are expected to change after a migration, while others aren't

test/liquidity-protection/LiquidityProtection.js
```diff
@@ -3802,6 +3426,565 @@ describe('LiquidityProtection', () => {
...
+                const readState = async (baseToken, provider) => ({
+                    providerGovBalance: await govToken.balanceOf(provider.address),
+                    providerBaseBalance: await baseToken.balanceOf(provider.address),
+                    providerNetworkBalance: await networkToken.balanceOf(provider.address),
+                    providerPoolBalance: await baseToken.poolToken.balanceOf(provider.address),
+                    vaultGovBalance: await govToken.balanceOf(bancorVault.address),
+                    vaultBaseBalance: await baseToken.balanceOf(bancorVault.address),
+                    vaultNetworkBalance: await networkToken.balanceOf(bancorVault.address),
+                    vaultPoolBalance: await baseToken.poolToken.balanceOf(bancorVault.address),
+                    walletGovBalance: await govToken.balanceOf(liquidityProtectionWallet.address),
```
yudilevi, line 3512: we can call it protectionGovBalance etc. instead of walletGovBalance which is very generic

test/liquidity-protection/LiquidityProtection.js
yudilevi: if we could check that the migrateLiquidity function is called with the proper input, that should be a sufficient, as we're testing the liquidity protection migration logic here, not the v3 network contract

yudilevi (changes requested):

contracts/liquidity-protection/LiquidityProtection.sol
lbeder: What about the original suggestion to provide multiple arrays (which is similar to `MigrationUnit`, just omitting the pool and reserve token and fetching them from the first position)?

yudilevi: it's better but 1 array is still cleaner - main issue is the binary size or something else?

lbeder: One array is both bad UX and also adds unnecessary logic to the code itself. Call?

yudilevi: sure

barakman: My original design was to fetch the pool token and the reserve token from the position of the first id, and then verify on the position of each one of the remaining IDs.
However, I then realized that it is better to just have the user (FE in this case) passing these tokens as part of the input, because of the following advantages:
1. Most importantly - it compels the user to "know what he's doing"
2. It reduces the gas cost of the transaction
3. It reduces the contract byte-code size
4. It simplifies the code

test/liquidity-protection/LiquidityProtection.js
barakman: As I've explained above. this is exactly what we're doing here (checking that the `migrateLiquidity` function is called with the proper input).

contracts/liquidity-protection/LiquidityProtection.sol
barakman: That's not so possible, since different positions may have different impermanent losses and/or different protection levels attached to them.
However, a part of function `_removeLiquidityTargetAmount` does seem to be identical for all positions - a call to function `_poolTokenRate`, which yields 2 external calls.
So we can split function `_removeLiquidityTargetAmount` in a similar manner to how we've split function `_packRates` (called right before it), and perform these 2 external calls once for all positions.

contracts/liquidity-protection/LiquidityProtection.sol
yudilevi: discussed with @lbeder , array of arrays might be a bit better but it's not worth it as it'll require a frontend anyway, so might as well just leave it as is (with some renames - different discussion)

contracts/liquidity-protection/LiquidityProtection.sol
yudilevi: protection levels are accelerated to 100% for all of them but I think you're right that it might just be too complicated to normalize them based on the IL/weight of each :) fair enough

contracts/liquidity-protection/LiquidityProtection.sol
barakman: That won't work, because the main function takes `MigrationUnit[] migrationUnits` as input.

yudilevi: not following you, it's a rename.
`AggregatedPositions aggregatedPositions`

test/liquidity-protection/LiquidityProtection.js
barakman: Done

contracts/liquidity-protection/LiquidityProtection.sol
barakman: The only significant part remaining in `_removeLiquidityTargetAmount` is:
```
uint256 minProtectionDelay = _settings.minProtectionDelay();
uint256 maxProtectionDelay = _settings.maxProtectionDelay();
```
We can replace it with constant values (which will impact both `migratePositions` and `removePosition`).
@yudilevi @lbeder : WDYT?

contracts/liquidity-protection/LiquidityProtection.sol
barakman: You're suggestion a rename from single to plural, which means that we'll need to use:
```
AggregatedPositions[] aggregatedPositionss
```
Which sounds pretty bad.

yudilevi: we can just use -
AggregatedPositions aggregatedPositions
AggregatedPositions[] aggregatedPositions
it's not perfect but it's clear enough and much more self explanatory than MigrationUnit which is very generic and doesn't help the user understand anything about the content

contracts/liquidity-protection/LiquidityProtection.sol
barakman: NM, refactored the code to get rid of these two external calls for `migratePositions` only.

contracts/liquidity-protection/LiquidityProtection.sol
barakman: Renamed to `PositionList`.
LMK if you want to add or change to `AggregatedPositions` or `AggregatedPositionList` or w/e.

yudilevi: Clear enough.

test/liquidity-protection/LiquidityProtection.js
barakman: Done

yudilevi (approved):

lbeder (approved):

PR #679, opened 2022-04-26
Reduce LiquidityProtection's contract size

lbeder:


yudilevi (approved):

barakman (commented):
Possibly better to replace all getters with a single getter, so we don't lose any vital functionality, while still reducing code size.

PR #694, opened 2022-11-13
Remove protection from withdraws

adamshai:
- Remove protection from withdraws
- If the pool is in deficit (total liquidity >= total user value), decrease the withdraw amount by the deficit factor as follows
```
deficit factor = total liquidity / total user value
withdraw amount *= deficit factor
```


adamshai:
[v2.1 - update withdrawal calculation](https://trello.com/c/iHO86ljJ/1248-v21-update-withdrawal-calculation)

PR #697, opened 2022-11-15
Release 0.7.13

yudilevi:


contracts/liquidity-protection/LiquidityProtection.sol
```diff
@@ -1290,62 +1203,12 @@ contract LiquidityProtection is ILiquidityProtection, Utils, Owned, ReentrancyGu
...
-        uint256 amount,
-        uint256 total,
-        Fraction memory loss,
-        Fraction memory level
-    ) internal pure returns (uint256) {
-        uint256 levelN = level.n.mul(amount);
-        uint256 levelD = level.d;
-        uint256 maxVal = Math.max(Math.max(levelN, levelD), total);
+    function _deductIL(uint256 value, Fraction memory loss) internal pure returns (uint256) {
+        uint256 maxVal = Math.max(1, value);
```
barakman, line 1209: I'm pretty sure that `value` is always larger than 0, making this redundant (i.e., we can just use `value` instead of `maxVal`).

barakman (changes requested):

contracts/liquidity-protection/LiquidityProtection.sol
```diff
@@ -636,24 +626,19 @@ contract LiquidityProtection is ILiquidityProtection, Utils, Owned, ReentrancyGu
...
 
+        // reduce the total positions value
+        uint256 totalValue = _totalPositionsValue[removedPos.poolToken];
+        if (totalValue < positionValue) {
+            _totalPositionsValue[removedPos.poolToken] = 0;
+        }
+        else {
+            _totalPositionsValue[removedPos.poolToken] = totalValue.sub(positionValue);
+        }
+
```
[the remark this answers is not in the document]
yudilevi: updated

contracts/liquidity-protection/LiquidityProtection.sol
```diff
@@ -662,132 +647,55 @@ contract LiquidityProtection is ILiquidityProtection, Utils, Owned, ReentrancyGu
...
-        }
-        _networkV3.migrateLiquidity{ value: value }(
-            reserveToken,
-            msg.sender,
-            fullyProtectedAmount,
-            removedAmount,
-            originalAmount
-        );
+        // get the pool deficit PPM
+        uint256 deficitPPM = poolDeficitPPM(poolToken);
```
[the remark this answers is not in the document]
yudilevi: trying to avoid returning a Fraction externally from this contract, and the poolDeficit function is public

contracts/liquidity-protection/LiquidityProtection.sol
```diff
@@ -876,6 +781,44 @@ contract LiquidityProtection is ILiquidityProtection, Utils, Owned, ReentrancyGu
...
+        uint256 protectedPoolTokenAmount = poolToken.balanceOf(address(_wallet));
+        // if the positions pool token amount is greater or equal to the total
+        // protected pool token amount, there's nothing to migrate
+        if (positionsPoolTokenAmount >= protectedPoolTokenAmount) {
+            return 0;
+        }
+
+        // deduct the positions pool toke amount from the total protected pool tokens amount
+        // and limit it by the system balance
+        uint256 poolAmountToMigrate = protectedPoolTokenAmount.sub(positionsPoolTokenAmount);
```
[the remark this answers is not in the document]
yudilevi: keeping the utilities for readability

contracts/liquidity-protection/LiquidityProtection.sol
yudilevi: I believe it's your code so I assume you had a reason for it, might as well keep it and not get surprised later :)

contracts/liquidity-protection/LiquidityProtection.sol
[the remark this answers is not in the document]
barakman: No need for 'sub', but ok...

test/liquidity-protection/LiquidityProtection.js
```diff
@@ -2920,94 +3133,6 @@ describe('LiquidityProtection', () => {
...
-                                        condOrAlmostEqual(
-                                            actual.lt(expected),
-                                            actual,
-                                            expected,
-                                            { 1: '0.0', 3: '0.0' }[converterType]
-                                        );
-                                } else if (config.increaseRate && !config.generateFee && numOfDays >= 100) {
-                                    test = (actual, expected) =>
-                                        condOrAlmostEqual(
-                                            actual.eq(expected),
```
barakman, line 2957: I wanna reinstate this entire block, and just change `eq` to `lt` here (which will that when the rate has changed AND no fees were generated AND the vesting period has ended, the user is bound to lose). The other edge cases remain unchanged, and will serve as part of a more thorough test suite.

yudilevi: go for it

contracts/liquidity-protection/LiquidityProtection.sol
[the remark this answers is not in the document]
yudilevi: it's only worth it if the precision loss is significant, can you give an idea with a real world example?
if we're talking about 0.001% deviation, it's not worth the change

contracts/liquidity-protection/LiquidityProtection.sol
```diff
@@ -92,7 +82,6 @@ contract LiquidityProtection is ILiquidityProtection, Utils, Owned, ReentrancyGu
     uint256 internal constant MAX_UINT128 = 2**128 - 1;
```
barakman, line 82: Minor: Can get rid of `struct PositionList` declaration (above).

yudilevi: done

contracts/liquidity-protection/LiquidityProtection.sol
```diff
@@ -908,6 +839,44 @@ contract LiquidityProtection is ILiquidityProtection, Utils, Owned, ReentrancyGu
...
+        // protected pool token amount, there's nothing to migrate
+        if (positionsPoolTokenAmount >= protectedPoolTokenAmount) {
+            return 0;
+        }
+
+        // deduct the positions pool toke amount from the total protected pool tokens amount
+        // and limit it by the system balance
+        uint256 poolAmountToMigrate = protectedPoolTokenAmount.sub(positionsPoolTokenAmount);
+        uint256 systemPoolAmount = _systemStore.systemBalance(poolToken);
+        if (poolAmountToMigrate > systemPoolAmount) {
```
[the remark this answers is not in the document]
yudilevi: updated

contracts/liquidity-protection/LiquidityProtection.sol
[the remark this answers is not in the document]
barakman: Submitted a PR for you to consider, so resolving this one

contracts/liquidity-protection/LiquidityProtection.sol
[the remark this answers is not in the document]
yudilevi: readability/lower risk > insignificant gas optimization

contracts/liquidity-protection/LiquidityProtection.sol
```diff
@@ -662,181 +619,137 @@ contract LiquidityProtection is ILiquidityProtection, Utils, Owned, ReentrancyGu
...
-        return _compensationAmount(reserveAmount, Math.max(reserveAmount, total), loss, level);
+        // if the protected liquidity is equal or greater than the total value,
+        // the pool is not in deficit
+        if (protectedLiquidity >= totalValue) {
+            return Fraction({ n: 0, d: 1 });
+        }
+
+        // the pool is in deficit
+        return Fraction({
+            n: totalValue.sub(protectedLiquidity),
```
barakman, line 750: Minor, can use `totalValue - protectedLiquidity` at this point.

contracts/liquidity-protection/LiquidityProtection.sol
```diff
@@ -662,181 +619,137 @@ contract LiquidityProtection is ILiquidityProtection, Utils, Owned, ReentrancyGu
...
+        view
+        returns (uint256)
+    {
+        Fraction memory poolDeficit = _poolDeficit(poolToken);
 
-        // calculate the protection level
-        Fraction memory level = _protectionLevel(addTimestamp, removeTimestamp);
+        // no deficit
+        if (poolDeficit.n == 0) {
+            return 0;
```
barakman, line 713: Already handled correctly in the final `return` statement.

contracts/liquidity-protection/LiquidityProtection.sol
```diff
@@ -662,181 +619,137 @@ contract LiquidityProtection is ILiquidityProtection, Utils, Owned, ReentrancyGu
...
-        Fraction memory loss = _impLoss(addSpotRate, removeAverageRate);
+        // get the total positions value
+        uint256 totalValue = totalPositionsValue(poolToken);
 
-        // calculate the compensation amount
-        return _compensationAmount(reserveAmount, Math.max(reserveAmount, total), loss, level);
+        // if the protected liquidity is equal or greater than the total value,
+        // the pool is not in deficit
+        if (protectedLiquidity >= totalValue) {
+            return Fraction({ n: 0, d: 1 });
```
barakman, line 745: Good thinking; this way, neither one of the two functions which use this function needs to worry about the of 'no deficit' being possibly indicated by 0/0.

contracts/liquidity-protection/LiquidityProtection.sol
```diff
@@ -662,181 +619,137 @@ contract LiquidityProtection is ILiquidityProtection, Utils, Owned, ReentrancyGu
...
-        // migrate the received tokens
-        uint256 removedAmount = reserveToken.balanceOf(address(this));
-        uint256 value;
-        if (reserveToken.isNativeToken()) {
-            value = removedAmount;
-        } else {
-            IERC20(address(reserveToken)).safeApprove(address(_networkV3), removedAmount);
+        // no deficit
+        if (poolDeficit.n == 0) {
+            return (targetAmount, targetAmount);
```
barakman, line 689: Already handled correctly in the final `return` statement.

barakman (changes requested):

contracts/liquidity-protection/LiquidityProtection.sol
yudilevi: should never use unchecked operators unless overflow/underflow is desired, the benefit is insignificant, a potential bug can be disastrous.

contracts/liquidity-protection/LiquidityProtection.sol
yudilevi: removed

contracts/liquidity-protection/LiquidityProtection.sol
yudilevi: removed 👍

contracts/liquidity-protection/LiquidityProtection.sol
yudilevi: removed

test/liquidity-protection/LiquidityProtection.js
```diff
@@ -2325,6 +2338,277 @@ describe('LiquidityProtection', () => {
...
+                            // behalf of the provider. HALF of the pool tokens are allocated to the provider, and HALF to
+                            // the system
+                            const poolTokenSupply = await poolToken.totalSupply();
+                            await addV2Liquidity(provider, depositAmount, isEthReserve);
+
+                            await liquidityProtection.setTotalPositionsValue(poolToken.address, depositAmount);
+
+                            const poolDeficit = await liquidityProtection.poolDeficitPPM(poolToken.address);
+                            expect(poolDeficit).equal(0);
+                        });
```
barakman, line 2591: Same test twice.

test/liquidity-protection/LiquidityProtection.js
```diff
@@ -2325,6 +2338,277 @@ describe('LiquidityProtection', () => {
...
+                            // the system
+                            const poolTokenSupply = await poolToken.totalSupply();
+                            await addV2Liquidity(provider, depositAmount, isEthReserve);
+
+                            await liquidityProtection.setTotalPositionsValue(poolToken.address, depositAmount.mul(100).div(80));
+
+                            const poolDeficit = await liquidityProtection.poolDeficitPPM(poolToken.address);
+                            expect(poolDeficit).equal(200_000);
+                        });
+                    });
```
barakman, line 2610: I would add the following tests:
1. `setTotalPositionsValue(poolToken.address, depositAmount)` // when total == protected
2. `setTotalPositionsValue(poolToken.address, 0)` // when total == 0 < protected
3. `setTotalPositionsValue(poolToken.address, 1000000)` when total > 0 == protected (i.e., without calling addV2Liquidity)

test/liquidity-protection/LiquidityProtection.js
```diff
@@ -2325,6 +2338,277 @@ describe('LiquidityProtection', () => {
...
+                                expect(newProtectedPoolTokens).equals(protectedPoolTokens.mul(80).div(100));
+
+                                // 40% of the system pool token balance should be migrated
+                                expect(await systemPoolTokensBalance()).equals(systemBalance.mul(60).div(100));
+
+                                // another call to migrate should do nothing
+                                await liquidityProtection.migrateSystemPoolTokens([poolToken.address]);
+                                const newProtectedPoolTokens2 = await poolToken.balanceOf(liquidityProtectionWallet.address);
+                                expect(newProtectedPoolTokens2).equals(protectedPoolTokens.mul(80).div(100));                                
+                                expect(await systemPoolTokensBalance()).equals(systemBalance.mul(60).div(100));
```
barakman, line 2529: So where exactly do you verify `'should migrate 20% of the protected pool tokens'`?

test/liquidity-protection/LiquidityProtection.js
```diff
@@ -2325,6 +2338,277 @@ describe('LiquidityProtection', () => {
...
+                            const totalLiquidity = await reserveTokenBalance();
+                            expect(totalLiquidity).equal(RESERVE1_AMOUNT.add(depositAmount));
+                            expect(await systemPoolTokensBalance()).equals(poolTokenSupplyDelta.div(2));
+
+                            await liquidityProtection.setTotalPositionsValue(poolToken.address, totalLiquidity.mul(2));
+                            await liquidityProtection.migrateSystemPoolTokens([poolToken.address]);
+                            expect(await systemPoolTokensBalance()).equals(poolTokenSupplyDelta.div(2));
+                        });
+
+                        context('when the total positions value equals 80% of the total protected liquidity', async () => {
```
barakman, line 2502: I would add a few more test cases, such as:
1. when the total positions value equals 100% of the total protected liquidity
2. when the total positions value equals 120% of the total protected liquidity
3. when the total positions value equals 0 and the total protected liquidity is larger than 0
4. when the total positions value is larger than 0 and the total protected liquidity equals 0

barakman (changes requested):

test/liquidity-protection/LiquidityProtection.js
yudilevi: can you be more specific? I'm not seeing the same test (but rather only the same result which is correct)

test/liquidity-protection/LiquidityProtection.js
yudilevi: `expect(newProtectedPoolTokens).equals(protectedPoolTokens.mul(80).div(100));`
the system is left with 80% of what it had

test/liquidity-protection/LiquidityProtection.js
yudilevi: I might be missing something but I believe these tests already exist - can you be more specific?

test/liquidity-protection/LiquidityProtection.js
barakman: In that case, I'd change the title

test/liquidity-protection/LiquidityProtection.js
barakman: The `context` has probably confused me.
Probably better to place that (single) test on the same level along with the other tests (i.e., outside of that context).

barakman (approved):

test/liquidity-protection/LiquidityProtection.js
yudilevi: migrating 20% of the pool tokens leaves the protocol with 80% which is what we're testing, so I don't see an issue

yudilevi: migrating 20% of the pool tokens leaves the protocol with 80% which is what we're testing, so I don't see an issue

test/liquidity-protection/LiquidityProtection.js
yudilevi: it's there to make the case clearer, otherwise the test description would be very long

